PR #8, opened 2026-04-29
ci: add lint job using uv + pre-commit, remove justfile

janickm:
## Summary

- Add a dedicated `lint` CI job that runs the full pre-commit suite (ruff lint, ruff format, trailing-whitespace, end-of-file-fixer, check-symlinks) inside the `ghcr.io/astral-sh/uv:latest` container
- Manage pre-commit as a `[dependency-groups]` lint dependency in the root `pyproject.toml`, invoked via `uv run --only-group lint pre-commit run -a` — no Docker image rebuild needed
- Remove the `justfile` in favor of uv-managed tooling
- Update `.pre-commit-config.yaml` default Python from 3.10 to 3.12

## Motivation

Lint checks weren't running in CI. This adds them as a lightweight, self-contained job that doesn't require GPU dependencies or a custom Docker image — just the official uv container.


janickm:
closing, we have #9 locally

PR #10, opened 2026-04-29
Chore: add pyright configuration

jatentaki:
Because of our "monorepo" design, `pyright` was not discovering `flashdreams` correctly (trying to load the repo root instead of the python package). This MR fixes this with an explicit configuration block.


janickm (approved):

janickm:
[rebased]

PR #17, opened 2026-04-29
Update Docker build script to include dynamic tagging with date and commit hash

wilsonCernWq:
Add both github and gitlab tags, so that we maintain both registries consistently.

### DIFF
--- a/docker/build_with_docker.sh
+++ b/docker/build_with_docker.sh
@@ -69,12 +69,13 @@
 
 set -eu -o pipefail
 
-TAG=base-v0.3
+TAG=base-v0.3-$(date +%Y%m%d)-$(git rev-parse --short HEAD)
 
 docker buildx build \
     --platform linux/arm64,linux/amd64 \
     --allow network.host \
     --network host \
     --push \
-    -t ghcr.io/nvidia/flashdreams:$TAG-$(date +%Y%m%d)-$(git rev-parse --short HEAD) \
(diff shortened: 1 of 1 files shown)


docker/build_with_docker.sh
```diff
@@ -69,12 +69,13 @@
...
+TAG=base-v0.3-$(date +%Y%m%d)-$(git rev-parse --short HEAD)
 
 docker buildx build \
     --platform linux/arm64,linux/amd64 \
     --allow network.host \
     --network host \
     --push \
-    -t ghcr.io/nvidia/flashdreams:$TAG-$(date +%Y%m%d)-$(git rev-parse --short HEAD) \
+    -t ghcr.io/nvidia/flashdreams:$TAG \
+    -t gitlab-master.nvidia.com:5005/sil/flashdreams:$TAG \
```
liruilong940607, line 80: Do we want to expose this URL in GH?

wilsonCernWq: I think it was already exposed in https://github.com/NVIDIA/flashdreams/pull/7?

liruilong940607: oh yeah true. ok im merging it then. we could do a cleanup later.

janickm: mhh, this is again exposing NV-internals in the public @wilsonCernWq - can we please try to avoid this? we could, e.g., implement this by passing a CLI arg instead for the target registry - but we should stop leaking NV-internal URLs in the public repo

Issue #26, opened 2026-04-30
Lingbot world fails in distributed inference


jatentaki:
Running
```
uv run torchrun --nproc_per_node 4 -m flashdreams.examples.run_lingbot_world --total_blocks 60 --config_name LingBot-World-Fast
```

(very similar to what is suggested at thetop of the example file) results in
```
[rank2]: Traceback (most recent call last):                                                                                                        
[rank2]:   File "<frozen runpy>", line 198, in _run_module_as_main
[rank2]:   File "<frozen runpy>", line 88, in _run_code                                                                                            
[rank2]:   File "/workspace/flashdreams/flashdreams/examples/run_lingbot_world.py", line 264, in <module>
[rank2]:     main()                                                                                                                                
[rank2]:   File "/workspace/flashdreams/flashdreams/examples/run_lingbot_world.py", line 194, in main
[rank2]:     .setup()                                             
[rank2]:      ^^^^^^^                                  
[rank2]:   File "/workspace/flashdreams/flashdreams/flashdreams/infra/config/base.py", line 35, in setup 
[rank2]:     return self._target(self, **kwargs)  # type: ignore[call-arg]
[rank2]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                   
[rank2]:   File "/workspace/flashdreams/flashdreams/flashdreams/recipes/wan/pipeline.py", line 133, in __init__
[rank2]:     super().__init__(config)
[rank2]:   File "/workspace/flashdreams/flashdreams/flashdreams/infra/pipeline/base.py", line 122, in __init__
[rank2]:     self.diffusion_model = config.diffusion_model.setup()                                                                                 
[rank2]:                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank2]:   File "/workspace/flashdreams/flashdreams/flashdreams/infra/config/base.py", line 35, in setup       
[rank2]:     return self._target(self, **kwargs)  # type: ignore[call-arg]
[rank2]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                   
[rank2]:   File "/workspace/flashdreams/flashdreams/flashdreams/infra/diffusion/model/base.py", line 79, in __init__
[rank2]:     self.transformer = self.config.transformer.setup()   
[rank2]:                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                    
[...]

PR #28, opened 2026-04-30
Fix lingbot inference distributed setup

jatentaki:
Lingbot was failing due to improper distributed setup, both with 1 and multiple GPUs - see the [issue](https://github.com/NVIDIA/flashdreams/issues/26). This MR fixes that (to best of my ability).

### DIFF
--- a/flashdreams/examples/run_lingbot_world.py
+++ b/flashdreams/examples/run_lingbot_world.py
@@ -103,20 +103,30 @@ def parse_args() -> argparse.Namespace:
     return parser.parse_args()
 
 
+def _should_init_distributed() -> bool:
+    return os.getenv("RANK") is not None and os.getenv("WORLD_SIZE") is not None
+
+
 def main() -> None:
     args = parse_args()
     print(f"Running Lingbot World inference with config: {args.config_name}")
 
     local_rank = int(os.getenv("LOCAL_RANK", 0))
-    distributed_init()
-    world_size = torch.distributed.get_world_size()
-    rank = torch.distributed.get_rank()
+    if _should_init_distributed():
+        distributed_init()
+        world_size = torch.distributed.get_world_size()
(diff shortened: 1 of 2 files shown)


liruilong940607:
@janickm CI seems always pending for this MR. do you know why

liruilong940607 (approved):

wilsonCernWq (approved):
Looks good, but I don't understand how this would trigger/fix the issue?
@liruilong940607 ?

janickm:
> @janickm CI seems always pending for this MR. do you know why

see https://github.com/NVIDIA/flashdreams/pull/28#issuecomment-4351871129 - you need to sign your commits so that the NV bot copies the commit to a `pull-request/N` branch that is executed by the NV-internal runners (I know it's annoying, but should be simple to setup)

you can also run it like this

janickm:
> > @janickm CI seems always pending for this MR. do you know why
> 
> see [#28 (comment)](https://github.com/NVIDIA/flashdreams/pull/28#issuecomment-4351871129) - you need to sign your commits so that the NV bot copies the commit to a `pull-request/N` branch that is executed by the NV-internal runners (I know it's annoying, but should be simple to setup)
> 
> you can also run it like this

I guess it's @jatentaki that needs to setup gpg signing?

liruilong940607:
> Looks good, but I don't understand how this would trigger/fix the issue? @liruilong940607 ?

I think if you run with `python -m torch.distributed.run --standalone --nnodes=1 --nproc_per_node=1` you probably wont hit the issue but just `python xxx.py` you probably will and this PR fixes this case.

liruilong940607:
/ok to test [9a402e9](https://github.com/NVIDIA/flashdreams/pull/28/commits/9a402e9f6e70780bc9825fbe43dc3f4906eefdfb)

PR #31, opened 2026-04-30
Offload one-shot encoders to lower VRAM in alpadreams inference

gtong-nv:
The Cosmos-Reason1-7B text encoder (~14 GB bf16) and the Wan VAE first-frame image encoder are only used in `initialize_cache`, but currently stay on GPU for the entire AR rollout. This MR adds three ways to reclaim that VRAM, trading peak-VRAM vs. complexity.

### Changes

**`AlpadreamsPipeline` (`flashdreams/recipes/alpadreams/pipeline.py`)**
- `text_encoder` / `image_encoder` configs are now `Optional`; `__init__` skips `setup()` when either is `None`.
- `release_oneshot_encoders()`: drops both encoders, runs `gc.collect()` (cycles in `nn.Module` graphs aren't refcount-freed), then `torch.cuda.empty_cache()`. Sets fields to `None` (not `delattr`) so post-release `initialize_cache` fails with a clear assertion.
- `initialize_cache_from_embeddings(text_embeddings, image_embeddings, view_names)`: same CP-split + cache-build path as `initialize_cache`, but takes precomputed tensors.
- `precompute_embeddings(text, image)`: runs only the encoders and returns CPU tensors. CP-split is intentionally not applied so one saved file works for any CP world size.
- `initialize_cache` now delegates to `initialize_cache_from_embeddings`; assert message covers both "configs were None" and "released".

**`examples/run_alpadreams.py`**
- Default: calls `release_oneshot_encoders()` right after `initialize_cache`.
- `--embeddings_path PATH`: loads a `.pt` produced by a prior `--save_embeddings_path` run; encoders never enter the inference process.
- `--save_embeddings_path PATH`: offline-precompute mode. Loads only the encoders, dumps a `.pt` payload (`text_embeddings`, `image_embeddings`, `view_names`, `metadata`), and exits — no `distributed_init`, no AR pipeline, no rollout. Single-GPU producer; pair with `--embeddings_path` on a subsequent `torchrun` invocation. Mutually exclusive with the other two flags.
- `offload_text_encoder`: loads only the encoders, computes embeddings, frees them, then builds the AR pipeline. No disk artifact, lower peak VRAM than default.

### Three modes
| Mode | Encoders loaded | Disk artifact | Peak VRAM |
|---|---|---|---|
| Default | Yes, with DiT | No | encoders + AR pipeline |
| `--embeddings_path PATH` (after `--save_embeddings_path`) | No | Yes (precompute step) | AR pipeline only |
| `--offload_text_encoder` | Yes, sequentially | No | max(encoders, AR pipeline) |


flashdreams/examples/precompute_alpadreams_embeddings.py
liruilong940607, line 1: Shall we fold this script into `run_alpadreams` with an argument to control it to only dump out embedding?

It would be nice if we can keep the example/ folder "clean" with one script for each model.

liruilong940607 (changes requested):

flashdreams/examples/precompute_alpadreams_embeddings.py
gtong-nv: Good idea. Let me consolidate the logic

gtong-nv: Updated

liruilong940607:
/ok to test [34ece73](https://github.com/NVIDIA/flashdreams/pull/31/commits/34ece73be4903ce3bfaffb7f177a24c3e2ff3d5b)

liruilong940607:
@gtong-nv some lint errors from CI. feel free to add some "type: ignore" for now -- we will do a deep cleaning up to fix the lint errors later.

liruilong940607:
/ok to test [f367f5d](https://github.com/NVIDIA/flashdreams/pull/31/commits/f367f5d198d8a1aa52da03fdda8928a941d14a4d)

gtong-nv:
/ok to test [3db6ce7](https://github.com/NVIDIA/flashdreams/pull/31/commits/3db6ce73d821b2c0c2b43ef43f9c294a29f101a4)

liruilong940607 (approved):

PR #38, opened 2026-05-02
feat: Add bidirectional Alpadreams teacher recipe, improved checkpoint and config logic, and new conversion script

wilsonCernWq:
- Implemented `run_alpadreams_teacher.py` with bidirectional Alpadreams teacher, matching existing teacher API/rollouts, and improved config handling including max frame limits and removal of hardcoded defaults.
- Integrated Cosmos teacher transformer with classifier-free guidance and UniPC scheduler; removed legacy Cosmos SIL HDMap teacher code.
- Refactored checkpoint and distributed loading for clearer logic, robust failure handling, easy path overrides (GCP), and more informative errors.
- Replaced previous conversion scripts with `convert_i4_dcp2pt.py`, standardizing arguments, usage help, and default credential handling for I4 Cosmos SIL DCP to FlashDreams PT format.
- Improved argument parsing and configuration overrides (renamed `--config_name` to `--overwrite_config_name`, clarified options, and added helper builder logic).
- Consolidated imports, improved documentation, and updated configuration for maintainability and reproducibility.
- All major transitions and configuration changes are documented for future reference.


flashdreams/flashdreams/recipes/alpadreams/pipeline.py
```diff
@@ -324,3 +325,340 @@ def get_num_frames(self, autoregressive_index: int) -> int:
...
+
+AlpadreamsTeacherPipelineCache: TypeAlias = StreamInferencePipelineCache[
+    EncoderAutoregressiveCache,
+    CosmosTeacherTransformerCache,
+    DecoderAutoregressiveCache,
+]
+
+
+@dataclass(kw_only=True)
+class AlpadreamsTeacherPipelineConfig(StreamInferencePipelineConfig):
```
liruilong940607: Any reason we need a new pipeline? Would the `AlpadreamsPipeline` with `len_t_latent=21` work for the bidirectional teacher model?

flashdreams/examples/run_alpadreams_teacher.py
liruilong940607, line 1: Would be nice to consolidate it into the same `run_alpadreams.py` script. We can package it into multiple "functions" in that script, one for each task. It feels cleaner if we have just 1 script for each demo under then `example` folder -- and reuse code as much as possible

flashdreams/scripts/convert_i4_dcp2pt.py
liruilong940607, line 1: Maybe we can package this into a function in examples/run_alpadreams.py as well? Feels a small enough function that we can carry in the example script.

flashdreams/flashdreams/recipes/alpadreams/config.py
```diff
@@ -467,3 +472,123 @@ def experiment1_skip_finalize_kv_cache_noise100(
     "experiment1_skip_finalize_kv_cache_noise150": experiment1_skip_finalize_kv_cache_noise150,  # rec
     "experiment1_skip_finalize_kv_cache_noise100": experiment1_skip_finalize_kv_cache_noise100,  # rec
 }
+
+
+# ---------------------------------------------------------------------------
+# Alpadreams teacher (full-block bidirectional 2B / 720p / 189-frame UniPC)
+# ---------------------------------------------------------------------------
+
+ALPADREAMS_TEACHER_CHECKPOINT = (
```
liruilong940607: Do we need to seperate "teacher" with other ckpts? Maybe gather all of them into `ALPADREAMS_CONFIG_BUILDERS`?

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
wilsonCernWq: Merged

flashdreams/examples/run_alpadreams_teacher.py
wilsonCernWq: Yeah, good idea, merged as well

flashdreams/scripts/convert_i4_dcp2pt.py
wilsonCernWq: Done

flashdreams/flashdreams/recipes/alpadreams/config.py
wilsonCernWq: Done

flashdreams/examples/run_alpadreams.py
```diff
@@ -98,6 +101,16 @@
 
 def _build_data(n_cameras: int) -> tuple[list[str], list[dict]]:
     """Hardcoded example data per camera count, matching legacy alpadreams."""
+    negative_prompt = (
```
liruilong940607: I think we can treat this as a CONSTANT. We don't expect user to change this because training is conduct with this negative prompt.

Shall we put it in a similar position like WAN's negative prompt (flashdreams/flashdreams/recipes/wan/constants.py)?

And also similar to how WAN's negative prompt is use -- we can directly access it from Pipeline.initialize_cache

flashdreams/flashdreams/infra/diffusion/scheduler/fm_unipc.py
```diff
@@ -328,7 +328,17 @@ def sample(
         last_sample: Tensor | None = None
 
         for i in range(N):
-            timestep = self.timesteps[i]
+            # ``self.timesteps`` is int64 (kept exact across ``.to(bf16)``
```
liruilong940607: Good catch.

Those comments are too wordy and bot like. Maybe just something similar to what's in the other fm scheduler:

            # Schedule buffers are pinned to fp32 (to preserve integer
            # timestep values under a stray `module.to(bf16)`), but the
            # network expects timesteps in the input dtype so that
            # downstream modulation / Linear layers stay consistent.

flashdreams/flashdreams/recipes/alpadreams/transformer/__init__.py
```diff
@@ -552,3 +551,212 @@ def finalize_kv_cache(
...
+#
+# The Cosmos bidirectional model reuses the Cosmos DiT mechanics above (patchify,
+# RoPE, HDMap branch, condition mask, CP) and adds an unconditional
+# branch for SIL-style classifier-free guidance. It is configured for a
+# single full-block rollout with the UniPC schedule (see the matching
+# pipeline + config in this package).
+
+
+@dataclass(kw_only=True)
+class CosmosBidirectionalTransformerCache(CosmosTransformerCache):
```
liruilong940607: I think we can just put `network_cache_uncond` into `CosmosTransformerCache` and set it to None by default. So that we don't need a different class. Putting them in one place is great to send out the message of "bidirectional is just a special case of causal". Btw -- the `CosmosTransformer` is actually already block bidirectional. What is missing is just the CFG support, so its not quite accurate to create another class and calling it "CosmosBidirectionalTransformer"

Maybe `Wan21TransformerCache` and `Wan21Transformer` in `flashdreams/flashdreams/recipes/wan/transformer/wan21.py` can be a good reference for how to have CFG supported in the same transformer class

flashdreams/flashdreams/recipes/alpadreams/config.py
```diff
@@ -452,13 +457,114 @@ def experiment1_skip_finalize_kv_cache_noise100(
     )
 
 
+# ---------------------------------------------------------------------------
+# Alpadreams bidirectional (single-view / 2B / 720p / chunk48 / UniPC)
+# ---------------------------------------------------------------------------
+
+# TODO: The model is trained with 48 chunks, but generating 48 blocks currently
+# results in OOM, so we use 24 chunks for now. We should fix this later.
+_BIDIRECTIONAL_MAX_NUM_CHUNKS = 24
```
liruilong940607: This is single GB200 OOM limit? This doesn't feel belong to config -- it's gpu arch specific and also related to number of GPU used. Maybe move this assert to the entry script is a better fit?

flashdreams/flashdreams/recipes/alpadreams/config.py
```diff
@@ -452,13 +457,114 @@ def experiment1_skip_finalize_kv_cache_noise100(
...
+        encoder=WanVAEEncoderConfig(
+            checkpoint_path=AVAILABLE_WAN_VAE_CHECKPOINT_PATHS["vae"],
+        ),
+        decoder=decoder_config,
+        diffusion_model=DiffusionModelConfig(
+            seed=seed,
+            # The bidirectional source script's extra KV-cache refresh does not affect
+            # the single-block output. Keep finalize cheap by using clean
+            # context.
+            context_noise=0,
```
liruilong940607: Now we have a "skip_finalize_kv_cache" in `CosmosTransformerConfig` that just skips the KV cache update. This is probably a more informative control.

flashdreams/flashdreams/recipes/alpadreams/config.py
```diff
@@ -452,13 +457,114 @@ def experiment1_skip_finalize_kv_cache_noise100(
...
+                batch_shape=_DEFAULT_BATCH_SHAPE,
+                num_views=1,
+                height=_DEFAULT_VIDEO_HEIGHT // _WAN_VAE_SPATIAL_COMPRESSION,
+                width=_DEFAULT_VIDEO_WIDTH // _WAN_VAE_SPATIAL_COMPRESSION,
+                len_t=num_chunks,
+                cp_size=cp_size,
+                enable_hdmap_condition=True,
+                encode_with_pixel_shuffle=False,
+                h_extrapolation_ratio=1.0,
+                w_extrapolation_ratio=1.0,
```
liruilong940607: Double check to make sure this is 1.0 to align with I4 -- For the final ckpt this is 3.0 for 720P (2.0 for 480P). I assume teacher might have the same logic.

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
```diff
@@ -267,9 +292,21 @@ def precompute_embeddings(
...
+            negative_text_embeddings = torch.stack(
+                [self.text_encoder(t) for t in negative_text], dim=0
+            ).cpu()
+        else:
+            negative_text_embeddings = None
+
         return {
             "text_embeddings": text_embeddings.cpu(),
             "image_embeddings": image_embeddings.cpu(),
+            "negative_text_embeddings": negative_text_embeddings,
```
liruilong940607, line 322: could use a `.cpu()` here to align with others?

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
```diff
@@ -267,9 +292,21 @@ def precompute_embeddings(
             [self.text_encoder(t) for t in text], dim=0
         )  # [B, V, L, D]
         image_embeddings = self.image_encoder(image)  # [B, V, 1, Cl, Hl, Wl]
+
+        if negative_text is not None:
```
liruilong940607: Can use `self.config.guidance_scale > 1.0` as control and fetch the negative text prompt from the constant file instead of accepting it as input. -- similar to `flashdreams/flashdreams/recipes/wan/transformer/wan21.py`

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
```diff
@@ -183,10 +184,19 @@ def initialize_cache(
             [self.text_encoder(t) for t in text], dim=0
         )  # [B, V, L, D]
         image_embeddings = self.image_encoder(image)  # [B, V, 1, Cl, Hl, Wl]
+        negative_text_embeddings: Tensor | None = None
+        if negative_text is not None:
```
liruilong940607: Can use `self.config.guidance_scale > 1.0` as control and fetch the negative text prompt from the constant file instead of accepting it as input. -- similar to `flashdreams/flashdreams/recipes/wan/transformer/wan21.py`

flashdreams/scripts/convert_i4_dcp2pt.py
liruilong940607: Seems like this script is still present -- forget to push?

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
```diff
@@ -224,23 +237,34 @@ def initialize_cache_from_embeddings(
...
+            transformer_context["negative_text_embeddings"] = negative_text_embeddings
+
         return super().initialize_cache(
-            transformer_context={
-                "text_embeddings": text_embeddings,
-                "image_embeddings": image_embeddings,
-                "view_names": view_names,
-            },
+            transformer_context=transformer_context,
         )
```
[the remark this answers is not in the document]
wilsonCernWq: @copilot

flashdreams/flashdreams/recipes/alpadreams/config.py
wilsonCernWq: It is OOM on 4 GB200

flashdreams/examples/run_alpadreams.py
wilsonCernWq: Yeah we can make them constants just as WAN

flashdreams/scripts/convert_i4_dcp2pt.py
wilsonCernWq: Oh, I misunderstood. Idk, so you meant opening a new mode for the run script? I feel these two scripts have different purpose? run script is for running demo, this script just to a data conversion? I actually prefer having them as two scripts?

flashdreams/flashdreams/recipes/alpadreams/config.py
wilsonCernWq: good idea

flashdreams/flashdreams/recipes/alpadreams/transformer/__init__.py
wilsonCernWq: Done

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
wilsonCernWq: negative_text_embeddings is optional, can be none so I think we cannot do .cpu?

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
wilsonCernWq: Do you like the current style?

flashdreams/flashdreams/recipes/alpadreams/config.py
wilsonCernWq: Good catch

flashdreams/flashdreams/infra/diffusion/scheduler/fm_unipc.py
wilsonCernWq: done

flashdreams/flashdreams/recipes/alpadreams/config.py
wilsonCernWq: If you prefer this way, then I need to add an option called num_blocks in the script to control generation.

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
wilsonCernWq: done

flashdreams/flashdreams/recipes/alpadreams/pipeline.py
liruilong940607: Ah right!

flashdreams/scripts/convert_i4_dcp2pt.py
liruilong940607: Actually on a second thought I think I would agree now. we will have more micro scripts we can gather them here.

liruilong940607 (approved):
Thanks for all the changes! LGTM now!

flashdreams/flashdreams/recipes/alpadreams/config.py
liruilong940607: Ok I think the current way is good! thanks!

PR #40, opened 2026-05-03
sync: align with GitLab main

jarcherNV:
Synced all files from GitLab main except internal tooling and gitlab CI.
This establishes a clean baseline for future automated syncing.


jarcherNV:
Looks like this PR undoes a recent github commit. We'll need to cherry-pick that to gitlab first, then I'll update this.

PR #41, opened 2026-05-03
ci: add GitHub to GitLab auto-sync workflow

jarcherNV:
Add GitHub Actions workflow to automatically sync commits from GitHub main to GitLab via merge requests:

- Cherry-picks commits from GitHub onto GitLab sync branches
- Excludes .github/ directory (GitHub-specific CI files)
- Creates/updates GitLab MRs for review before merging
- Preserves commit authorship and dates
- Triggers on PR updates (for testing) and main branch pushes

Requires GitHub secrets: GITLAB_TOKEN, GITLAB_HOST, GITLAB_PROJECT_PATH

Add trivial README marker to ensure sync changes are visible in GitLab MRs.

### DIFF
--- a/.github/workflows/sync-to-gitlab.yml
+++ b/.github/workflows/sync-to-gitlab.yml
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   sync-to-gitlab:
-    runs-on: nv-cpu-general
+    runs-on: linux-amd64-cpu8
     env:
       # Required secrets:
       #   GITLAB_TOKEN        (PAT with write_repository + api scope)
--- a/.github/workflows/sync-to-gitlab.yml
+++ b/.github/workflows/sync-to-gitlab.yml
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   sync-to-gitlab:
-    runs-on: [self-hosted, type/docker]
+    runs-on: nv-cpu-general
     env:
       # Required secrets:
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-to-gitlab.yml
```diff
@@ -0,0 +1,308 @@
...
+  group: gitlab-sync-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  sync-to-gitlab:
+    runs-on: linux-amd64-cpu8
+    env:
+      # Required secrets:
+      #   GITLAB_TOKEN        (PAT with write_repository + api scope)
+      #   GITLAB_HOST         (GitLab host, e.g. gitlab-master.nvidia.com)
```
janickm, line 23: can we actually resolve the internal gitlab from "external" runners?

janickm (changes requested):
most likely what you want to do instead is to have the GL "pull" the GH changes (not the other way around) - this would also simplify the integration on this public-facing repo a lot and we wouldn't "share" internal details

[👎 for now]

.github/workflows/sync-to-gitlab.yml
jarcherNV: No it cannot, so either this would need to use a self hosted runner, or we'd need a different sync mechanism.

jarcherNV:
Closing this PR as we will be sync'ing from the gitlab side instead.

PR #51, opened 2026-05-07
Profiling against self-forcing public codebase carefully

liruilong940607:
TL;DR:
- **A100: ~ same speed**
- **B200: less than 2x speedup: 250ms v.s. 450ms e2e.**

See PERFORMANCE.md for details.

Also fixes a few line of codes to be compatible with older versions of python / torch.

### DIFF
--- a/PERFORMANCE.md
+++ b/PERFORMANCE.md
@@ -1,6 +1,6 @@
 # Self-Forcing
 
-Baselines: 
+Baselines:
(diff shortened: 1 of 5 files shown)


liruilong940607:
/ok to test [17f114c](https://github.com/NVIDIA/flashdreams/pull/51/commits/17f114c4ec61e0082109c3c98c4da0f117eda644)

PR #56, opened 2026-05-08
alpadreams: replace ludus-renderer with SlangPy rasterizer

wlewNV:
The closed-source `cudaraster` C++/CUDA backend that ludus-renderer's CUDA path links against can't ship publicly, which has been blocking release of the Alpadreams gRPC server. This change drops the entire `ludus-renderer==0.5.0+cuda` dependency from `flash-alpadreams` and replaces both the HD-map rasterizer and the JPEG encoder with permissively-licensed alternatives:

- HD-map rasterizer: new `slang_renderer/` subpackage built on top of `slangpy`. Slang compute kernels (lifted from `references/omni-dreams-slangpy/samples/roaddreams`) handle line, triangle, and polygon raster + a per-pixel raytrace pass for near-camera polygons. The slangpy device is constructed via `spy.create_torch_device(...)` so it shares PyTorch's CUDA context; rendered RGBA8 lands in the same address space and is exposed back to PyTorch zero-copy via `Tensor.to_torch()`.
- JPEG encoder: `ludus_renderer.nvjpeg.encode(...)` → `torchvision.io.encode_jpeg(...)`.

The public API of `LudusRenderer` (class name, `__init__` signature, `render_all_frames_and_cameras`, `cleanup`, `load_and_attach_ludus_scene`) is preserved, so neither `AlpadreamsConditioningWrapper` nor `grpc/utils.py` need any changes — the SceneBundle is just attached to `scene_data.metadata['ludus_scene']` like the old SceneAdapter was.

Notable gotchas surfaced and fixed during bring-up:
* `FThetaCameraModel.__post_init__` branches on `reference_poly` because synthetic / single-coefficient forward polynomials make `_compute_bw_poly`'s least-squares inverse unstable, blowing up the angle→radius LUT.
* `triangulate_polygon_xy` auto-picks the dominant plane via `argmin(spread)` instead of always projecting onto (x, y); the flashdreams CLIPGT loader emits ground polygons in OpenCV RDF where y=0, and the original (x, y) projection collapsed every polygon to fan-triangulated wedges.
* `_box_has_null_geometry` and `_is_finite_polyline` guards on every layer builder + a final NaN guard at the GPU upload boundary, since a single non-finite vertex in the line/triangle buffers poisons the entire raster pass via NaN-vs-finite comparison failures (see https://gitlab-master.nvidia.com/sil/omni-dreams/-/commit/92091220).

Workspace + CI plumbing:
* `pyproject.toml` (workspace): drop `ludus_renderer.**` from ty's `replace-imports-with-any` list.
* `pyproject.toml` (alpadreams): drop `ludus-renderer==0.5.0+cuda` and the `ludus_renderer_gitlab` source/index, add `slangpy>=0.7.0`, `torchvision>=0.22`, and the previously-transitive deps (`scipy`, `pandas`, `pyarrow`, `imageio`).
* `.github/workflows/ci.yml`: drop the `--no-install-package ludus-renderer` workaround.
* `uv.lock`: regenerated; `slangpy` 0.41.0 resolved from PyPI.

Also adds `integrations/alpadreams/scripts/render_hdmap_preview.py` which renders an HD-map preview MP4 + first-frame PNG from a CLIPGT scene without standing up the gRPC server, useful for spot-checking the rasterizer and visual parity vs the old ludus path.

Known parity gaps (unchanged from migration plan):
* MSAA: ludus path ran 4x; the Slang software rasterizer ships without AA in v1. Add 2x supersample if diffusion output regresses.
* bbox tracks / dynamic geometry: not ported (default flashdreams config doesn't use them).
* Performance: compute-shader raster is likely slower than CudaRaster; benchmark on representative batch sizes before tuning.

Made-with: Cursor

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,10 +23,8 @@ jobs:
           # ty needs third-party packages to resolve imports. Skip packages
           # that cannot be installed on CPU-only runners:
           # - transformer-engine-torch: source-only, requires CUDA to compile
-          # - ludus-renderer: unavailable in CI
           uv sync --extra dev --group lint \
-            --no-install-package transformer-engine-torch \
-            --no-install-package ludus-renderer
+            --no-install-package transformer-engine-torch
 
       - name: Run linter checks
         run: uv run --no-sync pre-commit run -a
--- a/integrations/alpadreams/alpadreams/conditioning/renderer.py
+++ b/integrations/alpadreams/alpadreams/conditioning/renderer.py
@@ -13,78 +13,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Ludus-based HD map renderer.
+"""SlangPy-based HD-map renderer.
 
-This module provides the LudusRenderer class, which wraps the ludus_renderer
-library to render HD map scenes for conditioning video generation.
+This module provides the :class:`LudusRenderer` class, which renders HD-map
+conditioning frames for the Alpadreams video model. The class name is
+preserved for backwards compatibility, but the implementation no longer
+depends on the closed-source ``ludus-renderer`` wheel: it now uses a SlangPy
+software rasterizer (see :mod:`alpadreams.conditioning.slang_renderer`) that
+runs custom Slang compute kernels under PyTorch's CUDA context.
+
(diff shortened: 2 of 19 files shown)


liruilong940607:
Closing since the cuda and GL versions are all merged.

PR #62, opened 2026-05-11
Move bidirectional wan21 into integrations as a standalone repo

liruilong940607:
## Summary

- **Move bidirectional Wan 2.1 (T2V + I2V) out of `flashdreams/` core into its own `integrations/wan21/` plugin**, matching the layout of `self_forcing`, `causal_forcing`, and `fastvideo_causal_wan22` — with its own `pyproject.toml`, runner/config, README, bundled prompt + image assets, smoke tests, and `flashdreams.runner_configs` entry points for the `wan21-t2v-1.3b-480p` / `wan21-i2v-14b-480p` slugs.

- **Add a consolidated `flashdreams.recipes.wan.__init__` re-export surface** (`WanInferencePipeline*`, `Wan21Transformer*`, `WanDiTNetwork*`, `WanVAE*`, `WanI2VCtrlEncoderConfig`, …) so plugins import from one public path instead of deep submodules.

- **Rename `I2VCtrlEncoderConfig` → `WanI2VCtrlEncoderConfig`** for naming consistency with the rest of the `Wan*` public surface.

- **Relocate `fastvideo_causal_wan22`'s HF-diffusers checkpoint key-remap dictionary** out of `flashdreams.recipes.wan.transformer.wan22` and into the `fastvideo_causal_wan22` plugin alongside its checkpoint URL.

- **Add `RunnerConfig.offset_seed_by_global_rank: bool = True`** with the offset applied via `derive_config` in `Runner.__init__`, so multi-GPU runs draw distinct-but-reproducible RNG streams per rank without mutating the caller's literal config.


liruilong940607:
/ok to test [6510d35](https://github.com/NVIDIA/flashdreams/pull/62/commits/6510d35f30b1281a1804293a9871cc39406881c6)

PR #63, opened 2026-05-11
Use public HF mirrors for omni-dreams-reachable checkpoints

wlewNV:
## Summary

Switch the three checkpoints reachable from omni-dreams's `interactive-drive` demo to public Hugging Face URLs so external users can run it with just `HF_TOKEN`, no S3 credentials file.

- `lightvae` / `vae` / `lighttae` → `lightx2v/Autoencoders` (Apache-2.0, public)
- `1view-vae-chunk2` → `nvidia/omni-dreams-models/single_view/2b_res720p_30fps_i2v_hdmap_distilled.pt` (already mirrored from S3)

flashdreams's existing HF loader path handles `https://huggingface.co/...` URLs, so no loader changes — URL strings only.

Other checkpoints (multi-view, experimental single-view, opt-in chunk3/pshuffle) stay on S3 and are split out under a `# internal` comment.


flashdreams/flashdreams/core/io/hf_org.py
```diff
@@ -0,0 +1,66 @@
...
+    r"\bnvidia/omni-dreams-(models|scenes)\b"
+)
+
+
+def resolve_omni_dreams_hf_org() -> str:
+    """Return the configured omni-dreams HF org, defaulting to ``"nvidia"``."""
+    return os.environ.get(OMNI_DREAMS_HF_ORG_ENV_VAR, DEFAULT_OMNI_DREAMS_HF_ORG)
+
+
+def rewrite_omni_dreams_hf_url(url: str) -> str:
```
liruilong940607: any reason we need this? I thought the checkpoints are all under nvidia org HF. would be nice to reduce complexity if we dont need this.

wlewNV: This is specifically for the checkpoints for the Lighthouse adopters (https://huggingface.co/nvidia-omni-dreams-lha).

@jmccaffrey-nv customers will still be using the LHA and not "nvidia" mirror right?

jmccaffrey-nv: Yes, the EA uses [nvidia-omni-dreams-lha](https://huggingface.co/nvidia-omni-dreams-lha) HF repo, at least until the https://huggingface.co/nvidia/omni-dreams-models switches to public visibility . 

https://huggingface.co/nvidia-omni-dreams-lha separate HF Org enables private + external sharing during Early Access

flashdreams/flashdreams/recipes/alpadreams/config.py
```diff
@@ -63,8 +63,10 @@
 )
 
 AVAILABLE_ALPADREAMS_CHECKPOINT_PATHS: dict[str, str] = {
+    # public
+    "1view-vae-chunk2": "https://huggingface.co/nvidia/omni-dreams-models/resolve/main/single_view/2b_res720p_30fps_i2v_hdmap_distilled.pt",
+    # internal
```
liruilong940607: This change will force internal ppl to onboard this HF.

My proposal: set a global var `INTERNAL` and read it from system -- if it is set, then AVAILABLE_ALPADREAMS_CHECKPOINT_PATHS = {... internal paths ... } else {... external paths ... }

I think we can point the external path to LHA if that is the version we are currently shipping. then point back to the nvidia org before release. Does it sound reasonable?

jmccaffrey-nv: Hi @liruilong940607 , so you propose if INTERNAL: control which dict AVAILABLE_ALPADREAMS_CHECKPOINT_PATHS gets assigned value of ? 

 Proposal seems OK, if default to INTERNAL=false .

externals paths dict may have fewer entries than internal paths, if there are unreleased checkpoints .

wlewNV: Newest change:
Internal uses only have to include `FLASHDREAMS_INTERNAL_STORAGE=1` to take the old path. I chose STORAGE over CHECKPOINTS because it consists of both data + checkpoints.

liruilong940607:
/ok to test [a12dff8](https://github.com/NVIDIA/flashdreams/pull/63/commits/a12dff83fd915affb3730bf9d4c523a325f73093)

liruilong940607 (approved):
LGTM! I also triggered bot review -- feel free only pick up those make sense to you and self-resolve for the bot reviews.

liruilong940607:
Btw for lint fix:

```bash
uv sync --extra dev --group lint \
    --no-install-package transformer-engine-torch \
    --no-install-package ludus-renderer

uv run --no-sync pre-commit run -a
```

flashdreams/flashdreams/recipes/alpadreams/runner.py
```diff
@@ -202,11 +260,19 @@ def run(self) -> None:
...
-        _ensure_example_data_synced(is_rank_zero=self.is_rank_zero)
-        hdmap, first = _example_data_paths(num_views)
+        if not use_internal_storage() and num_views == 1:
+            hdmap, first = _ensure_hf_single_view_example_data_synced(
+                cfg.example_data_uuid
+            )
+        else:
+            hdmap, first = _ensure_s3_example_data_synced(
+                num_views, is_rank_zero=self.is_rank_zero
+            )
```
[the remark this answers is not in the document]
wlewNV: The huggingface_hub API doesn't have a direct subtree-only list_repo_files, you have to list all and filter, which is what we do. The dataset is 131 files (~30 KB response), trivial.

PR #44, opened 2026-05-04
build: vendor ludus-renderer v0.8.6 as workspace source package

janickm:
## Summary

- Integrate ludus-renderer source from `gitlab-master.nvidia.com/jseo/ludus-renderer` (commit `56538980`, v0.8.6) as a workspace member at `integrations/alpadreams/ludus-renderer/`
- Replaces the external PyPI dependency (previously hosted on internal GitLab/Artifactory registries unreachable from GitHub Actions runners) with an in-tree package that requires no network access to resolve
- Replace NVIDIA proprietary license headers with Apache-2.0 SPDX headers across all Python and C++/CUDA files

## Changes

| Area | What changed |
|------|-------------|
| `integrations/alpadreams/ludus-renderer/` | New workspace member — package source, tests, examples, README with provenance note |
| `integrations/alpadreams/pyproject.toml` | `ludus-renderer = { workspace = true }`, added `imageio` as direct dep (was transitive before) |
| `pyproject.toml` (root) | Added workspace member, extra-paths for ty/pyright, excluded ludus-renderer from ty (vendored, not yet type-clean) |
| `.github/workflows/ci.yml` | Updated comment — ludus-renderer excluded because JIT needs CUDA |
| `uv.lock` | Updated, no external registry references |

## Notes

- ludus-renderer ships pure Python + bundled C++/CUDA sources that are JIT-compiled at runtime via `torch.utils.cpp_extension` — it remains excluded from CPU lint runners (`--no-install-package ludus-renderer`)
- All 8 GPU-dependent tests are marked `@pytest.mark.manual` per flashdreams convention
- `imageio` was previously a transitive dependency via ludus-renderer v0.5.0; v0.8.6 dropped it, so it's now declared directly by flash-alpadreams which imports it

PR #65, opened 2026-05-11
replace nvdiffrast with cudaraster, adapt and upgrade

pknowlesnv:
- Replace the prior cudaraster implementation with the upstream BSD-3 Ludus renderer codebase, including the framework/runtime pieces it depends on, plus license and third-party notices.
- Rework the integration around CUDA-only rasterization: remove old GL-specific bindings, add static kernel compilation and host-callable wrappers, and adapt the host API to use external vertex/index buffers from the PyTorch side.
- Port the kernel code to modern CUDA/Linux toolchains by fixing framework/header compatibility, switching warp intrinsics to Volta+ _sync variants, and fixing ITS-sensitive race conditions in CoarseRaster and FineRaster.
- Add a broad validation layer: clean-room contract tests, Python API tests, and focused CUDA invariant tests for bin arbitration, ROP lane masks, and warp-scheduling-sensitive behavior.


.cursor/notes/cudaraster-decisions.md
liruilong940607, line 1: Are these notes intended for checking in? I suggest we do not check in them

liruilong940607:
Is this PR ready -- i.e. do we have it tested?

+ @janickm @jatentaki for vis.

.cursor/notes/cudaraster-decisions.md
pknowlesnv: Oops. Good call. Ty.

pknowlesnv:
There are two unit tests in the PR itself. The main one is failing because there are still broken features, but it seems to render hdmap images OK. I've been testing directly with this:

```
uv run python examples/render_hdmap_scene.py --scene example_data/clipgt_scene --frame 0 --cuda
```

integrations/alpadreams/ludus-renderer/ludus_renderer/_cpp/cudaraster/CudaRaster.cpp
```diff
@@ -1,96 +1,1737 @@
...
 }
 
+//------------------------------------------------------------------------
+
 void CudaRaster::setViewport(int width, int height, int offsetX, int offsetY)
 {
-    m_impl->setViewport(Vec2i(width, height), Vec2i(offsetX, offsetY));
+    m_viewportWidth = width;
+    m_viewportHeight = height;
+    m_viewportOffsetX = offsetX;
```
jmccaffrey-nv, line 215: Claude review flagged that m_viewportOffsetX and m_viewportOffsetY are not read: 


● #3 — Viewport offset is dropped

  What the API claims

  CudaRaster.hpp:64 declares:
  void setViewport(int width, int height, int offsetX, int offsetY);

  What actually happens

  CudaRaster.cpp:190-196:
  void CudaRaster::setViewport(int width, int height, int offsetX, int offsetY) {
      m_viewportWidth   = width;
      m_viewportHeight  = height;
      m_viewportOffsetX = offsetX;     // stored
      m_viewportOffsetY = offsetY;     // stored
  }

  drawTriangles then reads m_viewportWidth / m_viewportHeight and propagates them as params.viewportWidth/Height (CudaRaster.cpp:273-274, 375-376), but
  m_viewportOffsetX/Y are never read anywhere after they're stored — confirmed by grep:

  CudaRaster.hpp:157   int m_viewportOffsetX;
  CudaRaster.hpp:158   int m_viewportOffsetY;
  CudaRaster.cpp:71    m_viewportOffsetX(0),
  CudaRaster.cpp:72    m_viewportOffsetY(0),
  CudaRaster.cpp:194   m_viewportOffsetX = offsetX;
  CudaRaster.cpp:195   m_viewportOffsetY = offsetY;

  There's no params.viewportOffset* field in cuda/PrivateDefs.hpp:CRParams and no m_viewportOffset* read in any kernel — so the rasterizer cannot place geometry
  anywhere except (0, 0) regardless of what the caller passes.

  Why this matters in practice

  render/ludus_cuda.cu:2198-2208 (and a near-identical block at 2398-2408):
  int tilesX = (crWidth  + maxVp - 1) / maxVp;
  int tilesY = (crHeight + maxVp - 1) / maxVp;
  for (int ty = 0; ty < tilesY; ty++) {
      for (int tx = 0; tx < tilesX; tx++) {
          int vpW = (tx < tilesX - 1) ? maxVp : (crWidth  - tx * maxVp);
          int vpH = (ty < tilesY - 1) ? maxVp : (crHeight - ty * maxVp);
          s.cr->setViewport(vpW, vpH, tx * maxVp, ty * maxVp);   // offset = tile origin
          s.cr->drawTriangles(nullptr, false, stream);
      }
  }

  The main app's tile loop is the primary use of the offset arg — the audit at test_cudaraster_api.py:41 calls this out:

  ▎ setViewport(...)  # MULTIPLE per frame (nested tiles)

  The hdmap that the developer says is working probably fits in a single maxVp × maxVp tile (tilesX == tilesY == 1 → offset (0,0)). As soon as a target needs more
  than one tile, every tile after (0,0) will composite on top of the first one, garbage everywhere outside the first tile's footprint, no coverage where the later
  tiles should have landed.

  Test coverage that will fail

  There are two relevant tests in test_cudaraster_api.py:

  1. test_viewport_offset_shifts_output (line 893, plain test — expected to pass). Draws the same NDC triangle once at offset (0,0) and once at viewport (96, 96,
  32, 32), then asserts _pixel(shifted, 20, 20) == 0 — i.e., the shifted draw must leave the (20,20) pixel uncovered because the viewport starts at (32, 32). With
  the offset dropped, the shifted draw still writes at (20, 20), this fails.
  2. test_tiled_render_matches_single_view_interior (line 852, marked xfail(strict=True)). Renders the same scene as 1×1 vs 2×2 tiles and asserts pixel-equal
  output. Currently expected to fail with xfail(strict=True) — but if the offset is broken, the failure mode will be huge swathes of pixels diverging, not just
  "tile-boundary pixels dropped." When the offset is plumbed through, the xfail can be checked against actual interior-pixel equality.

  What the fix needs to touch

  The plumbing is short but spans every stage:

  1. cuda/PrivateDefs.hpp CRParams: add S32 viewportOffsetX, viewportOffsetY next to viewportWidth/Height.
  2. CudaRaster.cpp:380ish (drawTriangles): set params.viewportOffsetX = m_viewportOffsetX; params.viewportOffsetY = m_viewportOffsetY;.
  3. cuda/TriangleSetup.inl: the screen-space pixel coordinate of each vertex is computed from viewScale = Vec2f(viewportWidth/2, viewportHeight/2) * (1 <<
  CR_SUBPIXEL_LOG2) and a center-of-viewport origin. The offset needs to bias the destination pixel coordinates (the per-fragment output address into s_colorBuffer
   / s_depthBuffer), not the projection — i.e. it shifts where the rasterized tile lands in the framebuffer, not where the geometry is in NDC. Easiest place is in
  FineRaster.inl where pixelX, pixelY are computed from the per-tile origin and the per-fragment pixelInTile: add + params.viewportOffsetX / +
  params.viewportOffsetY before writing to s_colorBuffer. Worth confirming against upstream behavior — the original CudaRaster did support this because
  glViewport-style cropping is fundamental, so the right place is probably already obvious in m_sizeTiles / widthPixels accounting (which currently only cover
  viewport size, not framebuffer position).
  4. The deferred clear at CudaRaster.cpp:288-294 clears the entire framebuffer, not just the viewport. Per test_cudaraster_api.py:435-442, the harness does
  set_viewport(...) then draw_triangles(...) per tile, but deferredClear runs only on the first draw. That matches the desired behavior — the clear hits all
  pixels, then per-tile draws write only inside their viewport. So #4 is already correct; only the per-fragment write address needs the offset.

  A spec test to add: render a 1-pixel-wide triangle at NDC (0,0) inside a (W, H, ofsX, ofsY) viewport for several (ofsX, ofsY) values, assert the lit pixel is at
  (ofsX + W/2, ofsY + H/2). Today this would fail loudly for any non-zero offset.

integrations/alpadreams/ludus-renderer/ludus_renderer/_cpp/cudaraster/cuda/CoarseRaster.inl
```diff
@@ -333,33 +396,34 @@ __device__ __inline__ void coarseRasterImpl(const CRParams p)
...
-                                        {
-                                            U32 res = __ballot_sync(xyMask, b01 >= 0 && b02 >= 0 && b12 >= 0);
-                                            if (threadIdx.x == 31 - __clz(xyMask))
-                                                *(U32*)currPtr = res;
-                                            currPtr += 4, b01 -= d01y, b02 += d02y, b12 -= d12y;
-                                        }
-                                    }
-                                    currPtr += ptrYInc, b01 += d01x, b02 -= d02x, b12 += d12x;
+                                    if (x < lox) continue;
+                                    U32 ballot = __ballot_sync(__activemask(), b01 >= 0 && b02 >= 0 && b12 >= 0);
```
jmccaffrey-nv, line 411: Claude review flagged __activemask() usage in the Volta+ GPU upgrade . 

  #4 — __activemask() policy

  Why this matters

  CUDA's programming guide explicitly warns against __activemask() as the mask argument:

  ▎ Incorrect usage of __activemask() (e.g., assuming pre-Volta lockstep) is the most common source of subtle, intermittent bugs after porting Kepler-era warp code
  ▎  to Volta+. The reason: __activemask() is a snapshot of whichever lanes happen to be co-issued at that exact instruction — it does not match algorithmic
  ▎ intent.

  The cleanroom port's commit 0066475 flipped from explicit 0xFFFFFFFFu (which encodes "I assert the full warp is converged here") to __activemask() (which encodes
   "I'll trust whatever the scheduler hands me"). The decisions log shows the developer originally chose 0xFFFFFFFFu in commit 7489555 and then flipped policy. The
   very next commit (12c5f96) then fixes two ITS races in code that already uses __activemask(). That's the textbook failure mode the warning is about:
  __activemask() returned a partial mask, the kernel happily ran on a subset of lanes and produced a fragmented result, and the bug was visible only as occasional
  wrong pixels.

  Three classes of sites in this code, with examples

  I count 22 call sites using __activemask() as the mask argument. They split into three risk classes:

  Class 1: safe — predicate read at a convergence point.

  CoarseRaster.inl:308:
  if (__any_sync(__activemask(), triIdx != -1))
  { ... }
  At this line, all lanes are converged (no preceding divergent branch in the inner loop body that wouldn't have re-converged at the __syncthreads above).
  __activemask() returns 0xFFFFFFFFu, so the call is effectively the same as the explicit mask. Safe, but the explicit constant would be cheaper and clearer.

  Class 2: ambiguous — the algorithm wants "lanes with valid work," the mask says "lanes the scheduler co-issued."

  CoarseRaster.inl:400:
  if (__any_sync(__activemask(), warea * 4 <= area * 8))
  {
      if (triIdx != -1)
      { /* divergent body */ }
  }
  The predicate warea * 4 <= area * 8 is evaluated for lanes with triIdx == -1 too, where warea / area are garbage. __activemask() includes those garbage lanes in
  the vote — so the Case B branch can be taken on a vote that includes uninitialized data. (Note: this is inherited from upstream, not a porting bug, but it's the
  kind of thing __activemask() masks instead of surfacing.) Better would be:
  unsigned mask = __ballot_sync(0xFFFFFFFFu, triIdx != -1);
  if (__any_sync(mask, warea * 4 <= area * 8)) { ... }

  Class 3: actively risky — __activemask() inside a divergent loop body.

  CoarseRaster.inl:408-426 (the Case B inner loop):
  for (int y = wloy; y <= hiy; y++)

  CoarseRaster.inl:400:
  if (__any_sync(__activemask(), warea * 4 <= area * 8))
  {
      if (triIdx != -1)
      { /* divergent body */ }
  }
  The predicate warea * 4 <= area * 8 is evaluated for lanes with triIdx == -1 too, where warea / area are garbage. __activemask() includes those garbage lanes in the vote — so the Case B branch can be taken on a vote that includes uninitialized data. (Note: this is
  inherited from upstream, not a porting bug, but it's the kind of thing __activemask() masks instead of surfacing.) Better would be:
  unsigned mask = __ballot_sync(0xFFFFFFFFu, triIdx != -1);
  if (__any_sync(mask, warea * 4 <= area * 8)) { ... }

  Class 3: actively risky — __activemask() inside a divergent loop body.

  CoarseRaster.inl:408-426 (the Case B inner loop):
  for (int y = wloy; y <= hiy; y++)
  {
      if (y < loy) continue;
      for (int x = wlox; x <= hix; x++)
      {
          if (x < lox) continue;
          U32 ballot = __ballot_sync(__activemask(), b01 >= 0 && b02 >= 0 && b12 >= 0);
          atomicOr((U32*)currPtr, ballot);
          ...
      }
  }
  hix, hiy, lox, loy are per-lane (each triangle's AABB), so the loop trip count varies across lanes. The if (y < loy) continue skip is also per-lane. By the time we reach the __ballot_sync, the warp has fragmented into subgroups by trip count and skip pattern.
  __activemask() returns whichever subgroup is co-issued right now, not the lanes that needed to vote on this tile cell. Without the atomicOr fix from commit 12c5f96, multiple subgroups would race on the same currPtr. With the fix, the atomicOr composes the partial
  ballots — but the fact that this site needed an atomic at all is the same evidence that __activemask() is not the mask the algorithm wants.

  FineRaster.inl:464-466 (the ROP fix from 12c5f96):
  U32 activeMask = __activemask();
  atomicMin((U32*)pDepth, depth);
  __syncwarp(activeMask);
  if (depth == *pDepth) { ... write color ... }
  Same pattern, same risk. activeMask is "whatever lanes are co-issued at this instant" — but the algorithm wants "lanes that have a real fragment to deposit." The if (ropLaneIdx < fragWrite - fragRead) gate two scopes up partially serves this purpose, but inside that
   gate there's further per-lane divergence on (covered && !zkill) && !discard. If two lanes with the same min depth are co-issued in one subgroup and a third winning lane is co-issued in another, the syncwarp on activeMask won't see the third lane — its color write
  may happen before or after the other two, undefined order.

  What I'd recommend

  A single rule, written down at the top of cuda/Util.hpp (or cuda/WarpSync.hpp if you add one), enforced by review:

  ▎ __activemask() is forbidden as a mask argument to _sync intrinsics. Compute the mask from algorithmic intent. The three legal patterns are:
  ▎
  ▎ 1. Full-warp convergence asserted by control flow. Use 0xFFFFFFFFu. Add a comment naming the convergence point ("all 32 lanes converge here at __syncthreads() on line N"). If you can't name it, you can't use this pattern.
  ▎ 2. Predicate-derived participation. Build the mask explicitly:
  ▎ unsigned mask = __ballot_sync(0xFFFFFFFFu, hasWork);
  ▎ if (hasWork) { __any_sync(mask, ...); }
  ▎ 2. The outer __ballot_sync uses 0xFFFFFFFFu (legal by pattern 1; called at a convergence point) and gives every subsequent _sync the exact mask that matches the algorithm's "lanes that have work to do."
  ▎ 3. Loop-arrival mask. When the warp fragments inside a divergent loop and you need cross-lane work within the subgroup that arrived, derive the mask once before the loop with __ballot_sync(0xFFFFFFFFu, ...) over the loop entry predicate. Inside the loop body, the
  ▎ lanes that arrived may further diverge; if you need cross-lane communication within the body, you either (a) re-converge with __syncwarp(mask) then act, or (b) use an atomic (as Case B Fix does after 12c5f96).

  Apply that rule to the 22 sites I listed and the picture would be:

  - All BinRaster.inl _sync calls in the per-warp prefix-sum scan (lines 113-117): these are inside if (myThr < CR_BIN_WARPS) blocks where the warp is converged — should be 0xFFFFFFFFu per pattern 1.
  - BinRaster.inl:246, 258, 266, 346: inside divergent loops over triIdx. These look like pattern 2 sites — there's an outer if (triIdx != -1) gate that should drive an explicit mask.
  - CoarseRaster.inl:308, 339, 400, 496: pattern 1 candidates (all converged at the surrounding __syncthreads()).
  - CoarseRaster.inl:411 (the Case B inner loop): genuinely divergent; needs pattern 3 + atomicOr (which is what 12c5f96 does — but the mask should still come from explicit ballot of the loop-arrival predicate, not __activemask()).
  - FineRaster.inl:178, 514, 866, 869: these are all in clearly converged scopes; 0xFFFFFFFFu.
  - FineRaster.inl:657, 688, 1006, 1037: inside the ROP path, after if (ropLaneIdx < fragWrite - fragRead) ... if ((covered && !zkill)) — pattern 2 candidates.
  - FineRaster.inl:464: the ROP fix from 12c5f96. The current activeMask should be __ballot_sync(0xFFFFFFFFu, /* same predicate that gated entry to this branch */).
  - Util.hpp:94 singleLane(): this is an inline helper called from many places. The right answer is to delete this helper and force every call site to derive its own mask, because "single lane" is meaningful only relative to a known group of participating lanes, which
   the helper can't know.

integrations/alpadreams/ludus-renderer/ludus_renderer/_cpp/cudaraster/CudaRaster.cpp
```diff
@@ -1,96 +1,1737 @@
...
-    m_impl->setTiebreakerColorBuffer(colors);
+    m_tiebreakerColors = colors;
 }
 
+//------------------------------------------------------------------------
+
 void CudaRaster::setDeterministicTiebreaker(bool enable)
 {
-    m_impl->setDeterministicTiebreaker(enable);
+    m_deterministicTiebreaker = enable;
```
jmccaffrey-nv, line 262: Claude review flagged CudaRaster::setDeterministicTieBreaker and other setters for not being plumbed-through functionally : 

  Functional gaps in the documented API surface

  The PR adds setters for several knobs but doesn't actually plumb them to the kernels. Each is silent — no error, no warning — so a caller exercising the documented API gets behavior that doesn't match the contract.

  ranges parameter is stored and dropped

  // CudaRaster.cpp:267
  bool CudaRaster::drawTriangles(const int32_t* ranges, bool peel, cudaStream_t stream) {
      m_ranges = ranges;       // stored
      ...                      // never read again, anywhere
  }
  Grep confirms: m_ranges has exactly one write site (line 269) and zero read sites. The wrapper in torch_rasterize_cuda.cpp validates [N, 2] int32, N == numImages and holds the tensor alive — all that work is for nothing. The audit at test_cudaraster_api.py:42 says
  the main app calls drawTriangles(nullptr, false, stream) so this doesn't hurt production, but ~3 contract tests for ranges in test_cudaraster_api.py will silently produce wrong output.

  m_tiebreakerColors is stored and dropped

  // CudaRaster.cpp:232
  void CudaRaster::setTiebreakerColorBuffer(const uint32_t* colors) {
      m_tiebreakerColors = colors;     // stored
  }
  // never copied into CRParams, never accessed by any kernel
  Same situation: zero read sites. The audit says the main app always sets setTiebreakerColorBuffer(...) per frame (line 38). This means the main app's tiebreaker behavior is whatever the kernel happens to do without the buffer, not what the API promises.

  m_deterministicTiebreaker is stored and dropped

  // CudaRaster.cpp:239
  void CudaRaster::setDeterministicTiebreaker(bool enable) {
      m_deterministicTiebreaker = enable;     // stored, never read
  }
  The audit says the main app sets this to true every frame. The tiebreaker-determinism tests in test_cudaraster_api.py will fail or pass non-deterministically depending on what the kernel does without the flag.

  swapDepthAndPeel is effectively a no-op

  // CudaRaster.cpp:445
  void CudaRaster::swapDepthAndPeel(void) {
      uint32_t* tmp = m_depthBufferRaw;
      m_depthBufferRaw = m_peelBufferRaw;
      m_peelBufferRaw = tmp;
  }
  Only the linear readback pointers swap. m_depthSurfaceObj and the underlying m_depthArray don't move; m_peelArray is allocated but no kernel ever writes to it. Subsequent drawTriangles still rasterizes into the same physical depth surface. Depth peeling is broken
  end-to-end. The audit says depth peeling isn't used in production, so this isn't urgent — but the test suite has 4 depth-peeling tests + 4 xfail markers around peeling that will be confusing once anyone runs them against the new code.

  numImages > 1 is allocation-only

  setBufferSize(W, H, N) allocates W × (H × N) buffers but no kernel knows about the image dimension. There's no imageIdx in CRParams, no per-image branch in TriangleSetup.inl/FineRaster.inl. With numImages = 2, only rows 0..H-1 get written and rows H..2H-1 stay as
  the clear value. Combined with the dropped ranges parameter, multi-image rendering is fundamentally broken. Again, the audit says the main app uses numImages = 1, but the contract tests have multi-image cases.

  RenderModeFlag_EnableDepthPeeling is silently dropped

  The pipeline is instantiated with a fixed render-mode set in CudaRasterKernels.cu:49-55:
  CR_DEFINE_PIXEL_PIPE(
      crDefaultPipe,
      FW::ShadedVertexBase,
      FW::TriIdShader,
      FW::BlendReplace,
      0,
      FW::RenderModeFlag_EnableDepth)   // hardcoded
  The runtime setRenderModeFlags() only feeds params.enableBackfaceCulling (see CudaRaster.cpp:389). Setting RenderModeFlag_EnableDepthPeeling at runtime doesn't change the pipe template — it's a compile-time template parameter. The flag is silently dropped. Should
  either remove the flag from the public enum or compile multiple pipe instantiations.

  Recommendation for this whole cluster: either plumb each setter through to the kernel, or NVDR_CHECK-style fail at the setter when the not-yet-wired path is exercised. Silent no-ops in a contract-test-driven port are worst-of-both — you don't get the feature and you
   don't get a clear signal it's missing.

jmccaffrey-nv (commented):
I commented on three medium issues from claude review : functionality of m_viewportOffsetX and Y, CudaRaster::m_deterministicTiebreaker  and other setters, and correctness with __activemask() usage .

I think the PR is OK to merge, but has some correctness concerns with these open

.cursor/notes/cudaraster-decisions.md
janickm: resolving as removed, thx @pknowlesnv

janickm:
[the target branch #44 is quite outdated, trying to rebase both]

janickm:
rebased target #44 on `main` + rebased this PR, also applied formatting

janickm:
fixed (or ignored..) static type error checks, making `uv run pre-commit run -a` pass

integrations/alpadreams/ludus-renderer/tests/conftest.py
janickm, line 1: @pknowlesnv on my local workstation (RTX A6000, 595.58.03) I still have the following failing test cases

```
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_tiebreaker_uses_colors_not_triangle_order - assert 1 != 1
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_tiebreaker_enabled_without_color_buffer_is_deterministic - AssertionError: winner changed; if the new winner is documented, update this marker
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_depth_peeling_two_layers_currently_crashes_with_cuda700 - AssertionError: two-layer peel now fails with a different error:
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_get_buffer_dims_report_tile_rounded_allocation - assert 101 == 104
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_large_triangle_count_runs_without_silent_corruption - AssertionError: 100k overlapping triangles must produce visible coverage
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_overflow_currently_crashes_with_cuda700 - AssertionError: overflow now fails with a different error:
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_ranges_draw_first_triangle_only - assert 2 == 0
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_ranges_draw_tail_triangles_only - assert 1 == 0
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_depth_peeling_three_layers_currently_crashes_with_cuda700 - AssertionError: three-layer peel now fails with a different error:
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_depth_peeling_single_layer_currently_crashes_with_cuda700 - AssertionError: single-layer peel now fails with a different error:
FAILED integrations/alpadreams/ludus-renderer/tests/test_cudaraster_api.py::test_single_draw_multi_image_ranges_produce_distinct_outputs - assert 2 == 0
======================================================================================= 11 failed, 86 passed, 5 xfailed in 80.04s (0:01:20) ========================================================================================
```

11 test fail (+ 5 known to fail) - let me know if you need more info / more elaborate logs

integrations/alpadreams/ludus-renderer/README.md
janickm, line 114: @pknowlesnv  Running the code as

```
uv run python integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py \
  --scene {test data} \
  --frame 12 --cuda
```

from the repo root yields a fully black image frame, wheres the base version in #44 (prior to your refactoring) is functional, so there is a regression - the refactoring / changes are just too large for me to know where the issue is, can you please dig into this?

[I test with data from this maglev WF, trying to attach it here for your reference: [98554693-d510-40cd-b442-eef9e4102b88.zip](https://github.com/user-attachments/files/27629045/98554693-d510-40cd-b442-eef9e4102b88.zip)].

Moreover, running the same rendering example _without_ the `--cuda` flag fails completely with 

```
Render backend: OpenGL (GL_NV_mesh_shader)
Traceback (most recent call last):
  File "/home/janickm/Code/git/sil/flashdreams/integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py", line 1484, in <module>
    main()
  File "/home/janickm/Code/git/sil/flashdreams/integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py", line 1480, in main
    render_single_frame(args)
  File "/home/janickm/Code/git/sil/flashdreams/integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py", line 125, in render_single_frame
    ctx = _create_context(args, device)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/janickm/Code/git/sil/flashdreams/integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py", line 84, in _create_context
    ctx = LudusTimestampedContext(device=device)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/janickm/Code/git/sil/flashdreams/integrations/alpadreams/ludus-renderer/ludus_renderer/_ops/context.py", line 607, in __init__
    self.cpp_wrapper = _get_plugin(gl=True).LudusTimestampedStateWrapper(
                       ^^^^^^^^^^^^^^^^^^^^
  File "/home/janickm/Code/git/sil/flashdreams/integrations/alpadreams/ludus-renderer/ludus_renderer/_ops/_plugin.py", line 163, in _get_plugin
    _cached_plugin[gl] = torch.utils.cpp_extension.load(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/janickm/Code/git/sil/flashdreams/.venv/lib/python3.12/site-packages/torch/utils/cpp_extension.py", line 1818, in load
    return _jit_compile(
           ^^^^^^^^^^^^^
  File "/home/janickm/Code/git/sil/flashdreams/.venv/lib/python3.12/site-packages/torch/utils/cpp_extension.py", line 2218, in _jit_compile
    version = JIT_EXTENSION_VERSIONER.bump_version_if_changed(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/janickm/Code/git/sil/flashdreams/.venv/lib/python3.12/site-packages/torch/utils/_cpp_extension_versioner.py", line 49, in bump_version_if_changed
    hash_value = hash_source_files(hash_value, source_files)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/janickm/Code/git/sil/flashdreams/.venv/lib/python3.12/site-packages/torch/utils/_cpp_extension_versioner.py", line 16, in hash_source_files
    with open(filename, "rb") as file:
         ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/janickm/Code/git/sil/flashdreams/integrations/alpadreams/ludus-renderer/ludus_renderer/_ops/../_cpp/common/glutil.cpp'
```

but I guess this is the version you were trying to replace anyway? Let's please also update / remove all of the code paths properly / update all docs for a clean and self-contained version [but I guess highest priority is to fix the cuda-rendering]

janickm (changes requested):
Looks like there are regressions with the cuda-based rendering of hdmap scenes after the refactoring and requires some more love [see comments for details] - for now the PR doesn't look fully functional yet (FYI @liruilong940607)

integrations/alpadreams/ludus-renderer/README.md
janickm: I also tested execution from within the package dir with `uv run python examples/render_hdmap_scene.py ...` but same result

janickm: Although my tests using `examples/render_hdmap_scene.py` are negative, running the same command as in https://github.com/NVIDIA/flashdreams/pull/44#issuecomment-4430473259 on these changes seems to work [I might be using outdated (i.e., previous) package versions though, still checking] - maybe you can do a full end-to-end check given these references on your end also @pknowlesnv?

janickm: I cleaned my jit caches and end-to-end results still look ok - maybe I had a bad setup earlier when I saw the issues - I suggest you also verify on your end?

janickm: [But rendering via `render_hdmap_scene.py` in the `--cuda` mode is still non-functional for me..]

integrations/alpadreams/ludus-renderer/tests/conftest.py
pknowlesnv: These are known issues - people are aware but just care about hdmap rendering for now. I can skip them if we have automated tests.

janickm: you mean all are known issues? good to know - in that case I'd mark then as xfailed (known to be failed) [this way others can run locally and still expect everything to pass "clean" except for the explicitly marked ones)]

but I'm unsure if this is related to the incorrect rendering results I got via render_hdmap_scene.py

janickm: [as I said, it could also be a local problem to my system / setup, I'm not fully confident the current state isn't functional if things seem fine for you? but it'd still be good to double-check]

janickm:
Closing in favor of the rebased and integrated #76 which includes the functional cudaraster backend now also [@pknowlesnv - let's finish the review there]

PR #66, opened 2026-05-12
FlashVSR Integration

wilsonCernWq:
There should still be room for cleanup 🙏 Please comment.
There is one additional optimization missing that fuses bicubic interpolation into one kernel. I will introduce it in a separate PR

GB200
```
  {
    "autoregressive_index": 5,
    "pad_ms": 0.05548800155520439,
    "bicubic_ms": 1.284767985343933,
    "projector_ms": 25.4019832611084,
    "dit_concat_ms": 485.1327819824219,
    "denoise_ms": 0.021056000143289566,
    "decoder_ms": 97.21241760253906,
    "color_ms": 1.1745599508285522,
    "finalize_ms": 0.002176000038161874,
    "total_ms": 610.2852307839785,
    "total_ms_wo_finalize": 610.2830547839403,
    "mem_alloc_gib": 28.519792556762695,
    "mem_reserved_gib": 83.173828125,
    "mem_peak_gib": 56.22847890853882,
    "frames": 16,
    "fps": 26.217249235158846
  }
  ```

Parity Check:
```
uv run flashdreams-run flashvsr-v1.1-sparse-ratio-2.0 --input-path outputs/example4.mp4 --pipeline.encoder.scale 4 --chunk-size  8
```
Baseline
```json
  {
    "autoregressive_index": 201,
    "projector_ms": 23.62895965576172,
    "dit_ms": 520.8135375976562,
    "decoder_ms": 154.10342407226562,
    "color_ms": 1.6332800388336182,
    "total_ms": 700.1792013645172,
    "total_ms_wo_finalize": 700.1792013645172,
    "frames": 8,
    "fps": 11.42564644081045,
    "mem_alloc_gib": 5.247886657714844,
    "mem_reserved_gib": 51.453125,
    "mem_peak_gib": 44.09325456619263
  }
```

Ours
```json
  {
    "autoregressive_index": 203,
    "pad_ms": 0.03446400165557861,
    "bicubic_ms": 0.8677120208740234,
    "projector_ms": 16.711040496826172,
    "dit_concat_ms": 375.5096740722656,
    "denoise_ms": 0.016736000776290894,
    "decoder_ms": 63.91193771362305,
    "color_ms": 0.6891520023345947,
    "finalize_ms": 0.1563519984483719,
    "total_ms": 457.8970683068037,
    "total_ms_wo_finalize": 457.74071630835533,
    "mem_alloc_gib": 38.198251724243164,
    "mem_reserved_gib": 86.466796875,
    "mem_peak_gib": 74.15848636627197,
    "frames": 8,
    "fps": 17.471175409753396
  }
```


integrations/flashvsr/flashvsr/decoder/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+Wraps :class:`TAEHV` (in :mod:`.network`) + :class:`FlashVSRColorCorrector`
+for the ``flashdreams.infra.decoder.StreamingDecoder`` interface. One
+``forward()`` call ingests a chunk of clean latents plus the bicubic
+upres stashed by :class:`FlashVSRPipeline.generate` on
+``FlashVSRDecoderCache.last_upres``, and returns RGB frames in
+``[-1, 1]``. The upres threads through the cache rather than the
+standard ``input=`` arg because both the TC decoder (as ``cond``) and
+the color corrector (as the AdaIN reference) consume it. See
+``README.md`` (sibling) for the streaming chunk contract.
+"""
```
liruilong940607, line 27: I think this MR could benefit from a round of doc string cleaning up.

Could you try use our skill `agentic/skills/python-docstring-style` and ask agent to follow the rules in it?

integrations/flashvsr/flashvsr/decoder/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+from torch import Tensor
+
+from flashdreams.infra.config import InstantiateConfig
+from flashdreams.infra.decoder import StreamingDecoder, StreamingDecoderCache
+from flashdreams.infra.profiler import EventProfiler, record_event
+from flashvsr.corrector import (
+    ColorCorrectorImplementation,
+    FlashVSRColorCorrector,
+)
+from flashvsr.decoder.network import TAEHV, TAEHV_Cache
```
liruilong940607: Since it is a different TAEHV, maybe rename it to `FlashvsrTAEHV`?

flashdreams/flashdreams/recipes/wan/autoencoder/vae.py
```diff
@@ -153,6 +151,9 @@ def __init__(self, *args, **kwargs):
         self._spatial_pad = (pw, pw, ph, ph)
         self._has_spatial_pad = ph > 0 or pw > 0
         self._time_pad = 2 * self.padding[0]
+        # ``F.pad`` mode names: zero-pad is ``"constant"``, replicate is
+        # ``"replicate"`` (matches the user-facing ``pad_mode`` enum).
+        self._pad_mode = "replicate" if pad_mode == "replicate" else "constant"
```
liruilong940607: This is a bit error prone -- if user pass in `_pad_mode="something else"` there would be no error.

proposal:
assert  pad_mode in ["constant", "replicate"]
self._pad_mode = pad_mode

flashdreams/flashdreams/core/checkpoint/load.py
```diff
@@ -588,6 +675,12 @@ def load_checkpoint(
     Examples:
 
       >>> state = load_checkpoint("s3://bucket/foo.safetensors")
+      >>> state = load_checkpoint(
+      ...     "https://huggingface.co/foo/bar/resolve/main/model.pt"
+      ... )
+      >>> state = load_checkpoint(
+      ...     "https://raw.githubusercontent.com/org/repo/main/posi_prompt.pth"
```
liruilong940607: I think this is not a checkpoint technically -- it's a cached precomputed text prompt. I propose we use 

https://github.com/NVIDIA/flashdreams/blob/main/flashdreams/flashdreams/core/io/download.py#L39

to download the non-checkpoint files (I recently added this)

liruilong940607: And once we switch to that for this `posi_prompt.pth` file. all changes in this file can be reverted I think [for simplicity] except that adding `".ckpt"` to `(".pt", ".pth", ".ckpt", ".safetensors")`

integrations/flashvsr/tests/test_flashvsr.py
liruilong940607, line 1: I think those files are importing from the old path. Should move them the new destination.

`from flashdreams.recipes.flashvsr.config`

integrations/flashvsr/tests/test_dit_replacement.py
```diff
@@ -0,0 +1,316 @@
...
+
+_LEGACY_REF_PATH = Path(__file__).resolve().parent / "_wan_model_dit.py"
+_DEFAULT_WEIGHTS_ROOT = "~/.cache/flashdreams/upsampler/weights"
+_WEIGHTS_ROOT = Path(
+    os.environ.get("FLASHVSR_WEIGHTS_ROOT", _DEFAULT_WEIGHTS_ROOT)
+).expanduser()
+_MODEL_NAME = "FlashVSR-v1.1"
+_DIT_DIR = _WEIGHTS_ROOT / _MODEL_NAME / "flashvsr_tiny_long"
+_DIT_CFG = _DIT_DIR / "dit_config.json"
+_DIT_SD = _DIT_DIR / "dit_state_dict.pt"
```
liruilong940607: I'm not sure if those local paths will work in CI

flashdreams/flashdreams/recipes/taehv/impl.py
```diff
@@ -191,8 +176,27 @@ def forward(
         return x.reshape(b, bt // b, c_out, h_out, w_out)
 
 
+def _legacy_to_blocks_keys(
```
liruilong940607: Propose to put those logics besides the checkpoint path like this

https://github.com/NVIDIA/flashdreams/blob/main/integrations/fastvideo_causal_wan22/fastvideo_causal_wan22/config.py#L40-L73

each checkpoint might have its own remapping logic so we dont grow the code here.

flashdreams/flashdreams/recipes/taehv/impl.py
```diff
@@ -315,8 +326,50 @@ def __init__(
...
+        # ``nn.Module.__setattr__`` and register the same module a second time,
+        # duplicating every key in ``state_dict`` under ``_decoder_call.*``.
+        # ``object.__setattr__`` skips that and stores it as a plain attribute.
+        object.__setattr__(
+            self,
+            "_decoder_call",
+            self._decoder_wrapper
+            if self._decoder_wrapper is not None
+            else self.decoder,
+        )
```
liruilong940607: whats the reason we have to do this change? `object.__setattr__` is not quite elegant to me

flashdreams/flashdreams/recipes/taehv/impl.py
```diff
@@ -288,20 +304,15 @@ def __init__(
                 act_func=act_func,
             )
 
+        # Subclasses can mutate the module tree here (e.g., insert deepened
+        # layers) before any checkpoint is loaded.
+        self._post_model_init_hook()
```
liruilong940607: This hook here feels like a hidden magic -- i.e. when user read the flashvsr code, they would have to already open up this file to understand where this hook is placed.

An alternative solution that I propose: make `checkpoint_path` as optional here.

then you dont need this hook can just do something like

```
class FlashvsrTAEHV(TAEHV):

def __init__(self)
       super().__init__(checkpoint_path = None)
       hook()
       # load checkpoint here
       
```

flashdreams/flashdreams/recipes/taehv/impl.py
```diff
@@ -315,8 +326,50 @@ def __init__(
...
+        state_dict: Mapping[str, torch.Tensor],
+        strict: bool = True,
+        assign: bool = False,
+    ) -> torch.nn.modules.module._IncompatibleKeys:
+        """Route every state-dict load through ``_pre_load_state_dict_hook``."""
+        return super().load_state_dict(
+            self._pre_load_state_dict_hook(state_dict),
+            strict=strict,
+            assign=assign,
         )
```
liruilong940607: with the above changes i think we then dont need these maybe?

integrations/flashvsr/tests/_tcdecoder.py
```diff
@@ -0,0 +1,274 @@
...
+# limitations under the License.
+
+"""
+The original implementation from FlashVSR.
+
+Tiny AutoEncoder for Hunyuan Video (Decoder-only, pruned)
+- Encoder removed
+- Transplant/widening helpers removed
+- Deepening (IdentityConv2d+ReLU) is now built into the decoder structure itself
+"""
```
liruilong940607: If we check in their original code we should respect their license.

Otherwise we can do it via on-the-fly git clone like this https://github.com/NVIDIA/flashdreams/tree/main/integrations/self_forcing/tests/parity_check

integrations/flashvsr/README.md
```diff
@@ -0,0 +1,224 @@
...
+| `run_flashvsr.py` | Standalone argparse CLI; kept for backward compatibility with the pre-runner workflow. |
+
+## Tests
+
+CPU smoke + parity tests live under `flashdreams/tests/flashvsr/`
+(in-tree, not under this plugin) and run as part of the main test
+suite:
+
+```bash
+uv run --extra dev pytest flashdreams/tests/flashvsr -v
```
liruilong940607: Lets move tests to this folder

liruilong940607 (commented):
Thanks its in a pretty good shape. Just some minor structure suggestions

liruilong940607:
Lets also trigger bot review before merge but feel free to self resolve on those bot reviews that makes sense.

integrations/flashvsr/README.md
wilsonCernWq: The tests are not yet fixed, working on them still.

flashdreams/flashdreams/recipes/taehv/impl.py
wilsonCernWq: TAEHV is an nn.Module. PyTorch's nn.Module.__setattr__ inspects every assignment and, if the value is an nn.Module, registers it under self._modules[name]. That means it shows up in state_dict(), named_modules(), named_parameters(), gets device-moved by .to(), etc.

In the constructor, self.decoder = Decoder(...) (line 293) already registers decoder as a child. Then on line 325, the value being stored in _decoder_call is one of:

self._decoder_wrapper — a CUDAGraphWrapper, which is a plain Python class (not nn.Module), so a normal assignment would be harmless.

self.decoder — already a registered submodule. A normal assignment would register the same Decoder a second time under the name _decoder_call, doubling every entry in state_dict:

decoder.blocks.1.weight
_decoder_call.blocks.1.weight    # duplicate, points to the same tensor
That breaks load_state_dict (unexpected keys / double-loading on assign=True), inflates checkpoint dumps, and confuses anything that walks named_modules().

object.__setattr__ bypasses nn.Module.__setattr__ entirely and just sets the instance dict, so the reference is stored as a plain Python attribute with none of the bookkeeping side effects.

----------

But yeah, converted it to a @property ... that is cleaner

flashdreams/flashdreams/core/checkpoint/load.py
wilsonCernWq: Yes, much better idea, fixed

flashdreams/flashdreams/recipes/wan/autoencoder/vae.py
wilsonCernWq: Fixed

integrations/flashvsr/flashvsr/decoder/__init__.py
wilsonCernWq: Fixed

flashdreams/flashdreams/recipes/taehv/impl.py
wilsonCernWq: Sounds good, converted to state_dict_transform now.

flashdreams/flashdreams/recipes/taehv/impl.py
wilsonCernWq: Fixed

flashdreams/flashdreams/recipes/taehv/impl.py
wilsonCernWq: Yes, but I am still keeping a load_from_checkpoint function for child class to call it more conveniently

https://github.com/NVIDIA/flashdreams/pull/66/changes/394eadc723dd9454ceff11be02aed1b0fa0c4762#diff-385838c3fee4774f0dc0ec90522dd879c4e40182d49aa5c6e1d4897dfc9c4f38R304

integrations/flashvsr/tests/test_flashvsr.py
wilsonCernWq: fixed

integrations/flashvsr/flashvsr/decoder/__init__.py
wilsonCernWq: Did a path

integrations/flashvsr/tests/_tcdecoder.py
wilsonCernWq: Moved them into parity check now

pyproject.toml
```diff
@@ -27,6 +27,43 @@ override-dependencies = ["nvidia-cublas>=13.4"]
...
+# `extra-build-dependencies` above to succeed, and `match-runtime = true`
+# requires static metadata to function. Declare the metadata upfront,
+# mirroring the upstream `setup.py`'s `install_requires` and the `__version__`
+# in `block_sparse_attn/__init__.py` (the `.post1` git tag is an
+# `__init__.py` cleanup with no version bump), so uv skips the resolution-
+# time build and defers to install time, when the build env is fully wired.
+[[tool.uv.dependency-metadata]]
+name = "block-sparse-attn"
+version = "0.0.2"
+requires-dist = ["torch", "einops"]
```
liruilong940607, line 64: I think we can move block sparse attn dependency into `integration/flashvsr` since that the only place using it?

liruilong940607 (commented):
Only has one comment.

pyproject.toml
wilsonCernWq: If we do this, `uv sync` will not work it seems
```
Creating virtual environment at: /tmp/venv/flashdreams
  x Failed to build `block-sparse-attn @ git+https://github.com/mit-han-lab/Block-Sparse-Attention@v0.0.2.post1`
  |-> The build backend returned an error
  `-> Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit status: 1)

      [stderr]
      /root/.cache/uv/builds-v0/.tmp3oMHsA/lib/python3.12/site-packages/setuptools/_vendor/wheel/bdist_wheel.py:4: FutureWarning: The 'wheel' package is no longer the canonical location of the 'bdist_wheel' command, and will be removed in a future release. Please update to setuptools v70.1 or later which contains an integrated version of
      this command.
        warn(
      Traceback (most recent call last):
        File "<string>", line 14, in <module>
        File "/root/.cache/uv/builds-v0/.tmp3oMHsA/lib/python3.12/site-packages/setuptools/build_meta.py", line 333, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/root/.cache/uv/builds-v0/.tmp3oMHsA/lib/python3.12/site-packages/setuptools/build_meta.py", line 301, in _get_build_requires
          self.run_setup()
        File "/root/.cache/uv/builds-v0/.tmp3oMHsA/lib/python3.12/site-packages/setuptools/build_meta.py", line 520, in run_setup
          super().run_setup(setup_script=setup_script)
        File "/root/.cache/uv/builds-v0/.tmp3oMHsA/lib/python3.12/site-packages/setuptools/build_meta.py", line 317, in run_setup
          exec(code, locals())
        File "<string>", line 21, in <module>
      ModuleNotFoundError: No module named 'torch'

      hint: This error likely indicates that `block-sparse-attn` depends on `torch`, but doesn't declare it as a build dependency. If `block-sparse-attn` is a first-party package, consider adding `torch` to its `build-system.requires`. Otherwise, either add it to your `pyproject.toml` under:

      [tool.uv.extra-build-dependencies]
      block-sparse-attn = ["torch"]

      or `uv pip install torch` into the environment and re-run with `--no-build-isolation`.
```

These information has to be defined in the workspace where we run `uv sync`... A UV issue

.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     env:
       UV_PROJECT_ENVIRONMENT: /tmp/flashdreams-venv
       UV_LINK_MODE: copy
+      MAX_JOBS: 1
```
janickm: this might be a WAR to get around the build failures (we can likely increase this a bit further to like 16ish or even more), but the total build time is way to high for now if we need to re-build with every run. I'll look into how to cache these results better in our CI so that we can re-use previous build results

.github/workflows/ci.yml
```diff
@@ -76,7 +77,7 @@ jobs:
 
   gpu:
     runs-on: linux-amd64-gpu-rtxpro6000-latest-2
-    timeout-minutes: 30
+    # timeout-minutes: 30
```
janickm: same, we need to stay within 30min budget (trying with more jobs for now)

janickm:
Merged a couple of relevant CI updates in #121 - rebasing this one now

pyproject.toml
janickm: yeah I also just tried and it seems these dependency overwrites indeed need to live in the root workspace :-/ but given these overwrites here are version-associated, I'd say this is ok - this doesn't make `block-sparse-attn` a dependency of the "main" package either

pyproject.toml
```diff
@@ -33,6 +33,43 @@ override-dependencies = [
...
+# `psutil`, `ninja`). Instead, keep build isolation on and seed the isolated
+# build env explicitly; `match-runtime = true` on torch picks up the CUDA-
+# enabled torch resolved for the project venv, which is what the upstream
+# README's `python setup.py install` instructions implicitly assume.
+[tool.uv.extra-build-dependencies]
+block-sparse-attn = [
+    { requirement = "torch", match-runtime = true },
+    "setuptools",
+    "wheel",
+    "packaging",
```
janickm: [I simplified this a little / stripped it down in the last commit]

integrations/flashvsr/README.md
```diff
@@ -0,0 +1,259 @@
+# flashdreams-flashvsr
```
janickm, line 6: minor: we might be missing this in the rest of the repo, but we might still want to add copyright notes to `.md` files also, at least

```md
<!--
SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
SPDX-License-Identifier: Apache-2.0
-->
```

[self-resolve]

integrations/flashvsr/pyproject.toml
```diff
@@ -0,0 +1,72 @@
...
+# Pytest's ``configfile`` resolution picks the nearest ``pyproject.toml`` to
+# the test files, so when running ``pytest integrations/flashvsr/tests``
+# this file shadows the root ``flashdreams`` config -- mirror its marker
+# registrations here to avoid PytestUnknownMarkWarning on ``@pytest.mark.slow``
+# (used by ``test_color_corrector_benchmark.py`` /
+# ``test_projector_benchmark.py``). The ``manual`` marker comes from the
+# ``pytest-manual-marker`` plugin and doesn't need re-registering.
+[tool.pytest.ini_options]
+addopts = "--import-mode=importlib"
+markers = [
```
janickm, line 66: why do we have these overwrites here / shouldn't we "inherit" the global markers?

Note that they are outdated as there is no `slow` anymore, it should be `manual` / `ci_gpu` / `ci_cpu` only since #94 but seems like all are currently `manual` (we require at least one of these three if https://github.com/NVIDIA/flashdreams/pull/94/changes#diff-3a6a67912dbf9d95971baa457f1b557cce95b541b7b7e9e18a48400045306705 is working correctly ^^)

janickm: updated this in the latest version - now making use of the marker enformcement (annoying to repeat this) + using the available markers in all tests

janickm: [resolving]

janickm:
Checked that all GPU tests are now executed properly - once the PR is merged to `main` it should pickup `main`s UV cache also.

@wilsonCernWq note that the `compile_cg_bf16` parametrizations (torch.compile + CUDA graphs) account for most of the full 5min GPU unit test runtime (~2 min for just those 2 tests)  - unsure if this can be reduced a little still? [we don't have unit test caching as with bazel, so we are forced to run all CI tests unconditionally, making it important to stay fast if possible :)]

pyproject.toml
liruilong940607: Oh ok then lets unblock this!

liruilong940607 (approved):

PR #73, opened 2026-05-13
build: wire vendored ludus-renderer v0.9.0 into alpadreams workspace

janickm:
## Summary

Wire the vendored ludus-renderer v0.9.0 (EGL/OpenGL backend, landed in #68) into the alpadreams workspace so it replaces the unreachable private GitLab PyPI `ludus-renderer==0.5.0+cuda` wheel.

## Changes

- **Root `pyproject.toml`**: add `integrations/alpadreams/ludus-renderer` as explicit workspace member (the `integrations/*` glob only matches direct children), add to pyright/ty extra-paths
- **`integrations/alpadreams/pyproject.toml`**: switch `ludus-renderer` from pinned GitLab index to `{ workspace = true }`, remove `[[tool.uv.index]]` block, add `imageio` as direct dep (was transitive in old version)
- **`renderer.py`**: rename `LudusCudaTimestampedContext` -> `LudusTimestampedContext` (CUDA rasterizer was removed in v0.9.0; the GL context exposes the same API; `needs_vflip` is already handled)
- **CI workflow**: update comment explaining why ludus-renderer is still skipped on CPU lint runners (JIT needs EGL + CUDA headers)
- **New test**: GPU smoke tests (`@pytest.mark.manual`) for both the low-level `LudusTimestampedContext` and the high-level `LudusRenderer` wrapper, using the bundled `clipgt.zip` example data

## Verification

- `uv sync` resolves ludus-renderer 0.9.0 from workspace
- `pre-commit run -a` passes (ruff, ty, whitespace, EOF)
- `pytest -m "not manual"` -- no regressions (same 5 pre-existing failures: HF 404, CUDA OOM, importlib identity check)
- Full inference run (`flashdreams-run alpadreams-sv-2steps-chunk2-loc6-lightvae-lighttae --total-blocks 20 --example-data True`) completes all 20 AR steps
- `render_hdmap_scene.py` example script renders successfully with EGL backend
- Both new smoke tests pass (`pytest --runxfail integrations/alpadreams/tests/test_ludus_renderer.py`)

## Context

PR #68 vendored ludus-renderer v0.9.0 (EGL-only, cudaraster removed) into the tree, and PR #69 added Apache-2.0 SPDX headers. But neither wired it into the workspace -- alpadreams was still pulling `0.5.0+cuda` from the internal GitLab registry (unreachable from GitHub Actions). This PR completes the wiring that PR #44 had planned.

PR #74, opened 2026-05-13
fix: handle single-image batch in LudusRenderer.render_all_frames_and_cameras

janickm:
## Summary

Fix a crash in `LudusRenderer.render_all_frames_and_cameras` when rendering exactly one image (1 camera x 1 frame).

## Bug

`ctx.render()` returns `[N, H, W, 4]` where `N = n_cameras * n_frames`. The code then did:

```python
rgb.squeeze(0).permute(0, 3, 1, 2).contiguous().view(...)
```

When `N == 1`, `squeeze(0)` removes the batch dimension entirely, producing `[H, W, 3]` (3D). The subsequent `.permute(0, 3, 1, 2)` then fails with:

```
RuntimeError: permute(sparse_coo): number of dimensions in the tensor input
does not match the length of the desired ordering of dimensions i.e.
input.dim() = 3 is not equal to len(dims) = 4
```

## Fix

Remove the unnecessary `squeeze(0)` -- the tensor is already `[N, H, W, 3]` and can be directly permuted and reshaped to `[n_cameras, n_frames, 3, H, W]` regardless of batch size:

```python
rgb.permute(0, 3, 1, 2).contiguous().view(n_cameras, n_frames, 3, H, W)
```

## Test

Parametrized the existing wrapper smoke test over `n_frames=[1, 2]` so the single-image edge case is explicitly covered. All 3 tests pass (`--runxfail`):

```
test_ludus_timestamped_context_renders_frame            PASSED
test_ludus_renderer_wrapper_renders_frames[single-frame] PASSED
test_ludus_renderer_wrapper_renders_frames[multi-frame]  PASSED
```

Stacked on #73.

### DIFF
--- a/integrations/alpadreams/alpadreams/conditioning/renderer.py
+++ b/integrations/alpadreams/alpadreams/conditioning/renderer.py
@@ -228,12 +228,9 @@ class LudusRenderer:
         rgb = images[:, :, :, :3]
         if self.ctx.needs_vflip:
             rgb = rgb.flip(1)
-        return (
-            rgb.squeeze(0)
-            .permute(0, 3, 1, 2)
-            .contiguous()
-            .view(n_cameras, n_frames, 3, H, W)
-        )
+        # rgb is [N, H, W, 3] where N = n_cameras * n_frames.
+        # Rearrange to [n_cameras, n_frames, 3, H, W].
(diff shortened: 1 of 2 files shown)


janickm:
merging as simple bugfix only (uncovered by new ludus-renderer unit test)

PR #75, opened 2026-05-13
build: add PyPI deployment pipeline and unified versioning

janickm:
## Summary

- Adds unified versioning with `flashdreams/flashdreams/_version.py` as the single source of truth (currently `0.1.0a0`)
- Adds `.github/scripts/sync_version.py` + pre-commit hook to propagate version to all integration `pyproject.toml` files
- Switches `flashdreams/pyproject.toml` to `dynamic = ["version"]` via setuptools `attr`
- Adds `publish-testpypi` CI job that builds a pure-Python wheel and uploads to Test PyPI on every push to `main` (`--skip-existing --non-interactive`)
- Fills in the `cpu` job placeholder with actual `uv build --wheel` command
- Fixes docs CI to read version from `_version.py` instead of the removed static `pyproject.toml` field
- Adds `twine` to a `ci` dependency group in root `pyproject.toml`
- Documents versioning scheme and publishing in `DEV.md`

## Test PyPI

Published and verified: https://test.pypi.org/project/flashdreams/0.1.0a0

To test the core package:

```bash
uv venv --python 3.12
uv pip install --index-url https://test.pypi.org/simple/ --no-deps flashdreams==0.1.0a0

python -c "import flashdreams; print(flashdreams.__version__)"
# 0.1.0a0
```

`TEST_PYPI_API_TOKEN` secret is configured. The `publish-testpypi` job runs on pushes to `main` only.

**TEMPORARY:** Test PyPI is used while the project is pre-release. When going public, change `--repository testpypi` to `--repository pypi` and the secret from `TEST_PYPI_API_TOKEN` to `PYPI_API_TOKEN` in `.github/workflows/ci.yml`.

## Installing integration packages

Integration packages are not published to PyPI. Install them directly from the git repo:

```bash
# From main
uv pip install "flashdreams-wan21 @ git+https://github.com/NVIDIA/flashdreams.git@main#subdirectory=integrations/wan21"

# From a specific branch
uv pip install "flashdreams-wan21 @ git+https://github.com/NVIDIA/flashdreams.git@dev/janickm/pypi-deployment#subdirectory=integrations/wan21"

# From a specific commit
uv pip install "flashdreams-wan21 @ git+https://github.com/NVIDIA/flashdreams.git@3606e610#subdirectory=integrations/wan21"
```

Available integration packages:

| Package | Subdirectory |
|---------|-------------|
| `flashdreams-wan21` | `integrations/wan21` |
| `flashdreams-self-forcing` | `integrations/self_forcing` |
| `flashdreams-causal-forcing` | `integrations/causal_forcing` |
| `flashdreams-fastvideo-causal-wan22` | `integrations/fastvideo_causal_wan22` |
| `flash-alpadreams` | `integrations/alpadreams` |
| `flash-lingbot` | `integrations/lingbot` |

## Versioning scheme

| Package | Published | Version |
|---------|-----------|---------|
| flashdreams | Test PyPI | canonical (from `_version.py`) |
| flashdreams-wan21 | git only | synced |
| flashdreams-self-forcing | git only | synced |
| flashdreams-causal-forcing | git only | synced |
| flashdreams-fastvideo-causal-wan22 | git only | synced |
| flash-alpadreams | git only | synced |
| flash-lingbot | git only | synced |
| ludus-renderer | git only | independent |

## How to bump

1. Edit `__version__` in `flashdreams/flashdreams/_version.py`
2. Commit -- pre-commit hook syncs all integration `pyproject.toml` files
3. Push to `main` -- CI builds and uploads the wheel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,10 +43,13 @@ jobs:
       - name: Setup proxy cache
         uses: nv-gha-runners/setup-proxy-cache@main
 
-      - name: Run wheel build
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v6
+
+      - name: Build flashdreams wheel
         run: |
-          echo "Building wheels..."
-          # Add your wheel build commands here
+          uv run .github/scripts/sync_version.py || true
+          uv build --wheel --package flashdreams
 
       # Documentation is built and deployed by `.github/workflows/doc.yml`.
 
@@ -73,3 +76,36 @@ jobs:
         run: |
           echo "Running world model diffusion workloads..."
           # Add your world model diffusion commands here
+
+  # ===========================================================================
+  # Publish to Test PyPI -- runs only on main after lint passes.
+  # TEMPORARY: targets test.pypi.org while the project is pre-release.
+  # Switch --publish-url/--check-url to the real PyPI endpoints and the
+  # secret to PYPI_API_TOKEN when going public.
+  # ===========================================================================
+  publish-testpypi:
+    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+    needs: [lint]
+    runs-on: linux-amd64-cpu8
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v6
+
+      - name: Sync versions
+        run: uv run .github/scripts/sync_version.py || true
+
+      - name: Build wheel
+        run: uv build --wheel --package flashdreams
+
+      - name: Publish to Test PyPI
+        # --check-url queries the index before uploading and silently skips
+        # files whose version already exists (equivalent to twine --skip-existing).
+        run: >
+          uv publish
+          --publish-url https://test.pypi.org/legacy/
+          --check-url https://test.pypi.org/simple/
+          --token ${{ secrets.TEST_PYPI_API_TOKEN }}
+          dist/*
--- a/flashdreams/pyproject.toml
+++ b/flashdreams/pyproject.toml
@@ -19,7 +19,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "flashdreams"
-version = "0.1.0"
+dynamic = ["version"]
 description = "Flash video simulation pipeline"
 readme = "README.md"
 requires-python = ">=3.12,<3.13"
@@ -89,6 +89,12 @@ runners = [
     "opencv-python-headless",
     "scipy>=1.11",
 ]
+
+[tool.setuptools.dynamic]
+# Resolve version from _version.py (not __init__.py) so setuptools never
+# imports the full package at build time -- avoids triggering transitive
+# imports of heavy dependencies like torch / transformer-engine.
+version = {attr = "flashdreams._version.__version__"}
 
 [tool.setuptools.packages.find]
 include = ["flashdreams*"]
--- a/.github/scripts/sync_version.py
+++ b/.github/scripts/sync_version.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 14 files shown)


flashdreams/pyproject.toml
```diff
@@ -19,7 +19,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "flashdreams"
-version = "0.1.0"
+dynamic = ["version"]
 description = "Flash video simulation pipeline"
 readme = "README.md"
 requires-python = ">=3.12,<3.13"
```
janickm, line 25: @liruilong940607 we are currently quite restrictive on this front - are you sure we need to pin to 3.12 / could we maybe relax this (in a later change?)

flashdreams/pyproject.toml
```diff
@@ -19,7 +19,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "flashdreams"
-version = "0.1.0"
+dynamic = ["version"]
 description = "Flash video simulation pipeline"
```
janickm, line 23: minor: we also likely want to have more elaborate summary + description here - the ncore wheel has a minimal PyPi description like this

https://github.com/NVIDIA/ncore/blob/main/ncore/PYPI_README.md

as an example

pyproject.toml
```diff
@@ -97,6 +97,7 @@ invalid-method-override = "ignore"
 replace-imports-with-any = ["transformer_engine.**", "ludus_renderer.**", "triton.**"]
 
 [dependency-groups]
+ci = ["twine>=6.0"]
```
jatentaki: We should be fine with just `uv publish` and not need `twine` at all? [docs](https://docs.astral.sh/uv/guides/package/#publishing-your-package)

flashdreams/pyproject.toml
```diff
@@ -90,6 +90,9 @@ runners = [
     "scipy>=1.11",
 ]
 
+[tool.setuptools.dynamic]
+version = {attr = "flashdreams._version.__version__"}
```
jatentaki, line 97: Do we need to expose the private `_version` here instead of just `flashdreams.__version__`?

.github/workflows/ci.yml
```diff
@@ -73,3 +76,37 @@ jobs:
...
+
+      - name: Build wheel
+        run: uv build --wheel --package flashdreams
+
+      - name: Publish to Test PyPI
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        run: >
+          uv run --group ci twine upload
```
jatentaki: See the other comment, we shouldn't need `twine`.

jatentaki (approved):
Mostly LGTM, left minor comments.

flashdreams/pyproject.toml
janickm: The `attr` directive points at `flashdreams._version.__version__` intentionally -- setuptools imports the target module at build time to resolve the attribute. Pointing at `flashdreams.__version__` would import `flashdreams/__init__.py`, which today just re-exports from `_version`, but in the future could import heavy dependencies (torch, etc.) that aren't available in the build environment. Keeping the `attr` on the private `_version` module avoids that -- it's a one-liner with no imports, so the build never pulls in the runtime dependency graph. This is the standard pattern (pip, black, ruff all do the same).

The `_version.py` file also serves as the single source of truth that `.github/scripts/sync_version.py` reads to propagate the version to all integration `pyproject.toml` files.

pyproject.toml
janickm: Good call -- switched to `uv publish` with `--publish-url` / `--check-url` / `--token`. Removed the `twine` dependency group entirely.

.github/workflows/ci.yml
janickm: Done -- replaced with `uv publish` in the latest push.

flashdreams/pyproject.toml
jatentaki: Should we add a comment `# importing from the _version.py file should trigger no transitive imports`?

janickm: added comment on this intentional structure [resolving, assuming addressed]

flashdreams/pyproject.toml
janickm: created #78 to track - resolving

flashdreams/pyproject.toml
janickm: tracking in #79

janickm:
Thanks for the review @jatentaki - merging for now, we can still iterate on the details

janickm:
[vent: github's UI really sucks compared to gitlab - one can create new issues from comments, but not auto-resolve the existing ones. when updating a branch via "rebase" the rebased commits are not "verified" anymore (although I "clicked" this with my account ^^), so that the copy-pr-bot doesn't trigger the GPU runners..]

janickm:
/ok to test https://github.com/NVIDIA/flashdreams/commit/5043bfdaf84d9c3714525fabb34220e701ace01e

PR #83, opened 2026-05-14
Add Wan anti-drift KV-cache-relative RoPE preset

JunchenLiu77:
## Summary

This PR adds a Wan/Self-Forcing anti-drift preset benefit from KV-cache-relative RoPE. It keeps the default Wan path on standard RoPE-before-KV-cache-write, and the new anti-drift path stores raw cached keys and reapplies cache-relative RoPE to cached keys on read.

The public runner is:

`self-forcing-wan2.1-t2v-1.3b-anti-drift`

with `sink_size_t=5`, `window_size_t=7`, `network.apply_rope_before_kvcache=False`, and `use_cuda_graph=False`.

## Validation

Standard-path before/after speed and visual A/B:

| run | hot_blocks | hot_frames | hot_total_s | hot_fps_skip8 |
| --- | ---: | ---: | ---: | ---: |
| upstream/main default CP=1 | 12 | 144 | 3.96 | 36.374 |
| PR default CP=1 | 12 | 144 | 4.02 | 35.834 |
| upstream/main default CP=2 | 12 | 144 | 3.53 | 40.804 |
| PR default CP=2 | 12 | 144 | 3.53 | 40.797 |
| PR default SF 60s | 72 | 864 | 24.15 | 35.783 |
| PR anti-drift 60s | 72 | 864 | 24.50 | 35.259 |

</div>

Browser-review artifacts:

- Default standard-path 2x2 grid, upstream vs PR, CP=1/CP=2:
  http://oci-hsg-cs-001-vscode-01:8765/pr83_rebased_validation/pr83_default_standard_cp_grid.mp4
- Long default-vs-anti-drift comparison:
  http://oci-hsg-cs-001-vscode-01:8765/pr83_rebased_validation/pr83_antidrift_left_right_80blocks_seed42.mp4


flashdreams/flashdreams/core/attention/kvcache.py
```diff
@@ -80,6 +87,9 @@ class BlockKVCache:
     dtype: torch.dtype = torch.float16
     """Data type to store the cache in."""
 
+    compression_policy: KVCompressionPolicy | None = None
```
liruilong940607: As discussed offline, it feels easier for the user to understand this if we put it into the category of "sink": treat it as a dynamically updated sink comparing the default static sink.

My proposal is that here we define a `sink_update_policy` and it contains multiple ema alphas and multiple sink_sizes (we can push the sink_size config into this policy). And when ema alpha is zero, it should fall back to the standard sink.

The SinkUpdatePolicy can just be a simple `dataclasses` (treat it as config). We update the functions in `BlockKVCache` to respect this config when evicting tokens (to decide whether we want to update the sink or not). Don't hook those update functions into SinkUpdatePolicy, just keep them in `BlockKVCache`.

flashdreams/flashdreams/core/attention/kvcache.py
```diff
@@ -132,13 +148,29 @@ def __post_init__(self) -> None:
...
+        else:
+            assert (self.window_size + self.sink_size) % self.chunk_size == 0, (
+                f"window_size + sink_size ({self.window_size + self.sink_size}) must be divisible by chunk_size ({self.chunk_size})"
+            )
 
         self._k = torch.empty(self.k_shape, device=self.device, dtype=self.dtype)
         self._v = torch.empty(self.v_shape, device=self.device, dtype=self.dtype)
 
+    @property
+    def stores_prerope_keys(self) -> bool:
```
liruilong940607: i think whether prerope or postrope shouldn't belong to KVcache conceptually. It should be a config belongs to the SelfAttention class. can we move it to the SelfAttention.__init__ and expose it via 
```
NetworkConfig:
    apply_rope_before_kvcache: bool = True
```

flashdreams/flashdreams/core/attention/rope.py
```diff
@@ -213,19 +271,75 @@ def shift_t(self, autoregressive_index: int) -> Tensor:
...
+        1.0
+        / torch.pow(
+            10000,
+            torch.arange(0, dim, 2, device=device, dtype=torch.float64) / dim,
+        ),
+    )
+    return torch.polar(torch.ones_like(freqs), freqs)
+
+
+def apply_rope_frame_indices_official(
```
liruilong940607: This is largely duplicated with https://github.com/NVIDIA/flashdreams/blob/main/flashdreams/flashdreams/core/attention/rope.py#L64

can we merge whatever changes needed into https://github.com/NVIDIA/flashdreams/blob/main/flashdreams/flashdreams/core/attention/rope.py#L64?

Also this impl doesn't seem efficient -- it recompute things that can be cached -- and does it even work with multiGPU?

JunchenLiu77:
@liruilong940607 Integrated EMA memory token into sink and remove redundant rope function.

flashdreams/flashdreams/core/attention/kvcache.py
```diff
@@ -190,8 +377,8 @@ def _overwrite_rightmost_steady(self, k: Tensor, v: Tensor) -> None:
             self._k[sl_write] = k[sl_read]
             self._v[sl_write] = v[sl_read]
 
-    def _overwrite_rightmost_filling(self, k: Tensor, v: Tensor) -> None:
-        """Write the new chunk into the rightmost positions (filling phase)."""
+    def _overwrite_current_chunk(self, k: Tensor, v: Tensor) -> None:
+        """Overwrite the current chunk in the rightmost cache positions."""
```
liruilong940607: The doc string seems to indicate the same meeting as before -- why changing the function name?

flashdreams/flashdreams/core/attention/rope.py
```diff
@@ -211,6 +211,51 @@ def shift_t(self, autoregressive_index: int) -> Tensor:
...
         return freqs
 
+    def frame_indices(self, frame_indices: Tensor) -> Tensor:
+        """Build RoPE frequencies for explicit block-relative frame indices.
+
+        This is used by pre-RoPE KV caches: cached keys keep raw K values, then
+        attention applies RoPE at read time according to each cache slot's
+        physical frame index.
+        """
+        assert not self.is_context_parallel_enabled(), (
```
liruilong940607: I think we need to resolve this -- mgpu support is a must have for us

flashdreams/flashdreams/core/attention/rope.py
```diff
@@ -211,6 +211,51 @@ def shift_t(self, autoregressive_index: int) -> Tensor:
             freqs = torch.cat([freqs_t, freqs_h, freqs_w] * 2, dim=-1)
         return freqs
 
+    def frame_indices(self, frame_indices: Tensor) -> Tensor:
+        """Build RoPE frequencies for explicit block-relative frame indices.
```
liruilong940607: I don't understand what is "block-relative frame indices." ? maybe we could improve the doc string here to be more easily understand?

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
```diff
@@ -257,11 +270,25 @@ def apply_kv(
...
+        cached_k = kv_cache.cached_k()
+        if not self.apply_rope_before_kvcache:
+            assert rope_adapter is not None, (
+                "pre-RoPE KV caches require rope_adapter for read-time RoPE"
+            )
+            q = apply_rope_freqs(
+                q,
+                rope_adapter.frame_indices(kv_cache.query_frame_indices()),
+                interleaved=True,
+            )
```
liruilong940607: This is just the original `apply_rope_freqs(q, rope_freqs, interleaved=True)` I think? Should we stick with the original impl here to be 1/ easier to understand and 2/ more efficient?

Conceptually I think pre or post rope KV cache should not affect Q at all.

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
```diff
@@ -257,11 +270,25 @@ def apply_kv(
...
+            cached_k = apply_rope_freqs(
+                cached_k,
+                rope_adapter.frame_indices(kv_cache.cached_frame_indices()),
+                interleaved=True,
+            )
```
liruilong940607: Passing around `rope_adapter` and compute rope_freqs on the fly is neither easy to understand nor efficient. because each attention operator you are computing the same rope_freqs.

My proposal:

extend the original rope_freqs argument into two: 
```
rope_freqs_q: Tensor,
rope_freqs_k: Tensor | None
```

if `apply_rope_before_kvcache=True`, `rope_freqs_k` is allowed to be None, and just use `rope_freqs_q` in the attention operator for both q and k.

if `apply_rope_before_kvcache=False`, assert `rope_freqs_k` to be not None, and use it for k.

This way you can get rid of all these changes:
- cached_frame_indices() function for kv_cache
- tokens_per_frame attribute for kv_cache
- frame_indices() function for RotaryPositionEmbedding3D

flashdreams/flashdreams/recipes/wan/transformer/wan21.py
```diff
@@ -194,13 +198,30 @@ def __init__(self, config: Wan21TransformerConfig) -> None:
...
+        uses_sink_update = config.sink_update_policy is not None
+        uses_prerope_cache = not config.network.apply_rope_before_kvcache
+        if uses_sink_update or uses_prerope_cache:
+            assert self._cp_size == 1, (
+                "Sink update and explicit-frame RoPE cache paths currently require cp_size=1"
+            )
+            assert not config.use_cuda_graph, (
+                "Sink update and explicit-frame RoPE cache paths have not validated "
+                "CUDA graph support; set use_cuda_graph=False"
+            )
```
liruilong940607: We should support these. I think my proposal above would make it very easy (natually to support these)

flashdreams/flashdreams/recipes/wan/transformer/wan21.py
```diff
@@ -194,13 +198,30 @@ def __init__(self, config: Wan21TransformerConfig) -> None:
...
 
         # Pre-patchify temporal divisibility check; per-rollout
         # (height, width) is populated by initialize_autoregressive_cache.
         kt, _, _ = config.network.patch_size
         assert config.len_t % kt == 0, (
             f"len_t ({config.len_t}) must be divisible by patch_size[0] ({kt})."
         )
+        if config.sink_update_policy is not None:
+            assert kt == 1, "Sink update currently assumes patch_size[0] == 1"
```
liruilong940607: I dont think you need to assert this. just do 

```
config.sink_update_policy.validate_temporal_layout(
    len_t=config.len_t // kt,
    window_size_t=config.window_size_t,
    sink_size_t=config.sink_size_t,
)
```

KV cache also just see the "post-patchify" len_t

flashdreams/tests/test_kvcache.py
```diff
@@ -156,6 +158,113 @@ def test_block_kvcache_matches_baseline(
         torch.testing.assert_close(v_api, v_baseline)
 
 
+@pytest.mark.parametrize("num_dynamic_sinks", [0, 1, 2, 3])
+def test_block_kvcache_sink_update_policy_dynamic_sink_math(
+    num_dynamic_sinks: int,
+) -> None:
+    """Dynamic sink segments update from evicted window tokens; static sinks do not."""
+    batch, n_heads, dim = 1, 1, 1
+    chunk_size = 3
```
liruilong940607: This test is very hard to read and follow. It's fine to keep it if you want but could we have a test where you extend the _NaiveKVCache to support dynamic cache and test against that? Like the other tests

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
liruilong940607: I think we can have `RotaryPositionEmbedding3D.shift_t` to return a tuple [Tensor, Tensor | None]

with interface 

```
RotaryPositionEmbedding3D
def __init__(..., apply_rope_before_kvcache: bool = True)
```

and use network_config.apply_rope_before_kvcache to intialize it so it always align with the network.

then in `RotaryPositionEmbedding3D.shift_t` it can needs to calculate `rope_freqs_k` if `self.apply_rope_before_kvcache=False`. And it should already have all the info it needs to compute that so no additional argument need to pass in.

flashdreams/flashdreams/core/attention/rope.py
liruilong940607, line 1: Do we need those changes in this file?

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
```diff
@@ -561,7 +594,7 @@ def forward(
             x: Input tensor with shape [..., L, D].
             e: Modulation tensor with shape [..., 6, D].
             cache: KV cache container for this block.
-            rope_freqs: RoPE frequencies with shape [L, 1, 1, head_dim // 2].
+            rope_freqs: Cache-layout RoPE frequencies.
```
liruilong940607: Should keep the shape indicator and be more clear on what are the two types of rope_freqs it accepts

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
```diff
@@ -280,15 +308,18 @@ def forward(
         Args:
             x: Query tensor and, when updating, the source for new K/V ([..., L, n * d]).
             kv_cache: Cache read by attention; written when ``update_kv_cache`` is True.
-            rope_freqs: Optional RoPE frequencies for Q and (when updating) K.
+            rope_freqs: Optional RoPE frequencies. Post-RoPE K mode receives
```
liruilong940607: "Post-RoPE" "Pre-RoPE" lets try avoid use these words. Its very confusing whether it means
-  KVcache is post/pre RoPE, or
- RoPE is pre/post KVCache.

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
```diff
@@ -280,15 +308,18 @@ def forward(
...
+                the current standard table. Raw-K cache mode receives the
+                KV-cache-relative table.
```
liruilong940607: "table"? -- also weird word here

"Raw-K cache mode" -- I also dont understand

flashdreams/flashdreams/recipes/wan/transformer/impl/network.py
```diff
@@ -99,6 +99,9 @@ class WanDiTNetworkConfig(InstantiateConfig):
     patch_embedding_type: Literal["linear", "conv3d"] = "conv3d"
     """Type of patch embedding: ``"linear"`` (flattened patch MLP) or ``"conv3d"`` (strided conv)."""
 
+    apply_rope_before_kvcache: bool = True
+    """If False, store pre-RoPE self-attention keys and apply RoPE on cache reads."""
```
liruilong940607: here "pre-RoPE ". There are many places I'll skip commenting on those

flashdreams/flashdreams/recipes/wan/transformer/impl/network.py
```diff
@@ -402,8 +407,7 @@ def forward(
                 layout ``"... (t h w) (c kt kh kw)"``.
             timesteps: Diffusion timesteps, broadcastable to shape ``[...]``.
             cache: Network KV caches.
-            rope_freqs: RoPE frequencies after CP, shape
-                ``[L, 1, 1, head_dim // 2]``.
+            rope_freqs: Cache-layout RoPE frequencies after CP.
```
liruilong940607: Same comments

flashdreams/flashdreams/recipes/wan/transformer/wan21.py
```diff
@@ -94,11 +97,19 @@ def finalize(self, autoregressive_index: int) -> None:
 ## Transformer
 
 
+def _post_patch_sink_size_t(sink_size_t: int, kt: int) -> int:
```
liruilong940607: Can we avoid this function just inline `sink_size_t // kt` like before?

flashdreams/flashdreams/recipes/wan/transformer/wan21.py
```diff
@@ -369,16 +396,23 @@ def initialize_autoregressive_cache(
...
+            "head_dim": head_dim,
+            "h_extrapolation_ratio": self.config.h_extrapolation_ratio,
+            "w_extrapolation_ratio": self.config.w_extrapolation_ratio,
+            "interleaved": True,
+            "device": self.device,
+        }
+        if cfg.network.apply_rope_before_kvcache:
+            rope_adapter = RotaryPositionEmbedding3D(**rope_kwargs)
+        else:
+            rope_kwargs["cache_len_t"] = (
```
liruilong940607: suggestion: just pass in `sink_size_t` and `window_size_t` into `KVCacheRelativeRotaryPositionEmbedding3D` to be more readable (`cache_len_t` is not very clear whether it indicates kv cache or yet another cache for this class)

and in `KVCacheRelativeRotaryPositionEmbedding3D` you can do
```
kvcache_total_size_t = sink_size_t + window_size_t
```

integrations/self_forcing/self_forcing/config.py
```diff
@@ -114,6 +114,37 @@ def state_dict_transform(state_dict: dict[str, Any]) -> dict[str, Tensor]:
     pipeline=PIPELINE_WAN21_T2V_1PT3B_FLASH,
 )
 
+# KV-cache-relative RoPE static-sink sink5-window7 baseline:
+# static sink=5, rolling window=7 (recent=4 + current len_t=3).
+PIPELINE_WAN21_T2V_1PT3B_SINK5_WINDOW7 = cast(
+    WanInferencePipelineConfig,
+    derive_config(
+        PIPELINE_WAN21_T2V_1PT3B,
+        recipe_name="self-forcing-wan2.1-t2v-1.3b-sink5-window7",
```
liruilong940607: "sink5-window7"

this is not only sink with window7 but also `apply_rope_before_kvcache=False`.

Suggest call it `self-forcing-wan2.1-t2v-1.3b-anti-drift` to be more readable what this config does

liruilong940607 (commented):
Let's also update the MR title, description, with proper testing -- since it touches the fundamental part of the code base lets run a A/B test before and after this MR on self-forcing to make sure both speed and quality are not affected for the standard rope code path.

please update the description with the all results brought in from this MR and the testing.

liruilong940607:
Btw with this MR just landed: https://github.com/NVIDIA/flashdreams/pull/94 you now need to annotate each new test to be one of the three categories (see that MR)

liruilong940607:
/ok to test [2c61486](https://github.com/NVIDIA/flashdreams/pull/83/commits/2c614866977d19064d1c73a49e10fc293177d987)

flashdreams/flashdreams/core/attention/rope.py
```diff
@@ -174,6 +153,63 @@ def context_parallel_size(self) -> int:
...
+        """Split a KV-cache-layout RoPE tensor by chunk for CP."""
+        cp_group = unpack_optional(self.cp_group)
+        tokens_per_chunk = self.len_t * self.len_h * self.len_w
+        assert valid_len_t % self.len_t == 0
+        valid_tokens = valid_len_t * self.len_h * self.len_w
+        freqs = freqs[:valid_tokens]
+        freq_shape = freqs.shape[1:]
+        freqs = freqs.reshape(valid_len_t // self.len_t, tokens_per_chunk, *freq_shape)
+        freqs = split_inputs_cp(freqs, seq_dim=1, cp_group=cp_group)
+        return freqs.reshape(-1, *freq_shape)
```
liruilong940607: This should belong to only `KVCacheRelativeRotaryPositionEmbedding3D`?

flashdreams/flashdreams/core/attention/rope.py
```diff
@@ -87,18 +83,6 @@ def __init__(
...
-
-        Args:
-            head_dim: Attention head dimension; split into h/w/t sub-dims (2:2:2 ratio).
-            len_h: Sequence length along height.
-            len_w: Sequence length along width.
-            len_t: Sequence length along time.
-            h_extrapolation_ratio: NTK extrapolation ratio for height.
-            w_extrapolation_ratio: NTK extrapolation ratio for width.
-            t_extrapolation_ratio: NTK extrapolation ratio for time.
-            interleaved: Whether to interleave the frequency components.
```
liruilong940607, line 100: Removing them from the base class is fine -- but we need to have them for the derived classes `RotaryPositionEmbedding3D` and `KVCacheRelativeRotaryPositionEmbedding3D`

flashdreams/flashdreams/core/attention/rope.py
```diff
@@ -189,29 +225,100 @@ def shift_t(self, autoregressive_index: int) -> Tensor:
...
+        Args:
+            autoregressive_index: Accepted for API parity with standard RoPE.
+                Must be 0 because cache-relative positions are fixed by
+                KV-cache position, not by global AR step.
+
+        Returns:
+            RoPE frequencies of shape ``[S, 1, 1, head_dim]``, where ``S`` is
+            the valid KV-cache token count after optional CP splitting.
+        """
+        assert autoregressive_index == 0, (
```
liruilong940607: Instead of assert `autoregressive_index==0`, lets just ignore `autoregressive_index` here. Conceptually we shouldn't and dont need to repurpose `autoregressive_index`. It's just `KVCacheRelativeRotaryPositionEmbedding3D` is not related with `autoregressive_index` anymore

flashdreams/flashdreams/core/attention/rope.py
```diff
@@ -221,7 +328,7 @@ def apply_rope_freqs(x: Tensor, freqs: Tensor, interleaved: bool = False) -> Ten
     Args:
         x: Input tensor of shape ``[B, S, H, D]``; rotated in place.
         freqs: RoPE frequencies of shape ``[S, 1, 1, D]`` as emitted by
-            :meth:`RotaryPositionEmbedding3D.shift_t`.
```
liruilong940607, line 224: It should be 

>as emitted by :meth:`RotaryPositionEmbedding3D.shift_t` or :meth:`KVCacheRelativeRotaryPositionEmbedding3D.shift_t`

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
```diff
@@ -257,17 +263,39 @@ def apply_kv(
...
+    def _slice_rope_freqs(
+        self,
+        rope_freqs: Tensor | None,
+        kv_cache: BlockKVCache,
+    ) -> tuple[Tensor | None, Tensor | None]:
+        """Select Q/K RoPE frequencies for standard or cache-relative mode."""
+        if rope_freqs is None:
+            return None, None
+        if self.apply_rope_before_kvcache:
+            return rope_freqs, None
```
liruilong940607: Lets just return `rope_freqs, rope_freqs` here.

then in forward

just do

```
rope_freqs_q, rope_freqs_k = self._slice_rope_freqs(rope_freqs, kv_cache)
if update_kv_cache:
      kv_cache = self.update_kv(x, kv_cache, rope_freqs_k)
```

This feels slightly more elegant

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
```diff
@@ -561,7 +594,9 @@ def forward(
             x: Input tensor with shape [..., L, D].
             e: Modulation tensor with shape [..., 6, D].
             cache: KV cache container for this block.
-            rope_freqs: RoPE frequencies with shape [L, 1, 1, head_dim // 2].
+            rope_freqs: RoPE frequencies with shape ``[S, 1, 1, head_dim]``.
```
liruilong940607: `[S, 1, 1, head_dim]` -> `[L, 1, 1, head_dim // 2]`

flashdreams/flashdreams/recipes/wan/transformer/impl/network.py
```diff
@@ -403,7 +408,8 @@ def forward(
             timesteps: Diffusion timesteps, broadcastable to shape ``[...]``.
             cache: Network KV caches.
             rope_freqs: RoPE frequencies after CP, shape
-                ``[L, 1, 1, head_dim // 2]``.
+                ``[S, 1, 1, d // 2]``. Standard mode uses current-chunk
```
liruilong940607: `[S, 1, 1, d // 2]` -> `[L, 1, 1, d // 2]`

flashdreams/flashdreams/recipes/wan/transformer/wan21.py
```diff
@@ -60,11 +63,13 @@ class Wan21TransformerCache(TransformerAutoregressiveCache):
     network_cache_uncond: WanDiTNetworkCache | None = None
     """Unconditional caches; ``None`` disables CFG."""
 
-    rope_adapter: RotaryPositionEmbedding3D
-    """3D RoPE adapter; advances along T per AR step."""
+    rope_adapter: RotaryPositionEmbedding3D | KVCacheRelativeRotaryPositionEmbedding3D
+    """3D RoPE indexing runtime used to build the per-step cache layout."""
```
liruilong940607: "build per-step cache layout" -- not really a good description. Maybe the original description is better.

flashdreams/flashdreams/recipes/wan/transformer/wan21.py
```diff
@@ -77,8 +82,14 @@ def start(self, autoregressive_index: int) -> None:
...
-        # cat/repeat on every cond/uncond pass.
-        self.rope_freqs = self.rope_adapter.shift_t(autoregressive_index)
+        # cat/repeat on every cond/uncond pass. KV-cache-relative RoPE uses
+        # fixed cache positions, so it always requests the unshifted frequencies.
+        rope_shift_idx = (
+            0
+            if isinstance(self.rope_adapter, KVCacheRelativeRotaryPositionEmbedding3D)
+            else autoregressive_index
+        )
+        self.rope_freqs = self.rope_adapter.shift_t(rope_shift_idx)
```
liruilong940607: we can revert the change here after we apply the above proposal on ignoring `autoregressive_index` in `KVCacheRelativeRotaryPositionEmbedding3D`

liruilong940607 (commented):
I think we are getting there! just some minor comments now.

liruilong940607:
For lint fix:

```
#!/usr/bin/env bash
#
# Run pre-commit (ruff + ty) locally against the same env CI uses.
# See `.github/workflows/ci.yml` -> `lint` job.

set -euo pipefail

uv sync --extra dev --group lint \
    --no-install-package transformer-engine-torch \
    --no-install-package ludus-renderer

uv run --no-sync pre-commit run -a
```

liruilong940607:
> Long default-vs-anti-drift comparison:
http://oci-hsg-cs-001-vscode-01:8765/pr83_rebased_validation/pr83_antidrift_left_right_80blocks_seed42.mp4

This is nice. can you add another video with same sink size but with standard rope? To show how much the relative rope affects

flashdreams/flashdreams/recipes/wan/transformer/impl/modules.py
JunchenLiu77: The last dimension should be head_dim actually. see _cat_freqs fn in rope.py.

flashdreams/flashdreams/recipes/wan/transformer/impl/network.py
JunchenLiu77: The last dimension should be head_dim actually. see _cat_freqs fn in rope.py.

JunchenLiu77:
> > Long default-vs-anti-drift comparison:
> > http://oci-hsg-cs-001-vscode-01:8765/pr83_rebased_validation/pr83_antidrift_left_right_80blocks_seed42.mp4
> 
> This is nice. can you add another video with same sink size but with standard rope? To show how much the relative rope affects

Sure! Results here: 
http://cs-oci-ord-vscode-01:8765/pr83_rope_comparison/pr83_standard_vs_relative_rope_sink5_window7_seed42_60s.mp4

liruilong940607:
/ok to test [4ad16cc](https://github.com/NVIDIA/flashdreams/pull/83/commits/4ad16cc3a4bee0435fd05cfc8e821226dc981468)

liruilong940607 (approved):
Looking Great Now!

liruilong940607:
/ok to test [0752d54](https://github.com/NVIDIA/flashdreams/pull/83/commits/0752d544bbca0b77fc013a2cf76050c64e1e379a)

liruilong940607:
/ok to test [c5fe30c](https://github.com/NVIDIA/flashdreams/pull/83/commits/c5fe30c8363d874ac75f0cbf8e5fb94345180e67)

PR #84, opened 2026-05-14
Streaming update 

liruilong940607:
Recording:

https://github.com/user-attachments/assets/0d2313ea-cecf-4375-8852-ffd11375fb86

### DIFF
--- a/flashdreams/flashdreams/recipes/lingbot_world/config.py
+++ b/flashdreams/flashdreams/recipes/lingbot_world/config.py
@@ -88,8 +88,7 @@ LINGBOT_WORLD_FAST = LingbotWorldInferencePipelineConfig(
     enable_sync_and_profile=True,
     encoder=I2VCamCtrlEncoderConfig(
         i2v=WanI2VCtrlEncoderConfig(
-            encoder=WanVAEEncoderConfig(
-            ),
+            encoder=WanVAEEncoderConfig(),
         ),
     ),
     decoder=WanVAEDecoderConfig(),
--- a/integrations/lingbot/README.md
+++ b/integrations/lingbot/README.md
@@ -35,13 +35,20 @@ Lingbot integration package for `flashdreams` that exposes a minimal WebRTC serv
 From repository root:
 
 ```bash
-uv run --package flash-lingbot python -m lingbot.webrtc.server --host 0.0.0.0 --port 8080 --config_name LingBot-World-Fast
+uv run --package flash-lingbot python -m lingbot.webrtc.server --host 0.0.0.0 --port 8089 --config_name lingbot-world-fast-flash
+
+# 4 gpus
+uv run --package flash-lingbot \
+  python -m torch.distributed.run --standalone --nnodes=1 --nproc_per_node=4 \
+  -m lingbot.webrtc.server \
+  --host 0.0.0.0 --port 8089 \
+  --config_name lingbot-world-fast-flash
 ```
 
 Then open:
 
-- [http://localhost:8080/request_session](http://localhost:8080/request_session)
-- [http://localhost:8080/healthz](http://localhost:8080/healthz) (`runtime_ready` indicates preload completion)
+- [http://localhost:8089/request_session](http://localhost:8089/request_session)
+- [http://localhost:8089/healthz](http://localhost:8089/healthz) (`runtime_ready` indicates preload completion)
 
 ## Test
 
--- a/integrations/lingbot/lingbot/webrtc/controls.py
+++ b/integrations/lingbot/lingbot/webrtc/controls.py
@@ -13,8 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Keyboard state, sparse-edge event resampling, and camera pose integration."""
+
 from __future__ import annotations
 
+from collections import deque
 from dataclasses import dataclass, field
 
 import numpy as np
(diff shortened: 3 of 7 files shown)


liruilong940607:
## Changes 1

**Before:** the server kept a single "current key-state" set, updated as edges arrived. At each chunk trigger it took **one snapshot** of that set and applied it to **all 12 frames of the chunk**. The timestamps inside the chunk window were ignored.

```
BEFORE  (server stored only "current pressed-keys set", no timestamps)
─────────────────────────────────────────────────────────────────────────────

   edges from client:    ↓w         ↑w   ↓a              ↑a
                          │          │   │               │
   wall:        ──────────●──────────●───●───────────────●─────▶
                                                                ▲ chunk trigger
                                                                │
                                                          state = {a}
                                                          (current set
                                                          AT trigger time)
                                                                │
                                                                ▼
   chunk window: ─────────────────────────────────────────────────
   frame i:      {a}  {a}  {a}  {a}  {a}  {a}  {a}  {a}  {a}  {a}  {a}  {a}
                  ↑
                  every frame in the chunk gets the same state snapshot

   Consequences:
   • Hold w for the whole window: all 12 frames see {w} → continuous motion ✓
   • Tap w briefly, then release before trigger: state at trigger = {} → 
     the tap produces ZERO motion ✗
   • Press w, release a, press a all in one window: only the *final* state
     at trigger time matters; intermediate intervals collapse ✗
```

**After:** the server keeps the timestamped edges in a journal. At trigger time it builds a *timeline* of how the key-state evolved across the chunk window, and the integrator advances proportionally for each interval.

```
AFTER  (server stores edge journal: (arrival_t, ↓/↑, key))
─────────────────────────────────────────────────────────────────────────────

   edges from client:    ↓w         ↑w   ↓a              ↑a
                          │          │   │               │
   wall:        ──────────●──────────●───●───────────────●─────▶  trigger ▲

   chunk window:          ├──────────┼───┼───────────────┼─────┤
                          │   {w}    │{} │      {a}      │ {} │
                          │←─300 ms─→│←──│←── 700 ms ───→│    │
                                          80 ms

   integrator:
       advance(state={w}, duration=300 ms)   ← real forward motion
       advance(state={ }, duration= 80 ms)
       advance(state={a}, duration=700 ms)   ← real yaw
       advance(state={ }, duration=...   ms)

   Consequences:
   • Hold w for the whole window: same continuous motion as before ✓
   • Brief tap that started AND ended in the window: produces motion = 
     rate × tap_duration  (was lost entirely before) ✓
   • Multiple key changes within one chunk: each interval contributes its 
     real share (was collapsed before) ✓
```

### When you'd see the difference in practice

- **Quick taps** (< chunk duration, ~750 ms): used to vanish; now produce motion proportional to the tap.
- **Key changes mid-chunk** (e.g., release `w`, press `a` halfway through): used to collapse to whatever was held at trigger time; now show both phases inside the chunk.
- **Long holds across many chunks**: identical behaviour in both designs.


## Change 2: how frames are paced from producer to consumer

**Before:** the buffer between the generator and the WebRTC send loop was driven by several seconds-based knobs — a buffer ceiling, a prefetch low-water mark, and an idle-grace timer. The producer wasn't directly throttled by playback drain rate; it was nudged by watermark checks and timers, with separate branches for "idle", "low", and "near-full".

```
BEFORE  (several knobs, watermark-driven prefetch)
─────────────────────────────────────────────────────────────────────────────

   knobs:  max_buffer_seconds   prefetch_low_water_seconds   idle_grace_seconds
              │                       │                            │
              ▼                       ▼                            ▼

   worker: ──generate──▶ buffer  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                            ▲                                            │
                            │  "if depth < low_water → prefetch"         │
                            │  "if depth > max_buffer → idle"            │
                            └────────────────────────────────────────────┘
                                                                         ▼
                                                                       recv ──▶ browser

   Consequences:
   • Buffer depth could creep upward if producer outpaced consumer → latency drift
   • Tuning was guess-and-check across three interacting knobs
   • Idle / low / saturated cases each had their own code path
```

**After:** the video track owns a single bounded queue sized to exactly one chunk (`maxsize = num_frames`). The producer's `put()` blocks the moment the queue is full — that *is* the rate-limiter. No watermarks, no prefetch timers, no idle-grace.

```
AFTER  (one bounded queue; backpressure is the only rate-limiter)
─────────────────────────────────────────────────────────────────────────────

                                          maxsize = num_frames
                                       ┌──────────────────────┐
   worker: ──generate──▶  ─put─▶       │■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ │ ─get─▶ recv ──▶ browser
                            │          └──────────────────────┘                (paced 1/fps)
                            │  put() blocks when full ↑
                            └─────────  → producer paced to consumer drain rate

   Consequences:
   • Buffer depth bounded by construction → latency cannot drift upward ✓
   • One operational parameter (fps); num_frames comes from the pipeline
   • Stall, saturated, and underrun all flow through the same put/get path
```

### When you'd see the difference in practice

- **Long sessions:** buffer growth that used to creep up over minutes is gone; depth always tops out at one chunk.
- **Retuning workloads:** there's nothing to sweep anymore — `fps` and `num_frames` fully determine the behaviour.
- **Reasoning about behaviour:** a single mechanism (queue full → producer waits) covers what used to be three branches.

---

## Change 3: who decides when to start the next chunk

**Before:** chunks were pulled by the browser. After each chunk was enqueued the server emitted `chunk_done`; the browser replied with a `step` message; the server treated that reply as the cue to generate the next chunk. The pacing of generation therefore lived in the browser-↔-server round trip, and generation began the instant the peer connected — even with no user input.

```
BEFORE  (browser-pull: a round-trip triggers each chunk)
─────────────────────────────────────────────────────────────────────────────

   browser:                step                step                step
                            │                   │                   │
   server:  ◀──chunk_done───┤  ◀──chunk_done────┤  ◀──chunk_done────┤
                            │                   │                   │
                            ▼                   ▼                   ▼
                     gen chunk N+1      gen chunk N+2        gen chunk N+3

   Consequences:
   • Inter-chunk timing = browser round-trip; variable and irregular
   • A dropped or late `step` stalled generation
   • No "chunk window started at virtual time V_N"; server just generated on demand
   • GPU spun from peer-connect time, even with zero user input
```

**After:** the server runs a wall-clock-aligned worker that owns the trigger schedule. Chunk N's window is exactly `[V_N, V_{N+1})` with `V_{N+1} = V_N + num_frames / fps`. The worker sleeps until wall ≥ V_{N+1}, then samples and generates. It blocks at startup until the first user edge arrives; if the virtual clock drifts more than one chunk behind wall (cold-start warmup, transient stall), it rewinds to `wall − chunk_duration` so latency stays bounded.

```
AFTER  (server-driven, wall-aligned, idle-until-first-edge)
─────────────────────────────────────────────────────────────────────────────

   first edge ──▶ anchor V_0 = now
                       │
                       ▼
   ┌─── loop ────────────────────────────────────────────────────┐
   │ wait until wall ≥ V_{N+1}                                   │
   │ if wall − V_{N+1} > chunk_duration:                         │
   │     next_v ← wall − chunk_duration       (catch-up rewind)  │
   │ build timeline for window [V_N, V_{N+1})                    │
   │ generate → enqueue (blocks on full queue, ties into Change 2)│
   │ V_N ← V_{N+1}                                               │
   └─────────────────────────────────────────────────────────────┘

   Consequences:
   • Inter-chunk timing regular and predictable, set by num_frames / fps
   • Browser no longer in the timing loop
   • Idle session = zero GPU work until the user interacts
   • Cold-start warmup or transient stalls do not pin latency forever
```

### When you'd see the difference in practice

- **Slow or lossy network:** used to introduce visible chunk-rate jitter on every round-trip; now decoupled from network entirely.
- **First minute of a session:** cold-start used to bake the warmup delay (~11 s observed) into every subsequent chunk; now self-corrects within one cycle.
- **Idle viewer:** GPU used to spin generating still scenes; now stays asleep until the first key edge.

---

## Change 4: how playback recovers from a transient hiccup

**Before:** the send loop kept an absolute deadline that advanced by `+1/fps` each frame and slept until it. After any pause — a long generation, a scheduling stall — that deadline was already well in the past, so the next few `recv` calls computed hugely negative waits and emitted frames as fast as the browser asked. The jitter buffer read the burst of short inter-arrivals as a cue to speed playback up.

```
BEFORE  (deadline += dt; burst-drains after any pause)
─────────────────────────────────────────────────────────────────────────────

   ideal deadline:   t   t+dt    t+2dt    t+3dt    t+4dt    t+5dt   …
                     │    │       │        │        │        │
   wall:             │    │       │        │        │        │            ▶
   frames out:       ●────●───────●────────●────────●        │
                                                    │
                                            ── 200 ms pause ──
                                                    │
   wall:                                            │     ▶
   frames out:                                      ● ● ● ●        ← BURST
                                                    └ deadline far in past,
                                                      4 frames in < 1 ms

   viewer:    ────── steady ────────────────── ▶▶▶▶ fast-forward ▶▶▶▶
```

**After:** the send loop notices when it actually had to wait for a frame (queue stall) or when its deadline has drifted into the past, and re-anchors the deadline to `now` instead of plowing through queued frames. The result of a hiccup is at most one late frame — well within the browser's jitter buffer — instead of a visible speed-up.

```
AFTER  (re-anchor deadline to wall on stall / drift)
─────────────────────────────────────────────────────────────────────────────

   deadline:   t   t+dt    t+2dt   t+3dt   t+4dt              ← re-anchor here
   wall:       │    │       │       │       │                   │
                                                  ▲
                                            ── 200 ms pause ──
                                                  │
   wall:       │    │       │       │       │     │              ▶
   frames out: ●────●───────●───────●───────●───────────────────●─────●─────●
                                                                └ one late
                                                                  frame, then
                                                                  back to 1/fps

   viewer:    ────── steady ─────────────── ─── steady continues ──────────
```

### When you'd see the difference in practice

- **After a generation hiccup or any event-loop pause:** used to see the video fast-forward briefly; now glides past it.
- **At every chunk boundary on a marginal queue:** small recurring 60 ms drops used to compound into visible speed-ups; now absorbed silently by the jitter buffer.
- **Pristine steady state:** identical behaviour in both designs.

liruilong940607:
(I'm merging this so I can start to do some more cleanup on ling bot inference without needing rebase)

PR #89, opened 2026-05-14
Fix HF example-data UUID listing in alpadreams runner

aidanfnv:
`HfApi.list_repo_tree` in huggingface_hub 0.36 returns `RepoFile` / `RepoFolder` dataclass instances that do not expose a `.type` attribute, so the previous `getattr(entry, "type", None) == "file"` filter dropped every entry and `_ensure_hf_single_view_example_data_synced` always raised `FileNotFoundError` (even for UUIDs whose `*_hdmap.mp4` exists on HF). Switch the filter to `isinstance(entry, RepoFile)`, which is the canonical hf_hub idiom.

Upstreamed from nv-tlabs/flashdreams-partner-waabi PR #2 (commit 839200d). Reproduced against huggingface_hub 0.36.2 on a clean checkout: calling `_ensure_hf_single_view_example_data_synced(DEFAULT_EXAMPLE_DATA_UUID_1V)` raises `FileNotFoundError`; with this patch it returns the expected hdmap + first_frame paths.

### DIFF
--- a/flashdreams/flashdreams/recipes/alpadreams/runner.py
+++ b/flashdreams/flashdreams/recipes/alpadreams/runner.py
@@ -105,6 +105,7 @@ def _ensure_hf_single_view_example_data_synced(
     from :data:`EXAMPLE_DATA_HF_REPO` (the hdmap filename is per-clip so
     we list the dir first to find it). Returns ``((hdmap,), (first_frame,))``."""
     from huggingface_hub import HfApi, hf_hub_download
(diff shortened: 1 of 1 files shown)


aidanfnv:
/ok to test [7c64b7e](https://github.com/NVIDIA/flashdreams/pull/89/commits/7c64b7e7f4d840bbf8d29fe47854f6d2fda9a39e)

PR #93, opened 2026-05-16
hy_worldplay: phase-1 vendor wrapper + numerical parity

wenqingw-nv:
## Summary

Phase-1 integration of HY-WorldPlay (WAN-5B I2V) as a flashdreams plugin, plus the parity-fix follow-ups that bring it to numerical parity with upstream `wan/generate.py`.

- Vendor-wrapper runner under `integrations/hy_worldplay/` following the `self_forcing` / `causal_forcing` mini-repo pattern. Standalone CLI (`python -m hy_worldplay.cli`) for phase 1; `flashdreams-run` subcommand promotion tracked as phase 2.
- Isolated `tests/parity_check/` venv that clones upstream, downloads `tencent/HY-WorldPlay` checkpoints, and runs the reference benchmark.
- Numerical parity: mean |Δ| 3.41/255 vs upstream output, 0 frames over the visual threshold — tighter than the torch 2.11→2.12 self-self drift of 3.76. Two diagnosed root causes (torch version pin + DEFAULT_PROMPT trailing-period UMT5 tokenisation) are now guarded by smoke tests.

## Test plan

- [x] `uv run --extra dev pytest integrations/hy_worldplay/tests/test_smoke.py -v` (10/10 passing, includes new prompt-byte-match parity guards)
- [x] `bash integrations/hy_worldplay/tests/parity_check/run.sh` produces upstream baseline mp4
- [x] `python -m hy_worldplay.cli ...` produces plugin mp4 with same args
- [x] Numeric per-frame diff between the two: mean |Δ| 3.41/255, max 130, 0 frames mean|Δ|>5
- [x] `wan21-t2v-1.3b-480p` sanity check via `flashdreams-run` still works (existing-runner regression check)

## Follow-ups (not in this PR)

- Phase 2: refactor HY-WorldPlay's WAN pipeline into a `StreamInferencePipelineConfig` so it becomes a `flashdreams-run hy-worldplay-wan-i2v-5b` subcommand alongside the other recipes.
- Phase 3: HunyuanVideo-1.5 8B variant.


uv.lock
liruilong940607, line 1: This would effectively adding HY worldplay's dependency to the whole repo. Those dependencies would go away eventually after we switch to using flashdreams' DiT for HY worldplay. So to keep it clean, can we avoid this by organizing uv env under HY worldplay repo? like this has its own uv env:

https://github.com/NVIDIA/flashdreams/tree/main/integrations/self_forcing/tests/parity_check

liruilong940607:
/ok to test [6fb9321](https://github.com/NVIDIA/flashdreams/pull/93/commits/6fb93213d1edeb80a84f2ebd55f5d4e74df39aa3)

liruilong940607:
script to fix lint (dump it here in case CI complains about lint)

https://github.com/NVIDIA/flashdreams/pull/83#issuecomment-4474887618

PR #103, opened 2026-05-19
Isolated HY dependency scope

wenqingw-nv:
- Plugin packaging + workspace member + heavy deps scoped to parity sub-venv
- HyWorldPlayWanI2VRunner shim delegating to upstream's WanRunner.predict()
- Standalone `python -m hy_worldplay.cli` entry point, invoked from the parity sub-venv


integrations/hy_worldplay/hy_worldplay/cli.py
liruilong940607, line 1: I think the goal is to trigger with `flashdreams-run` so this `cli.py` is not needed

wenqingw-nv:
Pushed an update addressing:

**1. Rebase on `main` (was `DIRTY` / `CONFLICTING`).** Rebased onto `origin/main` (now at `6523e9b`, PR #100). Only conflict was in `uv.lock`; resolved by taking `main`'s version and regenerating with `uv lock` so `flashdreams-hy-worldplay` is reinstated as a workspace member alongside the newly-landed `flashdreams-lingbot`. PR is `MERGEABLE` again.

**2. SKILL.md was indeed missing — now added** at [`agentic/skills/hy-worldplay-env-setup/SKILL.md`](https://github.com/NVIDIA/flashdreams/blob/wenqing/hy-worldplay-integration/agentic/skills/hy-worldplay-env-setup/SKILL.md), matching the existing `agentic/skills/{flashdreams-recipes,python-docstring-style}/` layout. It's an agent-facing decision tree for the two-venv layout, HF auth, the ~52 GB checkpoint provisioning, the run path, common errors, extension pitfalls, and the phase-2 horizon.

**3. The slug is now properly registered**:

- Made `RunnerConfig.pipeline` optional in `flashdreams/infra/runner.py` (default `None`, guarded both `Runner.__init__` blocks that touched it). Purely additive — every existing recipe passes `pipeline=` explicitly so behavior is unchanged. This reserves a slot for vendor-wrapper runners that don't yet have a flashdreams `StreamInferencePipeline` to drive — exactly HY's phase-1 situation.
- `HyWorldPlayWanI2VRunnerConfig` now subclasses `RunnerConfig` and registers via `[project.entry-points."flashdreams.runner_configs"]` (same shape as `wan21` / `self_forcing`). `HyWorldPlayWanI2VRunner` stays a plain class (not a `Runner` subclass) because phase-1 owns its own distributed setup deferred to upstream's `WanRunner`.
- Deleted `hy_worldplay/cli.py` + the `[project.scripts]` entry. The CLI surface is now `flashdreams-run hy-worldplay-wan-i2v-5b` everywhere — README, parity-check README + pyproject, `run-docker.sh`, repo-root README, smoke tests.
- Smoke tests swap `test_cli_module_imports` for `test_runner_config_is_runner_config_subclass` + `test_pipeline_is_none` + `test_entry_point_registered` (the last one exercises the same `importlib.metadata.entry_points` path the plugin loader uses at CLI startup, so misnamed entries fail in CPU pytest rather than at GPU run time).

Heavy deps (`sageattention`, `accelerate`, `cloudpickle`, `torch==2.11.*`) still live in the parity sub-venv only — `flashdreams-run` is invoked from there via `uv run --project integrations/hy_worldplay/tests/parity_check ...` (the sub-venv installs `flashdreams` editable so the console script resolves). When the native WAN 2.2 5B recipe lands in phase 2, the sub-venv collapses back into the main flashdreams venv but the slug stays the same — which is the whole point of registering it this way now.

wenqingw-nv:
Pushed an update addressing previous issues

liruilong940607:
/ok to test [e7da321](https://github.com/NVIDIA/flashdreams/pull/103/commits/e7da321863b05d8cf3ef742e93164e7435931b4a)

liruilong940607 (approved):

liruilong940607:
/ok to test [b756be6](https://github.com/NVIDIA/flashdreams/pull/103/commits/b756be6dac4e258a304ebe05d7d84e086a0c00b7)

liruilong940607:
Could use a lint fix: https://github.com/NVIDIA/flashdreams/pull/83#issuecomment-4474887618

liruilong940607:
/ok to test [c6dfe31](https://github.com/NVIDIA/flashdreams/pull/103/commits/c6dfe31eb37c39a70183d151b7abbb8a99c1fae1)

PR #105, opened 2026-05-19
Remove GL related stuffs in ludus renderer (since we have the CUDA code path now)

liruilong940607:
## Test Commands

```bash
# 0. Drop the cached JIT build to force a clean recompile of the C++/CUDA plugin
rm -rf /root/.cache/torch_extensions/py312_cu130/ludus_renderer_plugin

# 1. Extract the bundled example scene
python3 -c "import zipfile; zipfile.ZipFile('assets/example_data/alpadreams/clipgt.zip').extractall('/tmp/clipgt_scene')"

# 2. End-to-end CUDA single-frame render
#    (no '--cuda' flag any more — CUDA is the only backend)
uv run python integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py \
  --scene /tmp/clipgt_scene --frame 12

# 3. Smoke tests, unmarking @pytest.mark.manual so the GPU tests actually run
uv run pytest integrations/alpadreams/tests/test_ludus_renderer.py -v --runxfail
```

All tests passes. Visual is 

hdmap_scene_0012


liruilong940607:
/ok to test [a7f6ab7](https://github.com/NVIDIA/flashdreams/pull/105/commits/a7f6ab7b8acd0a1e6fa7c12600b3d60e321f5250)

integrations/alpadreams/tests/test_ludus_renderer.py
```diff
@@ -201,19 +199,16 @@ def test_ludus_renderer_wrapper_renders_frames(
...
     )
 
-    # Attach the ludus GPU scene
     scene_data = load_and_attach_ludus_scene(
         str(clipgt_scene_dir),
         scene_data,
         device=device,
     )
 
-    # Get camera model
```
pknowlesnv, line 211: 😂 classic agent, deleting all the comments. Don't worry. It's inevitable/too much trouble to avoid.

integrations/alpadreams/tests/test_ludus_renderer.py
```diff
@@ -53,58 +72,34 @@ def clipgt_scene_dir() -> Path:
 
 
 # ---------------------------------------------------------------------------
-# Low-level: LudusTimestampedContext renders without crashing
+# Module-import smoke test (CPU-only, runs in every GPU-free CI lane).
 # ---------------------------------------------------------------------------
 
 
-@pytest.mark.manual
-def test_ludus_timestamped_context_renders_frame(clipgt_scene_dir: Path) -> None:
```
pknowlesnv, line 61: this test seems to have been outright deleted. it doesn't sound GL-specific

integrations/alpadreams/ludus-renderer/README.md
```diff
@@ -148,16 +135,6 @@ GPU-side spatial culling ensures that rendering cost stays constant regardless o
...
 ```
 
-### Benchmarking
-
-```bash
-# Single scene benchmark
-uv run python examples/benchmark_renderer.py --scene example_data/test_hdmap --iters 10
-
-# Multi-camera benchmark (8 cameras per timestamp)
-uv run python examples/benchmark_renderer.py --scene example_data/test_hdmap --multicam
```
pknowlesnv, line 158: is GL required for benchmarking? there's a few files deleted in the examples directory that sound like they could work with the CUDA backend

integrations/alpadreams/tests/test_ludus_renderer.py
```diff
@@ -13,16 +13,35 @@
...
+production.
+
+Tests in this module:
+
+- ``test_alpadreams_ludus_renderer_imports_resolve`` (``ci_cpu``):
+  imports the alpadreams renderer module and verifies that every
+  module-level symbol still resolves after the GL pruning. Catches
+  accidental references to removed GL symbols at module load -- the
+  most likely regression mode -- on every CPU CI run with no GPU
+  required.
```
pknowlesnv: I'm curious if this test adds any more useful coverage beyond existing CUDA backend tests that would necessarily have to load the module anyway

integrations/alpadreams/ludus-renderer/ludus_renderer/scene_cache.py
```diff
@@ -93,20 +93,18 @@ def compute_format_hash() -> str:
     Combines C++ struct sizes (from pybind) with Python packing constants.
     Any struct or packing change auto-invalidates cache.
     """
+    # The legacy GL plugin exposed a ``get_struct_sizes`` helper that fed
+    # into this hash. The CUDA-only plugin does not, so the format hash
```
pknowlesnv: fed into the `cpp_sizes` field of this hash? prob not worth it 🤷 this comment will be deleted by the next agent that sees it anyway :)

integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py
```diff
@@ -424,79 +418,59 @@ def _render_jpg_all_cameras(ctx, scene_id, timestamps, all_poses, camera_type_id
...
-        from ludus_renderer import nvjpeg
-        if not nvjpeg.is_available():
-            print("Warning: nvJPEG not available, falling back to PNG")
-            return _render_png_all_cameras(
-                ctx, scene_id, timestamps, all_poses, camera_type_id,
-                output_dirs, camera_names, width, height, n_frames, n_cameras,
-                batch_size, device
-            )
-    
+    from ludus_renderer import nvjpeg
```
pknowlesnv, line 425: what's the convention for importing at the top for type safety vs importing dynamically for init perf?

pknowlesnv (approved):
There's a pile of deleted examples and at least one test that might not need to be deleted. If that's expected and as long as it still works with omni-dreams I guess it's fine. +1

integrations/alpadreams/tests/test_ludus_renderer.py
janickm: no, this was actually the GL one (the GL context was `LudusTimestampedContext`, while the cuda one is `LudusCudaTimestampedContext`) - resolving

integrations/alpadreams/tests/test_ludus_renderer.py
janickm: I guess the value is that we feed the renderer with data loaded from the

`alpadreams.conditioning.`

modules - I'd say we can still get rid of this one later / independent of this change [resolving for a later time]

integrations/alpadreams/ludus-renderer/README.md
janickm: agree, there was way too much removed here - I'll try to recuperate what makes sense for cuda

integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py
janickm: it looks like this is the local convention in the renderer code, so we should keep it

(personally, I tend to not prefer with this pattern, as doing local imports has a small runtime fee and more importantly this shadows general missing dependencies issues that only show up later (or not at all) if some code is not executed - I guess it's better to load all once and be sure that things are available thereafter)

integrations/alpadreams/ludus-renderer/ludus_renderer/scene_cache.py
janickm: I think this is mostly to keep existing on-disk hashes functional (as `cpp_sizes=unavailable` was the fallback if the gl version could not be imported, e.g., on non-EGL machines I suppose) - I'm also unsure to which extend we need this also, I'm removing this

janickm (changes requested):
looking into updates still, in particular reviving some of the overly ambitious deletions in the example scripts

integrations/alpadreams/ludus-renderer/examples/render_hdmap_scene.py
janickm: [resolving as not blocking]

integrations/alpadreams/tests/test_ludus_renderer.py
janickm: yeah, hopefully they adopt a bit better sure instinct moving forward :) [resolving]

integrations/alpadreams/ludus-renderer/README.md
janickm: I tried to revert / recuperate the ones that are cuda-specific in the latest version [resolving]

janickm:
[rebased]

janickm (approved):
verified latest version for the cuda only version

janickm:
Merging if CI is happy - note that this also resolves the gl-header compilation issues in the docker image @jatentaki was running into..

janickm:
[there is still an issue that we removed the `PyNvVideoCodec` dep although this is still used - checking]

janickm:
previously passing tests are now failing (I suspect that before the GL renderer had a different behavior compared to the cuda one that is now used), checking

PR #107, opened 2026-05-19
Fix Lingbot Overlay Resizing and Held Controls

gtong-nv:
# Fix Lingbot Overlay Resizing and Held Controls

## Summary

This MR updates the Lingbot WebRTC overlay so the telemetry bar remains readable across window sizes and held movement keys keep driving generation continuously. It builds on the previously merged overlay UI work and keeps the change scoped to the browser client CSS and input handling.

## Changes

- Restores continuous held-key behavior for Lingbot controls. When a movement key remains pressed, the client now sends another `keydown` after each `chunk_done` message so actions like holding `W` continue marching forward instead of only sending one keydown/keyup pair.
- Tracks held key order so repeated keydown messages are sent deterministically when multiple controls are held.
- Keeps key release behavior intact by clearing held state and sending keyup when keyboard or pointer input is released.
- Improves the bottom metrics bar responsiveness by adding explicit desktop width constraints and stacked layouts for smaller windows.
- Prevents telemetry labels and values from overlapping by using content-aware metric column minimums and removing ellipsis behavior when there is enough horizontal space.

### DIFF
--- a/integrations/lingbot/lingbot/webrtc/web/request_session.css
+++ b/integrations/lingbot/lingbot/webrtc/web/request_session.css
@@ -388,12 +388,12 @@ body[data-status="generating"] .connectButton {
   bottom: clamp(20px, 4vh, 40px);
   display: grid;
   grid-template-columns:
-    minmax(72px, 0.7fr)
-    minmax(120px, 1fr)
-    minmax(132px, 1.15fr)
-    minmax(72px, 0.7fr)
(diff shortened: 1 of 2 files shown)


liruilong940607:
/ok to test [d3058ec](https://github.com/NVIDIA/flashdreams/pull/107/commits/d3058ecf44261c79b845f64c528e141539f61c43)

PR #108, opened 2026-05-19
ci: restructure CI jobs and enable uv caching

janickm:
## Summary

Restructure the CI workflow to reduce runner requests and improve correctness of the publish gate.

## Commit 1: Merge lint into cpu job and gate publish on both runners

- **Remove standalone `lint` job** -- linting now runs as the first step inside the `cpu` job's `nvidia/cuda:13.2.1-cudnn-devel-ubuntu24.04` container. This gives lint access to `nvcc` and the full CUDA toolchain, and eliminates a separate runner request.
- **Install git before checkout** so `actions/checkout` does a proper `git clone` (pre-commit requires a git repository, not a tarball extraction).
- **Set `safe.directory`** to work around the container uid mismatch with the runner workspace.
- **Two-phase system deps**: install only git/python/curl before checkout for lint, then install heavier build deps (gcc, ninja, nccl, ffmpeg) only before tests.
- **Lint install** uses `--extra dev --group lint` with `--no-install-package transformer-engine-torch` (source-only, needs GPU arch to compile); ty gets all other imports for resolution.
- **Remove redundant `sync_version.py` call** from wheel build (already handled by the sync-version pre-commit hook during lint).
- **`publish-testpypi` depends on `[cpu, gpu]`** instead of just `[lint]`, ensuring the package is only published when both CPU and GPU tests pass.
- **Set `UV_LINK_MODE=copy`** to suppress hardlink warnings in containers.

## Commit 2: Enable uv cache persistence

- Enable `setup-uv` built-in cache support with per-job suffixes (`cpu`/`gpu`) to avoid cache pollution between jobs that install different dependency sets.
- The action handles save/restore/prune automatically, keyed on the `uv.lock` hash.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,9 @@ jobs:
 
       - name: Setup uv
         uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+          cache-suffix: "cpu"
 
       - name: Install lint dependencies
         run: |
@@ -107,6 +110,9 @@ jobs:
 
       - name: Setup uv
         uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+          cache-suffix: "gpu"
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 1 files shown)


janickm:
The action caching might still be a little fragile, but given we need to change the CI rules on the checks to run + @wilsonCernWq needs this in his PR I'm moving forward with this (FYI @liruilong940607 / @jatentaki also)

Issue #101, opened 2026-05-18
Improve checkpoint caching system


liruilong940607:
Even though we update Cosmos Reason1 hash commit in the code, the caching system still picks up the old checkpoint because it only checks the path exists. Need to be improved

PR #109, opened 2026-05-19
Rework HF caching

jatentaki:
Previously we had `should_use_local_files_only` which relied on an imperfect check for availability of some common huggingface files (like `tokenizer.json`) but caused crashes with corrupted cache directories. See this issue: https://github.com/NVIDIA/flashdreams/issues/101

This MR changes the logic: we instead have `maybe_download_hf_repo_on_rank0` which is called unconditionally and skips the actual download if not on rank0 or flags like HF_OFFLINE are set. Afterwards, all ranks load with unconditional `local_files_only=True`, relying on preceding `maybe_download_hf_repo_on_rank0` to have provided the required files locally.

Note this will not work in distributed settings where individual ranks have separate filesystems (like node-local raid arrays). Is this something we want to support? The current design sends 1 HF request per run (from rank0, checking if cache is up to date), supporting disjoint filesystems would probably require 1 HF request per rank instead.


liruilong940607:
/ok to test [e3bde29](https://github.com/NVIDIA/flashdreams/pull/109/commits/e3bde29b635aa41f207bfa78afe6037301e0e7ba)

PR #111, opened 2026-05-19
Disclose cudaraster + LodePNG redistributions for OSRB

jmccaffrey-nv:
Disclose the in-source third-party redistributions in NOTICE, reproduce the BSD-3-Clause and Zlib license texts under LICENSES/, and rename reuse.toml to REUSE.toml so REUSE 3.3 / reuse-tool 6.x actually applies the project's annotations. With the rename the project reaches full REUSE 3.3 compliance (464/464 files covered).

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -91,3 +91,27 @@ path = [
 precedence = "aggregate"
 SPDX-FileCopyrightText = "Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved."
 SPDX-License-Identifier = "Apache-2.0"
+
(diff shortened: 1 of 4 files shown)


REUSE.toml
janickm, line 1: I'm not familiar with the semantics of reuse toml files, but I guess the rename to all-caps is intentional?

janickm (approved):

REUSE.toml
janickm: Oh I found your "so REUSE 3.3 / reuse-tool 6.x actually applies the project's annotations." remark, all good then 👍

PR #115, opened 2026-05-19
chore: enable uv managed mode across pyproject files

jarcherNV:
Set [tool.uv] managed = true in all flashdreams pyproject.toml files to prepare
for vulnerability scanning workflows.


gtong-nv (approved):
LGTM

PR #116, opened 2026-05-19
Fix vulnerable dependency pins from scan report

jarcherNV:
Bump transformers to 5.8.1, enforce urllib3 >=2.7.0, and constrain the
self-forcing parity environment to patched protobuf 6.33.x. Regenerate uv
locks for the root workspace and parity checks.

### DIFF
--- a/flashdreams/pyproject.toml
+++ b/flashdreams/pyproject.toml
@@ -36,8 +36,10 @@ dependencies = [
     "nvidia-ml-py>=12.0",
     "safetensors>=0.4",
     "tqdm>=4.60",
-    "transformers==4.57.6",
+    "transformers==5.8.1",
     "torchvision",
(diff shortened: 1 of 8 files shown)


jmccaffrey-nv:
LGTM

flashdreams/pyproject.toml
```diff
@@ -36,8 +36,10 @@ dependencies = [
     "nvidia-ml-py>=12.0",
     "safetensors>=0.4",
     "tqdm>=4.60",
-    "transformers==4.57.6",
+    "transformers==5.8.1",
```
liruilong940607: Would it be possible to relax this to like >= sth?

jarcherNV: Done.

PR #118, opened 2026-05-20
restored hy world changes

liruilong940607:
restore the changes in #103 that got accidentally merged then reverted


agentic/skills/hy-worldplay-env-setup/SKILL.md
liruilong940607, line 1: Actually after reviewing this file I think we shouldn't check in this file -- it is just a inteegration plan not a general skill for the reop

flashdreams/flashdreams/infra/runner.py
```diff
@@ -55,8 +55,15 @@ class RunnerConfig(InstantiateConfig):
     per-runner ``--help`` (it's metadata, not a knob); a non-empty
     value is enforced for in-tree runners by the registry test."""
 
-    pipeline: StreamInferencePipelineConfig
-    """Wrapped pipeline config; the runner instantiates and drives it."""
+    pipeline: StreamInferencePipelineConfig | None = None
```
liruilong940607, line 58: I think we shouldn't make it to optional. We want to enforce our infra structure that a runner must contain a pipeline [Can be inheritance of the StreamInferencePipelineConfig]

In your case you want a "hack" around you can inherit the `StreamInferencePipelineConfig` and `StreamInferencePipeline` override the `diffusion_model` file in the inheritated config make `StreamInferencePipeline` effectively an no op class that create nothing.

flashdreams/flashdreams/infra/runner.py
```diff
@@ -55,8 +55,15 @@ class RunnerConfig(InstantiateConfig):
...
+    """Wrapped pipeline config; the runner instantiates and drives it.
+
+    ``None`` is reserved for vendor-wrapper runners that do not yet have
+    a flashdreams :class:`StreamInferencePipeline` to drive (typically a
+    phase-1 plugin wrapping an upstream tree's own pipeline). Such
+    runners must override :meth:`Runner.__init__` -- or supply a
+    bespoke ``_target`` -- because the base ``__init__`` skips
+    pipeline construction and leaves :attr:`Runner.pipeline` unset."""
```
liruilong940607, line 66: Also those are quite obvious bot generated doc strings which has little useful information. you could use the doc string skill carried in the repo to help with doc string polishing in general.

for this specific case we can just revert the change if we don't allow pipeline to be None.

liruilong940607:
Closing in favor of #155

PR #122, opened 2026-05-20
Implement webRTC viewer for omnidreams

jatentaki:
* Implements a webRTC viewer for OmniDreams
* Moves webRTC logic common to omnidreams & lingbot to `flashdreams.serving`.
  * This is gated by the `flashdreams[serving]` feature flag
  * This is a "quick and dirty" solution. I am not convinced this is the right level of abstraction, e.g. the keyboard input parsing and integration likely shouldn't belong to flashdreams core as it is very model-specific (in fact the two models differ in their RDF vs FLU conventions)
* Fixes a bug where failures in `CUDAGraph` replay would emit entirely useless stack traces. My change replays the graph once immediately after capture, avoiding deferred failures.
* Adds a server warmup phase (with arbitrary WSAD requests) for both lingbot and omnidreams such that all triton kernels can pre-compile before the first user is connected.
* Adds a client heartbeat, disconnects and closes session a user disconnects (freeing it up for other users)

Missing:
* Physics in alpasim (currently ego stays at "flat" level at all times, will need to integrate basic physics or cherry-pick scenes with no significant height variations)

Usage for omnidreams:
(requires running `bash assets/download.sh` beforehand)
```
uv run torchrun --nproc_per_node 4 -m alpadreams.webrtc.server --pipeline_config_name alpadreams-sv-2steps-chunk2-loc6-lightvae-lighttae-perf
```

Usage for lingbot:
```
uv run torchrun --nproc_per_node=4 -m lingbot.webrtc.server --config_name lingbot-world-fast-flash
```


integrations/alpadreams/tests/test_grpc_server_integration.py
jatentaki, line 1: This is hacky and should be reworked to avoid this pythonpath and environment handling

integrations/alpadreams/alpadreams/webrtc/session.py
```diff
@@ -0,0 +1,1041 @@
...
+        # for throughput: Alpadreams uses CUDA graph capture/replay through
+        # torch.compile/cuDNN, and the captured state appears to depend on
+        # thread-local CUDA/cuDNN context. Replacing this with
+        # ``asyncio.to_thread`` lets the default executor move initialize,
+        # warmup, and generation calls across workers; that was observed to
+        # fail after a few chunks with
+        # CUDNN_STATUS_INTERNAL_ERROR_DEVICE_ALLOCATION_FAILED followed by
+        # cudaErrorStreamCaptureInvalidated during capture_end.
+        self._executor = ThreadPoolExecutor(
+            max_workers=1,
```
jatentaki, line 186: This is a painful lesson @wilsonCernWq @liruilong940607 - turns out that letting cuda graphs move between threads causes crashes with no clear traceback/reason.

flashdreams/flashdreams/infra/cuda_graph.py
jatentaki, line 1: During debugging I was somehow storing incorrect CUDA graphs which would fail at first replay without any apparent connection with how they were created. This change replays them immediately (to fail fast) and only binds to `self` once known to work.

liruilong940607 (approved):
LGTM! some lessons learnt from your comments @jatentaki !

liruilong940607:
/ok to test [df13c42](https://github.com/NVIDIA/flashdreams/pull/122/commits/df13c425f6d6d363e0a63e7a3b9164671e8b2c7e)

liruilong940607:
/ok to test [80a553d](https://github.com/NVIDIA/flashdreams/pull/122/commits/80a553d0228839e80511379452c096c0c0737c0e)

integrations/alpadreams/alpadreams/webrtc/session.py
wilsonCernWq: wow ...

wilsonCernWq (approved):

liruilong940607:
/ok to test [010f8a6](https://github.com/NVIDIA/flashdreams/pull/122/commits/010f8a6333a82ffa7920915e42690bba115b79f1)

PR #126, opened 2026-05-20
Add OmniDreams single-view third-party submodules

mchock-nv:
## Summary
- Add CUTLASS, SageAttention, and SpargeAttn as pinned submodules under integrations/alpadreams/omnidreams_singleview/3rdparty
- Match the dependency SHAs used by TrueSight
- Update NOTICE and REUSE metadata for the source-level redistributions

## Validation
- git diff --check origin/main...HEAD
- git submodule status --recursive integrations/alpadreams/omnidreams_singleview/3rdparty/{cutlass,SageAttention,SpargeAttn}
- CodeRabbit local review attempted; connection timed out after 120s without findings

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "integrations/alpadreams/omnidreams_singleview/3rdparty/cutlass"]
 	path = integrations/alpadreams/omnidreams_singleview/3rdparty/cutlass
 	url = https://github.com/NVIDIA/cutlass.git
+[submodule "integrations/alpadreams/omnidreams_singleview/3rdparty/SageAttention"]
+	path = integrations/alpadreams/omnidreams_singleview/3rdparty/SageAttention
+	url = https://github.com/thu-ml/SageAttention.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "integrations/alpadreams/omnidreams_singleview/3rdparty/SageAttention"]
 	path = integrations/alpadreams/omnidreams_singleview/3rdparty/SageAttention
 	url = https://github.com/thu-ml/SageAttention.git
+[submodule "integrations/alpadreams/omnidreams_singleview/3rdparty/SpargeAttn"]
+	path = integrations/alpadreams/omnidreams_singleview/3rdparty/SpargeAttn
+	url = https://github.com/thu-ml/SpargeAttn.git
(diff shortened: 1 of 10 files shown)


jmccaffrey-nv (approved):

integrations/alpadreams/omnidreams_singleview/3rdparty/SageAttention
jmccaffrey-nv, line 1: Submodule paths still reference alpadreams/ (renamed on main)

  All three paths in .gitmodules, REUSE.toml, and NOTICE are under integrations/alpadreams/omnidreams_singleview/3rdparty/. Main renamed alpadreams → omnidreams in
   PR #128 (7a4e8b9), two commits after this branch's merge base (6d33cc6).

integrations/alpadreams/omnidreams_singleview/3rdparty/cutlass
jmccaffrey-nv, line 1: CUTLASS (f3fde58) bundles python/CuTeDSL/, which its LICENSE.txt carves out from BSD-3-Clause. That subtree ships under python/CuTeDSL/EULA.txt  .

Will we need CuTeDSL, with future optimization patches ?
If not, we can exclude it.

.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "integrations/alpadreams/omnidreams_singleview/3rdparty/cutlass"]
```
janickm, line 1: could we please try to not use submodules in this repository and integrate these in the same way we integrate all other external dependencies? we tried very hard to not use submodules / git-lfs until now, this doesn't feel like a valid use-case for submodules either (also, as aside) these are not version / SHA tagged)

janickm (changes requested):

.gitmodules
jmccaffrey-nv: A particular issue for CUTLASS is we need to apply a bugfix patch which isn't upstream, which is possible with submodules but we couldn't figure how to solve with uv.sources .

mchock-nv:
Marking this PR as draft so it cannot be merged while we rework the external-source approach.

This submodule-based branch is superseded by the planned script-based third-party source import. The replacement will download the pinned external repositories at build/setup time, use the current `integrations/omnidreams/...` layout, apply the CUTLASS patch in a staged source tree, and exclude or otherwise address CUTLASS CuTeDSL licensing scope before review.

Please do not merge this PR as-is. Once the replacement PR exists, we can link it here and close this one.

mchock-nv:
Closing this PR as superseded. The replacement PR is #151, which carries the rewritten third-party source-sync mechanism and the native foundation changes in the current intended form for review.

PR #127, opened 2026-05-20
Remove stdout logging when skipping KV cache finalize

gtong-nv:
Stop printing "Skipping KV cache finalize" when `skip_finalize_kv_cache=True`
Cosmos and Alpadreams transformers now return early with no stdout noise.


gtong-nv:
/ok to test [3d719db](https://github.com/NVIDIA/flashdreams/pull/127/commits/3d719db845599a51498539fcd397e53e0442683b)

liruilong940607 (approved):

PR #130, opened 2026-05-20
Silence ludus-renderer JIT compile warnings via -w

pknowlesnv:
Adds -w to ludus_renderer plugin's gcc/nvcc flags so the C++/CUDA JIT build no longer floods stderr with header warnings (Math.hpp et al.) when consumers like the omnidreams webrtc server import the plugin. Includes a comment flagging that this hides all diagnostics and the underlying warnings should ideally be fixed rather than ignored.


janickm:
pushed an update that actually makes all (~most, up to a tricky `__device__` redeclaration one) errors + fixed all warnings instead - this way we don't blindly silence all, but respect the compiler's complaints (we don't have all warnings enabled anyway though..)

PR #131, opened 2026-05-21
Add flashVSR uplifting server

gtong-nv:
## Summary

Adds a standalone FlashVSR gRPC upsampler server with an optional browser viewer.

- Adds `UltraFlashVSRService` protobuf API for FlashVSR upsampling.
- Supports unary chunk flow: `StartSession` / `UpscaleChunk` / `EndSession`.
- Supports bidirectional streaming via `UpscaleVideo` for lower-latency live ingest.
- Adds an HTTP MJPEG viewer that shows input frames and upsampled output frames in the browser.
- Adds a 30 fps test client script, `feed_frames.sh`, for looping a video into the server.
- Adds Python client utilities and focused client tests.
- Documents server launch, viewer usage, and smoke-test commands in `integrations/flashvsr/README.md`.

## Usage

Start the standalone FlashVSR server:

```bash
uv sync --package flashdreams-flashvsr
uv run --no-sync --package flashdreams-flashvsr python -m flashvsr.grpc.server \
    --port 50051 \
    --viewer_port 8080 \
    --cuda_graph \
    --attention_mode sparse \
    --sparse_ratio 1.5
```

Open the viewer:

```text
http://<server-host>:8080/
```

Feed a video at live cadence:

```bash
./feed_frames.sh --server localhost:50051 --input /path/to/clip.mp4
```

Finite smoke test:

```bash
./feed_frames.sh --server localhost:50051 --input /path/to/clip.mp4 --max_chunks 4
```

## Notes

When the viewer is enabled, gRPC output frames are omitted by default to reduce server-to-client bandwidth. Pass `--viewer_return_grpc_frames` if the client also needs raw RGB output payloads.

## Testing

```bash
PYTHONDONTWRITEBYTECODE=1 uv run --package flashdreams-flashvsr --with pytest \
    python -m pytest integrations/flashvsr/tests/test_grpc_client.py -q

### CODE REVIEW

**Comment by gtong-nv** [comment]
/ok to test [b787c21](https://github.com/NVIDIA/flashdreams/pull/131/commits/b787c2168eaf5589f2d4a99c3a9f0777a03676d1)

File: integrations/flashvsr/flashvsr/grpc/server.py [hunk 1]
```diff
@@ -0,0 +1,1396 @@
...
+
+
+@dataclass
+class _DecodedStreamRequest:
+    request: pb2.UpscaleChunkRequest
+    frames: np.ndarray
+    decode_ms: float
+
+
+class UltraFlashVSRServicer(pb2_grpc.UltraFlashVSRServiceServicer):
```
wilsonCernWq: Maybe we should unify the name to FlashVSR?

integrations/flashvsr/flashvsr/grpc/server.py
```diff
@@ -0,0 +1,1396 @@
...
+            return_frames=return_frames,
+            pre_decode_ms=decode_ms,
+            on_after_infer=on_after_infer,
+        )
+
+    # ------------------------------------------------------------------
+    # RPC implementations
+    # ------------------------------------------------------------------
+
+    def GetStatus(self, request, context):
```
wilsonCernWq: can we also unify function name style? I think we always use lower case? not camel?

feed_frames.sh
```diff
@@ -0,0 +1,31 @@
...
+export PYTHONPATH="${REPO_ROOT}/integrations/flashvsr:${REPO_ROOT}/flashdreams${PYTHONPATH:+:${PYTHONPATH}}"
+export PYTHONDONTWRITEBYTECODE="${PYTHONDONTWRITEBYTECODE:-1}"
+
+if [[ $# -gt 0 && "$1" != -* ]]; then
+    input_path="$1"
+    shift
+    set -- --input "${input_path}" "$@"
+fi
+
+exec uv run --no-sync python -m flashvsr.grpc.continuous_stream_client "$@"
```
wilsonCernWq: this script is just for convenient right? do we really want to ship it?

integrations/flashvsr/flashvsr/grpc/streaming_view.py
```diff
@@ -0,0 +1,472 @@
...
+                    q.get_nowait()
+                except queue.Empty:
+                    pass
+                try:
+                    q.put_nowait(jpeg)
+                except queue.Full:
+                    pass
+
+
+class StreamingViewer:
```
wilsonCernWq, line 148: Nit: do we really need a separate file to hold the class? It just confused me for a while.

integrations/flashvsr/README.md
```diff
@@ -113,6 +113,55 @@ sparse presets. Use smaller inputs, `--chunk-size 8`, fewer GPUs, or override
...
+./feed_frames.sh --server localhost:50051 --input /path/to/clip.mp4
+
+# finite smoke test:
+./feed_frames.sh --server localhost:50051 --input /path/to/clip.mp4 --max_chunks 4
+```
+
+For a save-to-disk test client, use:
+
+```bash
+PYTHONPATH=integrations/flashvsr:flashdreams \
```
wilsonCernWq, line 167: do we really need to ship this mode to user? User can also directly run the inference script to convert a video, so this modality sounds more like a debugging tool?

wilsonCernWq:
Good stuff, left some minor comments. I would prefer we simplify the code structure to improve clarity for user?
Also CI needs to be fixed.

integrations/flashvsr/flashvsr/grpc/server.py
```diff
@@ -0,0 +1,1396 @@
...
+            )
+        return "sparse"
+
+    if requested != "auto":
+        raise ValueError(f"unknown attention mode: {requested}")
+
+    if sparse_available:
+        return "sparse"
+
+    log.warning(
```
wilsonCernWq: i don't know if we really want this check? I usually feel when the user's request cannot be fulfilled, we just fail loud rather than fallback?

And `block_sparse_attn` should always be available for flashvsr hmmm

integrations/flashvsr/flashvsr/grpc/uplift_server.py
```diff
@@ -0,0 +1,1396 @@
...
+            cache_holder[0] = None
+            log.info("UpscaleVideo stream %s: cache released", session_id_box[0])
+
+
+# ---------------------------------------------------------------------------
+# Entry point
+# ---------------------------------------------------------------------------
+
+
+def main():
```
wilsonCernWq, line 1151: This server does not support MGPU right? Do you feel we can consider this modality? I know this is out of the scope for CVPR...

integrations/flashvsr/flashvsr/grpc/streaming_view.py
gtong-nv: I would prefer this class in a separate file because it serves the http server for the frontend viewer, and better to be separated from the gRPC server

integrations/flashvsr/flashvsr/grpc/uplift_server.py
```diff
@@ -0,0 +1,1396 @@
...
+        type=int,
+        default=DEFAULT_STREAM_CHUNK_QUEUE_DEPTH,
+        help=(
+            "UpscaleVideo only: max UpscaleChunkResponse messages queued between GPU "
+            "completion and gRPC send (FIFO). Larger values overlap send with the next "
+            "chunk on the GPU (default: %(default)s)."
+        ),
+    )
+    parser.add_argument(
+        "--no_combine_8_frame_chunks",
```
wilsonCernWq, line 1234: why this is needed?

integrations/flashvsr/flashvsr/grpc/server.py
gtong-nv: Thanks! updated.

integrations/flashvsr/README.md
gtong-nv: yea, it's mostly for debugging this server, but also useful if the flashVSR service running in another machine (exposed by the public IP), and the client want to access it remotely

feed_frames.sh
gtong-nv: Thanks. This is for debugging. I will take this out.

integrations/flashvsr/flashvsr/grpc/streaming_view.py
wilsonCernWq: should we call it streaming_viewer then? idk, i feel "-er" makes it sounds more like a component rather than an entry point.

integrations/flashvsr/README.md
wilsonCernWq: I see, makes sense, maybe we can advertise this use case in the readme?

wilsonCernWq:
So the overall structure of the viewer is

integrations/flashvsr/flashvsr/grpc/streaming_view.py
gtong-nv: It's indeed a component to some degree, as this http server can't be started by itself as an entry point, rather, as a byproduct of `flashvsr.grpc.uplift_server` when `--viewer_port` is specified.

integrations/flashvsr/flashvsr/grpc/uplift_server.py
gtong-nv: I updated the help text to make it clear
```
            "Run FlashVSR on every incoming 8-frame request directly, instead "
            "of the default behavior of coalescing two 8-frame requests into "
            "one 16-frame FlashVSR call (with a 13-frame cold start on the "
            "first call). Smaller chunks reduce per-request latency because "
            "the server does not wait for a second 8-frame request before "
            "running the model, but usually hurt throughput since FlashVSR "
            "is most efficient at 16-frame chunks."
```

integrations/flashvsr/flashvsr/grpc/uplift_server.py
gtong-nv: yea, it does not consider multiple GPUs now. I will follow up that later

integrations/flashvsr/README.md
gtong-nv: yea, I updated the readme to advertise the server mode

integrations/flashvsr/flashvsr/grpc/server.py
gtong-nv: done. changed to snake style

integrations/flashvsr/flashvsr/grpc/server.py
gtong-nv: addressed. Make it fail if `block_sparse_attn` not found

gtong-nv:
/ok to test [102736c](https://github.com/NVIDIA/flashdreams/pull/131/commits/102736cee1f23db6bc19b97db272195e3c4c94c5)

PR #138, opened 2026-05-21
ci: add merge_group triggers for GitHub merge queue

jmccaffrey-nv:
- Add `merge_group:` trigger to `ci.yml` and `doc.yml` so the existing
    lint / cpu / gpu / docs jobs also run against GitHub's queue ref
    (`gh-readonly-queue/main/pr-<n>-<sha>`). Job names are unchanged so
    existing required-check rules keep working.
  - Tighten the two deploy-side gates in `doc.yml` from `!= 'pull_request'`
    to an explicit allow-list (`push || release || workflow_dispatch`) so
    queue runs build the docs but don't publish to `gh-pages`.
  - `publish-testpypi` was already gated on
    `github.ref == 'refs/heads/main' && github.event_name == 'push'`, so
    it correctly stays out of queue runs without any change.
  - Document the new "Merge when ready" flow in `CONTRIBUTING.md` and
    warn maintainers not to leave "Require branches to be up to date
    before merging" enabled alongside the queue.

  ## Branch-protection action required after merge
  The workflow changes alone don't activate the queue. After this PR
  lands, in **Settings → Branches → `main` → Edit rule**:
  1. Turn **on** "Require merge queue"
  2. Turn **off** "Require branches to be up to date before merging"
  3. Keep `lint`, `cpu`, `gpu`, and the docs job in required checks
  4. Optional: set queue "Maximum pull requests to build" >1 for parallel

PR #140, opened 2026-05-21
ci(reuse-lint): add merge_group trigger for merge queue

janickm:
The `reuse-lint` workflow (`REUSE 3.3 compliance (repo-wide)` and
`OSRB collateral sanity check`) is a required check in branch protection
but was not updated in #138 to fire on `merge_group` events. This causes
the merge queue to stall indefinitely waiting for checks that never start.

Adds the same `merge_group: branches: [main]` trigger that `ci.yml` and
`doc.yml` received in #138.

**Impact without this fix:** any PR entering the merge queue hangs until
timeout because the two reuse-lint jobs never run on the queue ref.

### DIFF
--- a/.github/workflows/reuse-lint.yml
+++ b/.github/workflows/reuse-lint.yml
@@ -2,6 +2,8 @@ name: reuse-lint
 
 on:
   pull_request:
+  merge_group:
+    branches: [main]
   push:
     branches:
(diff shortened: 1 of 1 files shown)


jmccaffrey-nv:
LGTM

PR #143, opened 2026-05-21
Add parity check and profiling against fastvideo

liruilong940607:
On
- Wan2.1 T2V
- Cosmos Predict2.5 T2V
- Self-forcing T2V
- Causal Wan2.2 T2V [FastVideo version]

Also skip sync version for the path `integration/**/tests/` where FastVideo / other official code would downloaded to on-the-fly.


liruilong940607:
/ok to test [f8d544f](https://github.com/NVIDIA/flashdreams/pull/143/commits/f8d544f7624a205a7724e614c58c20bfde129e21)

PR #145, opened 2026-05-22
Benchmark against fastvideo

liruilong940607:
Configuration fix and alignment between flashdreams v.s. fashvideo v.s. official for cosmos-predict2 bidirectional. The similar runtime now makes sense [before it has a huge difference]
 
Setting: Cosmos-Predict2.5-2B T2V

DiT (per step w/ CFG). 720P. 93 frames

Screenshot 2026-05-21 at 5 36 08 PM

all uses torch.compile with CUDNN attn backend.


liruilong940607:
/ok to test [ff59092](https://github.com/NVIDIA/flashdreams/pull/145/commits/ff590921bb41483e9531e12ad146c9a70b319c18)

PR #149, opened 2026-05-22
Move from stdlib logging to loguru

jatentaki:
Standardizes on using `loguru` over `stdlib` (`loguru` is configured to emit only on rank0 when in multiprocess contexts, which is helpful). Additionally expands some diagnostics in omnidreams webrtc path.


wilsonCernWq:
Oh nice I bumped into the same issue as well, vote for it

wilsonCernWq (approved):
LGTM, I would just add some note in the merge message that the PR also includes some additional profiling logs?

PR #182, opened 2026-05-27
Add omnidreams-interactive-drive sample under integrations/omnidreams

wlewNV:
image

## Summary

Inlines the `interactive-drive` driving demo (previously at [`nv-tlabs/omni-dreams`](https://github.com/nv-tlabs/omni-dreams) `samples/interactive-drive`) as a `flashdreams` workspace member under [`integrations/omnidreams/samples/interactive-drive/`](integrations/omnidreams/samples/interactive-drive). The standalone version pinned `flashdreams` and `flashdreams-omnidreams` via `git+ssh://...` URLs and had already drifted from current `main` (`LudusTimestampedContext` was renamed to `LudusCudaTimestampedContext` upstream); this move catches that class of bug at lock time going forward.

## End-to-end UX

Single sync, single run, no `--extra` flags, no `cd` into the sample dir — same convention as every other `integrations/*` member:

```bash
# From the flashdreams workspace root, first time:
uv sync --package omnidreams-interactive-drive
export HF_TOKEN=<your-hf-token>      # for nvidia/omni-dreams-scenes

# Run the demo (default scene auto-stages from HF on first launch):
uv run --package omnidreams-interactive-drive interactive-drive --stream-mjpeg :8080
# or, with the slangpy HUD on a graphical desktop:
uv run --package omnidreams-interactive-drive interactive-drive
```

For multi-scene staging or pre-warming the ~14 GB Cosmos-Reason1 text encoder ahead of first launch:

```bash
uv run --package omnidreams-interactive-drive interactive-drive-prepare \
  --scene-uuid clipgt-01d503d4-449b-46fc-8d78-9085e70d3554
# Or pre-warm only the text encoder, skip scene staging:
uv run --package omnidreams-interactive-drive interactive-drive-prepare --skip-scene
```

Dev / test tooling:

```bash
uv sync --package omnidreams-interactive-drive --extra dev
uv run --no-sync --package omnidreams-interactive-drive pytest \
  integrations/omnidreams/samples/interactive-drive/tests -m "not gpu"
integrations/omnidreams/samples/interactive-drive/scripts/check.sh
```

### New workspace member
- [`integrations/omnidreams/samples/interactive-drive/`](integrations/omnidreams/samples/interactive-drive) — 83 new files (sample source, tests, configs, assets, README).
- Registered in root [`pyproject.toml`](pyproject.toml) under `[tool.uv.workspace] members` alongside `ludus-renderer`. pyright/ty `extra-paths` updated to include its `src/` layout (the first `src`-layout member in this workspace).
- Distribution name **`omnidreams-interactive-drive`** (matches the `flashdreams-omnidreams` sibling naming). Executables stay short: `interactive-drive`, `interactive-drive-prepare`.

### Dependency consolidation
- `ui` and `world-model` extras folded into the base `[project] dependencies`. Only `dev` extra remains. One sync pulls the whole demo runtime.
- `prepare.py` moved into the package as `src/interactive_drive/prepare.py` and exposed as the `interactive-drive-prepare` script entry. The old top-level script is gone.

### Path / CLI ergonomics
- CLI defaults (`--scene`, `--manifest`, `--scene-dir`, `--wheel-profiles-dir`) now resolve from `__file__` instead of cwd, so workspace-root invocation works without any path arguments.
- New `demo._maybe_autostage_scene()` hook auto-downloads the default scene on first launch if missing and `HF_TOKEN` is set. Only fires for the bundled `assets/scenes/clipgt-*.usdz` filenames; user-pointed external paths fall through to the normal "file not found" error with a pointer at `interactive-drive-prepare`.

### CLI surface cleanup
- `--backend` hidden from `--help` via `argparse.SUPPRESS`. Default is the omnidreams pipeline. The raster path stays available for the existing smoke test ([`test_app_smoke.py`](integrations/omnidreams/samples/interactive-drive/tests/test_app_smoke.py) uses `--backend raster` to exercise CI without HF + CUDA-graph compile).
- Internal value `world_model` renamed to `omnidreams` across `cli.py`, `demo.py`, `app.py`, `config.py`, the `WorldModelRenderBackend` log labels, and error messages.

### Workspace hygiene
- Workspace-wide [`.pre-commit-config.yaml`](.pre-commit-config.yaml) excludes the sample subtree so the sample-local hook config (preview ruff rules + CPY001 copyright check + pyright with sample-specific excludes) doesn't fight workspace-wide hooks. Mirrors the existing `ludus-renderer` precedent.
- Sample's own `.pre-commit-config.yaml` and `scripts/check.sh` use `uv --project "$WORKSPACE_ROOT"` so they keep resolving the workspace correctly regardless of cwd.

### API drift fix included
- `rasterizer.py`: `from ludus_renderer.torch import LudusTimestampedContext` → `LudusCudaTimestampedContext`. Upstream rename that the pinned standalone hadn't picked up.

## Why bare dep names instead of `workspace = true`

The sample lives at `integrations/omnidreams/samples/interactive-drive/` — nested inside another workspace member (`flashdreams-omnidreams`). uv's workspace-discovery walk stops at the enclosing `integrations/omnidreams/pyproject.toml` and refuses to continue up ([astral-sh/uv#16640](https://github.com/astral-sh/uv/issues/16640); workspace.rs explicitly comments "we ignore all `albatross` is doing and any potential workspace it might be contained in"). The workaround is to leave the dep names bare in `[project] dependencies` — `flashdreams`, `flashdreams-omnidreams`, `ludus-renderer` aren't on PyPI, so uv resolves them against the workspace by default. No `[tool.uv.sources]` block needed in the sample's pyproject.


jarcherNV:
I will stand in for greptile since it's not yet enabled.
Note: These comments are mostly (not entirely) from codex so some may or may not be valid :p

* New sample tests don't have the repo-required ci_cpu / ci_gpu / manual markers. The root pytest config loads marker enforcement, so root CI can fail collection or never exercise this sample as intended. Representative unmarked test: test_camera.py (line 12); root enforcement is enabled in pyproject.toml (line 142).
* HUD keyboard driving is tied to drawing the right-side panel. When the window is narrow enough for panel_w to become 0, _draw_panel() is skipped, so _keyboard_drive.update() never publishes key state; releases can also leave stale drive commands active. See slangpy_hud_presenter.py (line 657), slangpy_hud_presenter.py (line 686), and slangpy_hud_presenter.py (line 807). I don't know if this is really a big deal in practice though.
* Wheel support is documented and enabled by default, but no configs/wheels/ profiles are committed. The default path points there, _load_wheel_profiles() returns an empty tuple, and --wheel-profile thrustmaster from the README cannot work. See demo.py (line 530), demo.py (line 992), and README.md (line 244).
* Prompt variant discovery is inconsistent. The HUD accepts prompt_1.txt as variant 1, but scene_loader._discover_prompts() stores it as _1, so selecting variant 1 falls back to the default prompt. See scene_loader.py (line 90) and demo.py (line 876).
* The committed VS Code launch configs use "module": "interactive-drive", but the importable module is interactive_drive; interactive-drive is only the console script name. Debug launch will fail before reaching the app. See launch.json (line 8).

Also, do we want to commit the vs code configs? And I think I noticed a bunch of files in there that only had the nvidia copyright header, do we need those too?

wlewNV:
**1. Sample tests missing `ci_cpu` / `ci_gpu` / `manual` markers** — Fixed in [e942575](https://github.com/NVIDIA/flashdreams/commit/e942575). The sample carries its own `gpu` / `xvfb` markers (separate tier convention from the workspace's), so adding the sample tests dir to root `[tool.pytest.ini_options] norecursedirs` is cleaner than retrofitting ~70 markers. Sample tests still run via `uv run --package omnidreams-interactive-drive pytest ...` and `scripts/check.sh`.

**2. HUD keyboard driving tied to `_draw_panel`** — Fixed in [687bd02](https://github.com/NVIDIA/flashdreams/commit/687bd02). Extracted `_poll_drive_state()` from `_draw_panel`, call it (and `_update_speed`) unconditionally from `_render_canvas`, and pass the resulting `wheel_state` into `_draw_panel`. Narrow windows / mid-keypress resizes now behave like the bare `--no-hud` viewer for input — releases land, no latched commands.

**3. Wheel profiles documented but not shipped** — Fixed in [e942575](https://github.com/NVIDIA/flashdreams/commit/e942575). Decided not to ship `thrustmaster.yaml` (licensing). README rewritten to document the BYO-YAML workflow (drop a profile into `configs/wheels/`), points at `--no-wheel` / `--wheel-device` for the no-profile path, notes keyboard-only driving works without any profile. `_load_wheel_profiles()` already returns `tuple()` cleanly for the empty / missing-dir case.

**4. Prompt variant discovery `_1` vs `1`** — Good catch; would have shipped silently broken on real HF scenes. Fixed in [e942575](https://github.com/NVIDIA/flashdreams/commit/e942575) by rewriting `_discover_prompts` to use the underscore-required convention that `_discover_first_images` already had (`prompt.txt` → default, `prompt_<N>.txt` → variant `<N>`, bare `promptN.txt` rejected). Also patched the synthetic-scene fixture — it was inconsistently writing `prompt1.txt` (no underscore) while writing `first_image_1.png` (with underscore). It was only "passing" because the buggy parser happened to round-trip the no-underscore form. Now writes `prompt_1.txt` matching the convention.

**5. VS Code `launch.json` `"module": "interactive-drive"` broken + should we commit?** — Deleted in [e942575](https://github.com/NVIDIA/flashdreams/commit/e942575). Configs were user-specific. Also removed the `!.vscode/launch.json` whitelist from the sample's `.gitignore` so stray local configs don't get re-tracked.

**6. Files with only NVIDIA copyright header** — Kept. Those are 6 empty `__init__.py` files (`assets/`, `input/`, `runtime/`, `simulation/`, `video_model/`, `world_model/`) that are required Python package markers. Each has the SPDX header plus a one-line "what this package is" comment. Standard convention.

**7. CI lint failures** — Fixed in [0f35dc8](https://github.com/NVIDIA/flashdreams/commit/0f35dc8). Two issues:

- `sync-version` hook auto-updated the sample's `pyproject.toml` from `0.1.0` to `0.1.0a2` (the workspace's single-source version). Bumped to match so the hook is a no-op next run.
- `ty` reported 38 errors on the sample, mostly in the flashdreams / slangpy / numpy interop surface that the sample's own pyright config already documents as known-untyped. The workspace-wide `ty` doesn't honor the sample's pyright excludes, so added `integrations/omnidreams/samples/interactive-drive/**` to `[tool.ty.src] exclude`. Mirrors the same pattern used for `norecursedirs` and the pre-commit `exclude` block — the sample owns its own type-checking via `scripts/check.sh`.

Verified locally: 69 / 69 sample tests pass (`pytest -m "not gpu and not xvfb"`), `ty check` from workspace root reports 0 diagnostics in the sample tree, root `pytest --collect-only` (393 tests) collects nothing from the sample.

jarcherNV:
A few findings still:

* Sample tests still are not running in CPU/GPU CI. The logs confirm CI installs/builds omnidreams-interactive-drive, but neither CPU nor GPU pytest lists any tests from integrations/omnidreams/samples/interactive-drive/tests, and there is no scripts/check.sh invocation. This lines up with the repo config excluding the sample from root checks: pyproject.toml (line 135), pyproject.toml (line 162), .pre-commit-config.yaml (line 13). Evidence: CPU runs root pytest -m ci_cpu at cpu-logs.txt (line 1860); GPU runs root pytest -m ci_gpu at gpu-logs.txt (line 1513). If the sample is intended to be supported, CI should invoke the sample’s own check/test command explicitly.
* --cuda-visible-devices=auto has a developer-machine default. On any multi-GPU machine with no existing env var, it sets CUDA_VISIBLE_DEVICES=1: demo.py (line 520), demo.py (line 785). That assumes the “RTX6000 + GB300” layout and can silently pick the wrong GPU elsewhere.
* Stale package/extra commands remain. The sample only defines the dev extra: pyproject.toml (line 35). But docs/scripts still reference --extra ui --extra world-model: scripts/check.sh (line 28), integrations/omnidreams/README.md (line 45). There is also a bad suggested package name, --package interactive-drive, in demo.py (line 593).
* Prompt variant parsing is still inconsistent in helper paths. scene_loader now enforces prompt_<N>.txt, but the HUD/helper code still accepts bare prompt1.txt, and test_scene_helpers still tests that older form: scene_loader.py (line 85), demo.py (line 912), scene_bundle.py (line 15), test_scene_helpers.py (line 17).

wlewNV:
**1. Sample tests not running in CPU/GPU CI** — Solved without retrofitting markers across 20 test modules: added a `pytest_collection_modifyitems` hook in [`tests/conftest.py`](integrations/omnidreams/samples/interactive-drive/tests/conftest.py) that auto-tags each sample test from its existing sample-local marker (`xvfb` → `manual`, `gpu` → `ci_gpu`, otherwise → `ci_cpu`; xvfb wins over gpu so the smoke test stays opt-in since the GPU CI image isn't guaranteed to have Xvfb). Registered the workspace markers in the sample's pyproject so `strict_markers` doesn't reject the auto-assignments. Removed the root `norecursedirs` entry. Result:

| Tier | Count | Tests |
|---|---|---|
| `ci_cpu` | 70 | the pure unit tests |
| `ci_gpu` | 2 | `test_raster_reference_image[real]`, `[synthetic]` |
| `manual` | 2 | the two `test_app_smoke` Xvfb smoke tests |

Root CI `pytest -m ci_cpu` and `-m ci_gpu` now pick them up automatically. `scripts/check.sh` is unchanged (still runs the sample-local pytest with the `gpu`/`xvfb` markers).

**2. `--cuda-visible-devices=auto` dev-box default** — `auto` is now a no-op (leaves whatever the user already exported alone). Specific values still pass through verbatim, empty string still force-unsets. Dropped the `nvidia-smi` shell-out plus the now-unused `shutil` / `subprocess` imports.

**3. Stale `--extra ui --extra world-model` and `--package interactive-drive` refs** — Fixed all three I missed when folding the extras and renaming the package: `scripts/check.sh` (now just `--extra dev`), `integrations/omnidreams/README.md` (no extras), and the `_maybe_autostage_scene` SystemExit message in `demo.py`.

**4. Prompt variant parsing still inconsistent in helpers** — Right, only `scene_loader._discover_prompts` got the fix last round. `demo._variant_from_stem` still had the `return suffix or "default"` fallback that quietly accepted bare `prompt1.txt`, and `assets/scene_bundle.py` was an entirely separate copy of the original buggy logic for the unpacked-directory case. Extracted the underscore-required parser into [`scene_loader.variant_from_stem`](integrations/omnidreams/samples/interactive-drive/src/interactive_drive/scene_loader.py) and reused it from all three discovery paths:

69 sample-local CPU tests still pass, `ty` clean, lint clean.

jarcherNV:
I think just 1 more issue:

* Some stale user-facing install/GPU guidance remains. The code now makes --cuda-visible-devices=auto a no-op, but the sample README still says it sets CUDA_VISIBLE_DEVICES=1 on the RTX6000+GB300 dev box: README.md (line 237). Also, runtime error messages still refer to removed extras: presenter.py (line 20), slangpy_hud_presenter.py (line 261), prepare.py (line 141). There are a few stale comments too, but those are less important. The fix is just to align these strings with uv sync --package omnidreams-interactive-drive / uv run --package omnidreams-interactive-drive.

integrations/omnidreams/samples/interactive-drive/.pre-commit-config.yaml
liruilong940607, line 1: Curious whats the reason we need this dedicated pre-commit-config and can not use the repo level pre-commit-config?

wlewNV:
Fixed in [e405a69](https://github.com/NVIDIA/flashdreams/commit/e405a69).

**Stale strings:**

- README `--cuda-visible-devices` paragraph rewritten to match the no-op `auto` behaviour (was still describing the removed RTX6000+GB300 heuristic).
- `presenter.py:20` and `slangpy_hud_presenter.py:261` SlangPy-missing `RuntimeError` messages: `uv sync --extra ui` → `uv sync --package omnidreams-interactive-drive`.
- Four `prepare.py` `RuntimeError` messages: `uv sync --extra world-model` → same.
- `.pre-commit-config.yaml` ty hook: `--extra dev --extra ui --extra world-model` → `--extra dev`.

**CI failure:** root cause was `from conftest import SAMPLE_SCENE` and `from pipeline_fakes import ...` in the sample's test modules. Both only work under pytest's legacy `prepend` import mode (which puts every test dir on `sys.path`). Root CI runs `--import-mode=importlib` via the `marker_enforcement` plugin, and bare `import conftest` under importlib resolves to the *closest* conftest on `sys.path` — for these tests, the sibling `integrations/omnidreams/ludus-renderer/tests/conftest.py`. Five test modules failed collection.

integrations/omnidreams/samples/interactive-drive/pyproject.toml
liruilong940607, line 1: This makes `integrations/omnidreams/samples/interactive-drive/` a package. and `integrations/omnidreams` itself is already a package. So now we have a package inside a package -- a bit unusal structure to me. Would it causal any problems?

jarcherNV (approved):
LGTM

liruilong940607 (commented):
This is a great starts to unify the code paths thanks Williams!

One suggestion and one high level question:
- Does it make sense to put this desktop demo entrypoint besides `integrations/omnidreams/omnidreams/webrtc` with probably name like `integrations/omnidreams/omnidreams/local_app`?

- Do we want to make it more tightly integrated? Right now it is still basically another standalone code path. Many logics such as clipgt parsing should share the logics with  `integrations/omnidreams/omnidreams/webrtc`. 

Both of the above points don't necessarily blockers for this PR. Just want to align on the endpoint.

integrations/omnidreams/samples/interactive-drive/.pre-commit-config.yaml
wlewNV: Will be addressing in a follow-up PR. Tracking in #184

PR #185, opened 2026-05-27
Lower Python requirement from >=3.12 to >=3.10

janickm:
## Summary

Lower the minimum Python version from `>=3.12,<3.13` to `>=3.10` (no upper bound) for the `flashdreams` core package and all integration packages.

## Motivation

The codebase uses no Python 3.12-exclusive syntax (no PEP 695 type aliases, no type parameter syntax, no `@override` from `typing`). The actual language floor is Python 3.10 due to:
- `X | Y` union annotations evaluated at runtime
- `dataclass(kw_only=True)` and `dataclass(slots=True)`
- `TypeAlias` from `typing`

All `typing_extensions` backports (`TypeVar` with `default=`, `Self`) are already imported from `typing_extensions`, not `typing`.

No runtime dependency requires Python 3.12+ either (torch 2.11, triton 3.6, transformer-engine 2.14 all support 3.10+).

## Changes

- `flashdreams/pyproject.toml`: `requires-python = ">=3.10"`
- All 8 integration packages: `requires-python = ">=3.10"`
- All test harness packages: `requires-python = ">=3.10"`
- Root `pyproject.toml`: `[tool.ty.environment] python-version = "3.10"` so the type checker validates against the minimum supported version
- `.pre-commit-config.yaml`: `default_language_version: python: python3` (unversioned, uses whatever interpreter is available)
- 6 integration test files: `import tomllib` -> `import tomli as tomllib` (`tomllib` is 3.11+ stdlib)
- `flashdreams/pyproject.toml`: add `tomli>=2.0` to dev dependencies
- Remove unused `ty:ignore` comment in `ftheta.py` surfaced by the version change
- `CONTRIBUTING.md`: reference Python 3.10+
- Version bump to `0.1.0a3`
- Regenerated `uv.lock`

## Verification

- `ty check` -- all checks passed
- `pre-commit run -a` -- all hooks passed
- `pytest -m ci_cpu` -- 261 passed (4 pre-existing env-specific failures unrelated to this change)


jatentaki (changes requested):
We should probably adjust CI to use `py310` instead of `py312` ([currently in use](https://github.com/NVIDIA/flashdreams/actions/runs/26504200826/job/78052196486?pr=185#step:10:16)) to properly ensure this actually works as intended?

janickm:
## Smoke Test Results

All integration CLI commands verified after the Python version change:

### `flashdreams-run --help`
All 33 registered runners discovered and listed (wan21, self-forcing, causal-forcing, lingbot, omnidreams, cosmos2, flashvsr, fastvideo-causal-wan22, template).

### Individual runner `--help`
Tested: `wan21-t2v-1.3b-480p`, `self-forcing-wan2.1-t2v-1.3b-flash`, `causal-forcing-wan2.1-t2v-1.3b-framewise`, `lingbot-world-fast` -- all expose their full CLI flag tree via tyro correctly.

### `--no-instantiate` (config resolution without model loading)
All resolved cleanly:
- `template-offline`
- `self-forcing-wan2.1-t2v-1.3b-flash`
- `wan21-t2v-1.3b-480p`
- `lingbot-world-fast`
- `cosmos2-t2v-2b-720p`
- `flashvsr-v1.1-sparse-ratio-2.0`
- `omnidreams-sv-2steps-chunk2-loc6-lightvae-lighttae-perf`

### `template-offline --device cpu` (end-to-end inference attempt)
Models instantiated successfully. Pipeline ran until hitting the pre-existing CUDA-only RoPE Triton kernel (`apply_rotary_pos_emb requires a CUDA tensor`) -- this is an existing limitation unrelated to the Python version change.

### Summary
No regressions from lowering to Python 3.10. Plugin discovery, config dataclass resolution, and CLI argument parsing all work identically.

janickm:
> We should probably adjust CI to use `py310` instead of `py312` ([currently in use]

good point, checking - it could be that this is the system interpreter in the docker image (`/usr/bin/python3`, which happens to be 3.12), but let me have a deeper look that we make sure to use py310 wherever applicable (in particular when running the unit tests + ty etc.)

janickm:
Addressed in aff82e7c: added `UV_PYTHON="3.10"` to both the `cpu` and `gpu` job `env:` blocks in `ci.yml`. This makes uv download a standalone Python 3.10 interpreter and use it for venv creation, dependency resolution, and test execution -- so CI now validates against the declared minimum version rather than defaulting to the system Python 3.12 from Ubuntu 24.04.

janickm:
> Addressed in [aff82e7](https://github.com/NVIDIA/flashdreams/commit/aff82e7c98860a648924c2ae840171f5bb1e37cd): added `UV_PYTHON="3.10"` to both the `cpu` and `gpu` job `env:` blocks in `ci.yml`. This makes uv download a standalone Python 3.10 interpreter and use it for venv creation, dependency resolution, and test execution -- so CI now validates against the declared minimum version rather than defaulting to the system Python 3.12 from Ubuntu 24.04.

alternatively, we could introduce a `.python-version` file to the repo, but that would "force" everyone to use this version locally also

janickm:
[this is where it's used when uv creates the venv

https://github.com/NVIDIA/flashdreams/actions/runs/26508772290/job/78068011042?pr=185#step:9:19

btw: why are there no proper threads to discuss some topic :D]

janickm:
[checking the GPU build issues]

integrations/omnidreams/omnidreams/conditioning/world_scenario/ftheta.py
```diff
@@ -480,7 +480,7 @@ def rescale(
 
         elif self.reference_poly == "fw":
             fw_poly_coef = self._fw_poly.coef.copy()
-            fw_poly_coef /= 1 / avg_ratio
+            fw_poly_coef /= 1 / avg_ratio  # ty:ignore[unsupported-operator]
```
jatentaki, line 483: This is such a weird operation, wouldn't it make more sense to just `fw_poly_coef *= avg_ratio`? I know it's not part of your MR but maybe it's better to fix than just `ty:ignore`

janickm: yeah agree this would be better, let me add this as a separate PR [this only popped up as I accidentally updated `ty` that flagged this - but a `ty` update is not necessary in this PR for the python3.10 stuff alone - I'll do this separately]

liruilong940607 (approved):
This is great!

janickm:
@wilsonCernWq FYI - this uses a fixed upstream TE version now that addressed the build issues (no need to override the sources anymore..)

jatentaki (approved):

integrations/omnidreams/omnidreams/conditioning/world_scenario/ftheta.py
janickm: added in #193

PR #192, opened 2026-05-28
Add cuDNN SlangPy worker primer workaround

gtong-nv:


gtong-nv:
This WAR is no longer needed with the slanpgy upgrade, which fixes the blindly import of `VK_NVX_binary_import` extension

PR #205, opened 2026-05-28
Wire skip_finalize_kv_cache to interactive demo

gtong-nv:
This change adds `skip_finalize_kv_cache` to the demo's config file.
By default it's False so it's not changing the current behavior.


gtong-nv:
/ok to test [194b626](https://github.com/NVIDIA/flashdreams/pull/205/commits/194b626f726a6c75b9ca9c1952713f60457610b6)

csyonghe (approved):

PR #214, opened 2026-05-29
ci: add minimal-version resolution matrix to cpu and gpu jobs

janickm:
## Summary

Add a `resolution: [pinned, minimal]` matrix dimension to both CI jobs, validating that declared minimum dependency bounds actually work.

## How it works

| Matrix cell | Resolution strategy | What it validates |
|-------------|--------------------|--------------------|
| `pinned` | `uv sync` (lock file) | Current behavior -- tests against exact locked versions |
| `minimal` | `uv sync --resolution lowest-direct` | Declared minimum bounds are installable and pass tests |

The `minimal` cell re-resolves the workspace from scratch using the lowest allowed version for each direct dependency (e.g. `torch>=2.9` resolves to `torch==2.9.0` instead of `2.12.0`). This catches cases where code uses APIs not available at the declared minimum.

## Details

- Both cells run linting (pre-commit) and unit tests
- Wheel build only runs on `pinned` (we publish against locked versions)
- `publish-testpypi` requires all matrix cells (`needs: [cpu, gpu]`) -- both pinned and minimal must pass before publishing
- `fail-fast: false` -- both cells run to completion even if one fails
- Cache keys are separated: `cpu-pinned` / `cpu-minimal`, `gpu-sm{arch}-pinned` / `gpu-sm{arch}-minimal`

## Prerequisites

Depends on #211 which pins lower bounds on all workspace dependencies (enabling `--resolution lowest-direct` to resolve without failures).

## Local validation

Tested locally with torch 2.9.0 (lowest-direct resolution):
- CPU tests: 50 passed
- GPU tests: 21 passed (RTX A6000)


jarcherNV:
This still seems useful, but it needs a refresh before review/merge.

The `gpu (minimal)` job failed because the minimal environment built `block-sparse-attn` against a CUDA 13.2 container while PyTorch was compiled with CUDA 12.8.

There is also a design issue to resolve: the lock file appears to have been regenerated with `resolution-mode = "lowest-direct"`, which means the pinned and minimal cells no longer test different dependency sets. If the intent is to keep `uv.lock` as the normal/latest-compatible pinned environment, it should be regenerated without sticky lowest-direct mode and only the minimal CI cell should use `--resolution lowest-direct`.

Could you confirm whether you still want to pursue this? If yes, I think the next step is to rebase, fix the lock/resolution behavior, resolve the CUDA mismatch in the minimal GPU job, and rerun the full matrix.

PR #216, opened 2026-05-29
fix(ludus-renderer): scope JIT cache by GPU arch to prevent shared-FS poisoning

wlewNV:
## Summary

Renames the Ludus JIT cache slot from `ludus_renderer_plugin` to `ludus_renderer_plugin_sm_<cap>` (e.g. `_sm_100` on Blackwell GB200/GB300) and pins `TORCH_CUDA_ARCH_LIST` to the live device's compute capability instead of clearing it. Targets the ~4x cudaludus slowdown @Junchen / @Ruilong reported on the HSG GB200/GB300 Slurm cluster, while the same code runs fast on a locally-installed GB300.

## Root cause

`torch.utils.cpp_extension.load(...)` keys its build cache off a hash of the **source files** but NOT off the GPU arch / compile flags. On any shared-filesystem host (NFS `$HOME` on Slurm) a single user's `~/.cache/torch_extensions/` can hold a `.so` built on one node (CPU login, Hopper dev box, ...) and silently reuse it on a later job that lands on a different GPU generation. The mismatched binary lacks the right SASS, kernels fall back to PTX-JIT at every launch, and cudaludus throughput collapses by ~4x — exactly the omnidreams-webrtc demo's "20 fps on cluster, 30 fps locally" pattern. Local installs work because the cache is necessarily built fresh on the running GPU.

## Changes (`ludus_renderer/_ops/_plugin.py`)

1. **Pin `TORCH_CUDA_ARCH_LIST`** to the device's compute capability (e.g. `"10.0"` on GB200/GB300) instead of clearing it. Deterministic gencode; no dependency on `torch.cuda.get_arch_list()` (which varies per PyTorch wheel and can silently lack SM 10.0+ entries).
2. **Arch-tag the cache slot name** so a Hopper `.so` can't be reused on Blackwell, and vice versa. One-time ~30–60 s rebuild on first load after this lands.
3. **Richer startup log line** records `device_capability`, `arch_tag`, the value of `TORCH_CUDA_ARCH_LIST` actually used, and cache hit/miss.
4. **Warn (don't delete) the legacy cache slot** if it still exists, so operators can clean it up.

Fallback: no CUDA at import → `arch_tag="nocuda"`; capability detect fails → `arch_tag="auto"`. Both behave exactly as before the patch.

Out of scope: the same `TORCH_CUDA_ARCH_LIST=""` pattern in `ludus_renderer/nvjpeg.py` line 71 — separate extension, separate cache, follow-up.

### DIFF
--- a/integrations/omnidreams/ludus-renderer/ludus_renderer/_ops/_plugin.py
+++ b/integrations/omnidreams/ludus-renderer/ludus_renderer/_ops/_plugin.py
@@ -103,25 +103,86 @@ def _get_plugin():
         "../_cpp/bindings/torch_rasterize_cuda.cpp",
     ]
 
-    # Reset CUDA arch list to let PyTorch detect the installed GPU
-    os.environ["TORCH_CUDA_ARCH_LIST"] = ""
+    # Pin TORCH_CUDA_ARCH_LIST to the live device's compute capability and
+    # scope the JIT cache slot by that arch so a ``.so`` built for one SM
+    # can never be reused on another. ``torch.utils.cpp_extension.load``
+    # keys its cache off a hash of the SOURCE FILES but NOT the arch /
+    # compile flags, so on shared-filesystem hosts (NFS ``$HOME`` on
+    # Slurm/HPC clusters is the common case) a single user's
+    # ``~/.cache/torch_extensions/`` can hold a ``.so`` built on one node
+    # (e.g. a CPU-only login or a Hopper dev box) and then silently get
+    # reused on a node with a completely different GPU (Blackwell GB200 /
+    # GB300). The mismatched binary doesn't carry the right SASS, kernels
+    # fall back to PTX-JIT (or degraded paths) at every launch, and
+    # steady-state cudaludus throughput collapses by ~4x -- the exact
+    # symptom Junchen / Ruilong reported on the HSG GB200/GB300 cluster
+    # while the same code is fast on a locally-installed GB300 (where the
+    # cache was built fresh on the running GPU).
+    device_capability: tuple[int, int] | None = None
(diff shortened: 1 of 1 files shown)


JunchenLiu77:
Profiling result with this PR applied:
gen_ms:                        avg 402.9 ms / 8 frames  (~19.9 FPS)
pipeline_total_ms:             avg 179.0 ms / 8 frames
wrapper_render_condition_ms:   avg 219.4 ms
renderer_ctx_render_ms:        avg 218.3 ms
ctx_render_plugin_cuda_ms_sum: avg 214.7 ms
enqueue_ms:                    avg 26.5 ms

integrations/omnidreams/ludus-renderer/ludus_renderer/_ops/_plugin.py
```diff
@@ -103,26 +103,87 @@ def get_sort_key(x):
...
+        # source of arch-mismatch slowdowns prior to this patch.
+        legacy_dir = os.path.join(
+            os.path.dirname(build_directory),
+            "ludus_renderer_plugin",
+        )
+        if os.path.isdir(legacy_dir):
+            logger.warning(
+                "Found legacy Ludus extension cache at {}; this directory "
+                "was produced by a pre-arch-pinning build and may contain "
+                "a .so compiled for a different GPU. Safe to delete -- "
```
jatentaki, line 182: I wouldn't say it's safe to delete precisely because another operator may be depending on it :) I'd remove this legacy detection code branch for simplicity and assume the little bit of extra files won't hurt anyone and eventually get cleaned up when somebody clears the cache for whatever reason (disk cleanup, etc).

Issue #203, opened 2026-05-28
HY-WorldPlay WAN-5B I2V — follow-ups from PR #155


wenqingw-nv:
Tracks deferred work from #155 (merged).

## Code follow-ups — DONE (awaiting merge approval)

All CPU+GPU verified on RTX 6000 Ada; `cpu`/`gpu`/`docs`/OSRB/REUSE all pass on each.

- [x] **VAE `.pth`** — PR #223. Native `.pth` + 4-rule transform vs diffusers + ~50-rule → **bit-identical weights** (196/196 fp32, max |Δ|=0). Production default **flipped** to the `.pth`; diffusers kept as opt-in fallback.
- [x] **DiT native path + 404 fix** — PR #224. Native vs diffusers → **bit-identical** (825/825 fp32, max |Δ|=0). Remap is *load-bearing* (HY distilled ckpt is diffusers-keyed) so kept; native path is a proven-equivalent option. Also fixes the diffusers-DiT single-file **404** (→ sharded index).
- [x] **Pose JSON default** — PR #222. `_pose.py` `==`→`>=` prefix-slice + `--example-data` auto-pose. Verified by a real `--example-data` rollout → valid mp4.
- [x] **Base-pipeline load fix** — PR #227 (spun off from verification). Running without `--ckpt-path` crashed (base Wan ckpt lacks zero-init HY keys + strict `load_state_dict`). Fixed to tolerate exactly the HY zero-init keys; verified end-to-end (base rollout → valid mp4).

## Perf / docs — DONE → **PR #231** (awaiting merge approval)

Done on a GPU with larger memory (branch `wenqing/hy-worldplay-perf-handoff`).

- [x] **Re-bench at `num_chunk=8`, `warmup_chunks=5`**, DiT + VAE enc/dec scope, both legs cuDNN SDPA + torch.compile. Native **~1.91× on DiT** (632 vs 1206 ms), **parity on VAE decode** (383 vs 372 ms) → **~1.55× DiT+VAE per chunk**; input-independent across 5 `data_local/` inputs. Corrects the earlier "6×" wall-clock figure (apples-to-oranges) and the VAE-wins expectation — the matched, same-kernel, steady-state win is on **DiT**.
- [x] **Curated sample MP4s** — native rollouts for all `data_local/*`; hero + 4-clip gallery on the model card.
[...]

PR #155, opened 2026-05-24
feat(hy_worldplay): HY-WorldPlay WAN-5B I2V integration

wenqingw-nv:
## Summary

Adds the HY-WorldPlay WAN-5B I2V integration as a new `integrations/hy_worldplay/` plugin — Tencent Hunyuan's real-time interactive world model (streaming video diffusion with action + camera-trajectory conditioning and reconstituted-context memory). The native path is the production default; the vendor wrapper stays available via `--no-use-native-pipeline` for byte-for-byte match against upstream's `torchrun wan/generate.py`.

Phase breakdown:

- **Phase 1 — vendor wrapper.** Plugin packaging as a `uv` workspace member; `HyWorldPlayWanI2VRunner` shim over upstream's `wan/generate.py` `WanRunner.predict()` for bit-identical output to `torchrun wan/generate.py`; registered with `flashdreams-run` via the `flashdreams.runner_configs` entry-point group; parity-check harness under `tests/parity_check/`.
- **Phase 2a — Wan 2.2 TI2V-5B recipe** (in `flashdreams.recipes.wan`). Fills the gap between Wan 2.1 (1.3B / 14B) and Wan 2.2 14B: `Wan22TI2V5BVAE{Encoder,Decoder}Config` (16× spatial, 48-ch latent, residual + outer-patchify=2), `WanDiTNetworkTI2V5BConfig` (3072d / 30-layer, no CLIP cross-attn), `Wan21TransformerConfig.ti2v_first_frame_per_token_timestep` flag (AR-0 per-token / AR≥1 scalar dispatch), `PIPELINE_WAN22_TI2V_5B` pre-rolled config, and diffusers safetensors remaps. Useful independent of HY-WorldPlay.
- **Phase 2b — native HY-WorldPlay integration.** Each conditioner gated behind its own flag and zero-initialised so flipping flags on without the distilled checkpoint is a strict identity:
  - **2b.1 / 2b.2.** Native runner over `PIPELINE_WAN22_TI2V_5B` + distilled 4-step Euler schedule (`FlowMatchEulerDiscreteSchedulerConfig`).
  - **2b.3.** 81-class action conditioner (AdaLN add on `HyWorldPlayWanDiTNetwork`).
  - **2b.4.** PRoPE dual-branch self-attention (`HyWorldPlayPRoPEBlock`; `prope_qkv` math in `flashdreams.core.attention.prope`).
[...]

PR #222, opened 2026-05-29
feat(hy_worldplay): pose-JSON default for --example-data

wenqingw-nv:

Makes the pose parser tolerant of longer pose sources and wires upstream's sample trajectory into `--example-data`.

- `_pose.py`: relax the strict `len(poses) == n_latents` check to `>=` and prefix-slice to the rollout's `num_chunk * 4` budget (vendor's behaviour). Under-supply still raises with the same message.
- `runner.py`: `--example-data` now also lazy-downloads `assets/pose/test_forward_32_latents.json` (33 entries) into `data_local/hy_worldplay/` and uses it as the default pose when `--pose` is unset. The 33-entry file drives any `num_chunk <= 8` run via the prefix slice.
- Tests: `test_pose_data_longer_source_is_prefix_sliced` + renamed the length-mismatch test to reflect that only under-supply raises. README updated.

CPU-verified (parser tests). No production default changes — additive/opt-in via `--example-data`.

Part of #203.

### DIFF
--- a/integrations/hy_worldplay/README.md
+++ b/integrations/hy_worldplay/README.md
@@ -81,13 +81,17 @@ uv run flashdreams-run hy-worldplay-wan-i2v-5b \
     --example-data \
     --ckpt-path /path/to/models/wan_distilled_model/model.pt \
     --num-chunk 1 \
-    --pose "w-3" \
     --output-dir outputs
 ```
 
-`--example-data` lazy-downloads upstream's `assets/img/test.png`
-into `data_local/hy_worldplay/` (gitignored) and uses it as the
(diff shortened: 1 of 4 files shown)


wenqingw-nv:
Verified end-to-end on RTX 6000 Ada: `flashdreams-run hy-worldplay-wan-i2v-5b --example-data True --num-chunk 1 --ckpt-path <distilled>` downloaded the live pose JSON + image, the prefix-slice parser drove the distilled pipeline, and it wrote a valid 13-frame 704×1280 MP4. (Base-pipeline-without-ckpt path is blocked by two unrelated pre-existing bugs — see #203.)

liruilong940607 (approved):

PR #223, opened 2026-05-29
feat(wan-vae): native .pth state-dict transform for Wan 2.2 TI2V-5B VAE

wenqingw-nv:

#155 review asked whether loading upstream's single-file `Wan2.2_VAE.pth` lets us drop the diffusers `state_dict_transform`.

**Finding:** it can't be *dropped*, but it shrinks a lot. The `.pth` uses Wan's native module layout (each down/up stage is a flat `Sequential` of residual blocks + a resample) while our `WanVAE` groups them (`resnets.{j}` + `downsampler`/`upsampler`). So it still needs a remap — **4 rules vs the ~50-rule diffusers remap**. The earlier "missing params / meta tensors" was a misdiagnosis: all params are present under different key names; without a transform `load_state_dict(assign=True)` simply left our slots on meta.

- Add `wan22_ti2v_5b_vae_pth_state_dict_transform` (4-rule remap). Verified **196↔196 key/shape bijection** with the WanVAE module tree (no meta leftovers, no unexpected keys).
- CPU tests: `test_wan22_vae_pth_remap_is_full_bijection` (builds the TI2V-5B WanVAE on meta, proves full coverage) + `test_wan22_vae_pth_remap_spot_checks_real_keys` (guards the regex against real `.pth` key strings).
- Correct two stale comments that claimed the `.pth` matched our layout directly.

**Opt-in:** diffusers stays the production default; a follow-up flips the default + drops the ~50-rule remap once a GPU decode-parity smoke confirms identical output.

Part of #203.

### DIFF
--- a/flashdreams/flashdreams/recipes/wan/autoencoder/vae.py
+++ b/flashdreams/flashdreams/recipes/wan/autoencoder/vae.py
@@ -68,14 +68,17 @@ _PUBLIC_WAN_VAE_CHECKPOINT_PATHS = {
 # / ``quant_conv`` / ``post_quant_conv`` etc. -> our internal layout).
 WAN22_TI2V_5B_VAE_DIFFUSERS_PATH = "https://huggingface.co/Wan-AI/Wan2.2-TI2V-5B-Diffusers/resolve/main/vae/diffusion_pytorch_model.safetensors"
 
-# Alternative upstream single-file checkpoint. Top-level key prefixes
-# (``encoder.*`` / ``decoder.*`` / ``conv1`` / ``conv2``) line up with
-# our internal :mod:`_residual_vae` model shape, BUT the ``.pth`` does
-# not cover every parameter our ``WanVAE`` wrapper builds (some encoder
-# / decoder slots stay on meta and ``model.to(device)`` raises
-# ``NotImplementedError: Cannot copy out of meta tensor``). Kept as a
-# constant for callers who want to invest in a tighter audit + a
-# tailored loader; the diffusers path above is still the default.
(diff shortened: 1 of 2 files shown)


flashdreams/flashdreams/recipes/wan/autoencoder/vae.py
```diff
@@ -1632,25 +1636,88 @@ def wan22_ti2v_5b_vae_state_dict_transform(
...
+    # the stage's ``downsampler`` / ``upsampler``; its sub-block index is
+    # dropped since there is exactly one.
+    r"^encoder\.downsamples\.(\d+)\.downsamples\.\d+\.(resample|time_conv)\.(.*)$": (
+        r"encoder.downsamples.\1.downsampler.\2.\3"
+    ),
+    r"^decoder\.upsamples\.(\d+)\.upsamples\.\d+\.(resample|time_conv)\.(.*)$": (
+        r"decoder.upsamples.\1.upsampler.\2.\3"
+    ),
+}
+
```
liruilong940607, line 1670: I think we just need to name the component in the same way it is named in pth they it can load pth?

wenqingw-nv: Done — the production VAE config now loads the native Wan2.2_VAE.pth directly, whose keys already match our WanVAE layout, so no remap runs.

liruilong940607: Weirdly I don't see any changes fixes this. is that you forget to push the change?

PR #224, opened 2026-05-29
feat(wan22): native DiT checkpoint path — drops the diffusers DiT remap

wenqingw-nv:

Same question as the VAE follow-up, for the DiT remap (~25 rules).

**Finding:** the remap can be dropped *entirely*. Upstream's native `Wan-AI/Wan2.2-TI2V-5B` DiT checkpoint (sharded safetensors + index) uses byte-for-byte the `WanDiTNetwork` key names (`blocks.N.self_attn.q`, `text_embedding.0`, `head.head` ...), because our network was ported from native Wan. Verified an **825↔825 name-identical bijection**, so it loads with `state_dict_transform=None`. `load_checkpoint` already handles the `.safetensors.index.json` shard format.

- Add `WAN22_TI2V_5B_DIT_NATIVE_PATH` (+ docstring documenting the zero-remap finding); note the native alternative on the diffusers remap dict.
- `test_native_dit_checkpoint_needs_no_remap` (manual; fetches the ~250 KB index, no weights) guards the key-identity claim.

**Opt-in:** diffusers stays the production default; a follow-up flips the default + deletes `_WAN22_TI2V_5B_DIT_KEY_REMAP` once a GPU decode-parity smoke confirms identical output. (Native DiT is sharded fp32 ~20 GB vs the diffusers single file — worth confirming download/merge cost in that smoke.)

Part of #203.


integrations/wan22/wan22/config.py
```diff
@@ -55,12 +55,41 @@
...
 
-The 5B variant ships as a single safetensors file (no sharded index)
-under the ``transformer/`` subfolder of the ``Wan-AI`` diffusers repo."""
+The ``transformer/`` subfolder ships 5 shards + a
+``.safetensors.index.json``; there is **no** single-file
+``diffusion_pytorch_model.safetensors`` (that bare-filename URL 404s --
+it was the prior value of this constant, which broke any base-pipeline
+load). ``load_checkpoint`` resolves the index directly. Loads via
+:func:`wan22_ti2v_5b_dit_state_dict_transform` (the diffusers naming
+differs from ours). This is the production default."""
```
liruilong940607, line 68: Can you do a pass in a separate PR later on to clean up all those chain of thoughts comment / doc strings from the agent? Not only in this file but also many places in the HY/wan22 integration. 

Imagine you are a user reading the code and these doc strings, Many details here are not needed to be known for the user. Only be explainable enough for user to know how to use. The history of how we get to here doesn't need to be marked.

liruilong940607 (commented):
Do we have any test to verify the output is the same from two different way of loading the checkpoint?

wenqingw-nv:
> Do we have any test to verify the output is the same from two different way of loading the checkpoint?

Yes — test_native_dit_matches_diffusers_weights in integrations/wan22/tests/test_dit_remap.py loads both paths (native + identity remap, and diffusers + remap) and asserts max |Δ| == 0 across all 825 DiT tensors. Identical weights ⇒ identical output.

integrations/wan22/wan22/config.py
wenqingw-nv: Done in a separate PR: #314. It trims the agent's chain-of-thought and historical narration from comments/docstrings across the HY-WorldPlay + wan22 integration, leaving tight user-facing docs and one-line correctness notes.

PR #227, opened 2026-05-30
fix(hy_worldplay): load base Wan checkpoint into the HY network (no --ckpt-path path)

wenqingw-nv:
Spun off from #203 (surfaced while GPU-verifying the HY-WorldPlay follow-ups).

**Bug:** running the HY pipeline **without `--ckpt-path`** — the path the README documents as "loads base Wan 2.2, HY conditioners stay zero-init, strict identity / parity-safe vs base Wan" — actually raises:

```
RuntimeError: Error(s) in loading state_dict for HyWorldPlayWanDiTNetwork:
  Missing key(s): blocks.*.self_attn.o_prope.{weight,bias}, action_embedding.{0,2}.{weight,bias}
```

The base Wan checkpoint has no `action_embedding` / `o_prope` params, and `Wan21Transformer.__init__` does a strict `load_state_dict`. Latent because CI/GPU runs always pass `--ckpt-path` (distilled), and the smoke tests only assert the static config — never a real base load.

**Fix:** override `HyWorldPlayWanDiTNetwork.load_state_dict` to tolerate *only* the HY-specific zero-init keys (`action_embedding.*`, `*.o_prope.*`) when absent — they keep their zero-init values (the intended identity). Any **other** missing key, or any unexpected key, still raises (not a blanket `strict=False`). The distilled checkpoint carries these keys, so that path is unchanged.

**Verified:**
- New CPU tests: tolerates missing HY zero-init keys (and confirms they stay zero) / still rejects other-missing + unexpected.
- End-to-end on RTX 6000 Ada: `flashdreams-run hy-worldplay-wan-i2v-5b --example-data True --num-chunk 1 --pose w-3` (no `--ckpt-path`) → valid 13-frame 704×1280 mp4.

Part of #203.


liruilong940607:
Why would it load Wan2.2 base checkpoint? We only consider inference so it should directly load the HY ckpt

wenqingw-nv:
Fixed in #315: the static pipeline now defaults its transformer checkpoint to HY-WorldPlay's distilled WAN-5B weights (tencent/HY-WorldPlay, via hy_worldplay_distilled_state_dict_transform). --ckpt-path is now just a local override; the base-checkpoint identity path remains reachable through it. Superseding this PR with #315 — closing in its favor.

> Why would it load Wan2.2 base checkpoint? We only consider inference so it should directly load the HY ckpt

PR #230, opened 2026-05-30
Support legacy prompt variant filenames

gtong-nv:
Some USDZ scene bundles use legacy numeric prompt filenames such as
`prompt1.txt`, `prompt2.txt`, and `prompt3.txt`, while the newer canonical
form is `prompt_1.txt`, `prompt_2.txt`, and `prompt_3.txt`.

Before this change, the interactive-drive prompt discovery path rejected the
legacy no-underscore numeric form. In those scenes, selecting variant `1`, `2`,
or `3` could silently fall back to the default prompt, making the first frame
and prompt variant selection disagree.

This change makes the shared variant parser accept numeric legacy names while
still rejecting ambiguous non-numeric suffixes such as `promptnight.txt`. It
also logs the actual prompt source selected at scene load, so prompt fallback is
visible during demo bring-up.

### DIFF
--- a/integrations/omnidreams/omnidreams/interactive_drive/demo.py
+++ b/integrations/omnidreams/omnidreams/interactive_drive/demo.py
@@ -1065,12 +1065,10 @@ def _make_thumbnail(image: Image.Image, size: tuple[int, int]) -> Image.Image:
 
 
 # ``_variant_from_stem`` removed in favour of the shared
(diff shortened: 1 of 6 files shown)


gtong-nv:
/ok to test [7fe4394](https://github.com/NVIDIA/flashdreams/pull/230/commits/7fe439444ce88510d439d6564923d643f7825f6a)

liruilong940607 (approved):

PR #233, opened 2026-05-30
demo: wire native fp8 vae opt to manifest

gtong-nv:
This change adds an toggle to the demo manifest to enable or disable fp8 VAE native implementation.
By default, it's not enabled.


gtong-nv:
/ok to test [194fab1](https://github.com/NVIDIA/flashdreams/pull/233/commits/194fab1157ad846b97b06bcd9655d0000f9fd65c)

wlewNV (approved):

PR #234, opened 2026-05-31
demo: make HDMap raster cuda-interop

gtong-nv:
## Summary

- Make the interactive-drive CUDA path the default.
- Use `INTERACTIVE_DRIVE_DISABLE_CUDA_INTEROP` as the single opt-out switch.
- Use that opt-out to disable both presenter CUDA interop and CUDA-backed HDMap raster frames.
- Update interactive-drive docs and add focused tests.

## Perf

Measured with:

```bash
xvfb-run -a ./run_demo.sh
INTERACTIVE_DRIVE_DISABLE_CUDA_INTEROP=1 xvfb-run -a ./run_demo.sh
```

At local `1168x640` perf config:

| Mode | Steady FPS | Steady next chunk |
| --- | ---: | ---: |
| default CUDA-backed path | ~20.7 avg | ~296 ms |
| disabled CUDA interop | ~18.0 avg | ~340 ms |

Default path is roughly 15% faster in this run.

### DIFF
--- a/integrations/omnidreams/omnidreams/interactive_drive/README.md
+++ b/integrations/omnidreams/omnidreams/interactive_drive/README.md
@@ -347,22 +347,22 @@ Then open `http://localhost:8080/`.
 For a richer browser frontend with lower latency, prefer the separate
 `omnidreams.webrtc.server` entry point.
 
-Both Vulkan presenters can use SlangPy CUDA interop for generated RGB frames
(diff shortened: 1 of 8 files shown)


gtong-nv:
/ok to test [e22520d](https://github.com/NVIDIA/flashdreams/pull/234/commits/e22520de79dbbe0dda76105a0b9197aa706dcb5a)

wlewNV (approved):

Issue #232, opened 2026-05-30
kill block sparse attn dependency and replace it with native triton impl


liruilong940607:
Goal: replace the block-sparse-attn dependency in flashvsr integration with our triton implementation

Reason: the block-sparse-attn is a cuda lib that would trigger long compiling when run uv sync. replace it with triton implmentation avoids the long compiling [triton do JIT compiling] so this could greatly improve user experience. long term we want to avoid cuda kernels and only use triton/cute/cutile in the codebase for customized kernel fusion

JunchenLiu77:
#247

PR #247, opened 2026-06-01
Remove dependency of external block-sparse attention dependency for FlashVSR 

JunchenLiu77:
This PR is target to fix Issue #232. I port the [agent-written triton kernel](https://gitlab-master.nvidia.com/ruilongl/block-sparse-attn/-/merge_requests/1/diffs#diff-content-7a858105917e3cde88433e15ecc79ec891326093) into flashdreams with some alignment and tuning.

### FlashVSR GB200 End-to-End Benchmark Result

I benchmarked the in-tree Triton sparse-attention backend against the external block-sparse baseline on a single GB200 GPU. The benchmark uses the FlashVSR 2x upsampling path with chunk size 8, follow the [existing profiling benchmark](https://verbose-adventure-7plnn5m.pages.github.io/main/models/flashvsr.html). Reported numbers are median steady per-chunk latency; compile/warmup times are ignored.

All tested resolutions completed successfully and wrote output videos plus per-AR-step stats.

| input resolution | external ms | triton ms | triton delta | speedup |
| --- | ---: | ---: | ---: | ---: |
| 384x384 | 83.8 | 74.7 | -10.9% | 1.12x |
| 672x384 | 120.5 | 114.5 | -5.0% | 1.05x |
| 384x672 | 117.0 | 111.2 | -5.0% | 1.05x |
| 640x480 | 138.8 | 132.5 | -4.5% | 1.05x |
| 768x416 | 140.2 | 135.5 | -3.3% | 1.03x |
| 1280x704 | 365.3 | 372.5 | +2.0% | 0.98x |

The Triton backend matches external-backend efficiency across the sweep: it is faster on 5/6 tested resolutions and within 2% on the largest 1280x704 case. Peak memory is also comparable, with no regressions at the largest resolution.

In addition to the end-to-end benchmark, the FlashVSR parity and dispatch tests pass with the in-tree Triton backend. The tests cover sparse-attention backend selection, the FlashVSR-specific Triton dispatch configuration, and the opaque inference path used to keep sparse attention from being expanded by Inductor during compiled inference.


integrations/flashvsr/tests/parity_check/changes.patch
```diff
@@ -378,3 +378,24 @@ diff --git a/setup.py b/setup.py
...
+diff --git a/block_sparse_attn/block_sparse_attn_interface.py b/block_sparse_attn/block_sparse_attn_interface.py
+new file mode 100644
+--- /dev/null
++++ b/block_sparse_attn/block_sparse_attn_interface.py
+@@ -0,0 +1,5 @@
++"""Route upstream FlashVSR sparse-attention calls to flashdreams' Triton backend."""
++
++from flashvsr.transformer.triton_sparse_attn import block_sparse_attn_func
++
++__all__ = ["block_sparse_attn_func"]
```
liruilong940607: Are we using our triton impl for the baseline? I think for baseline let's just keep what it have

pyproject.toml
```diff
@@ -119,6 +89,8 @@ exclude = [
     # Bench / parity-debugging scaffolding for the hy_worldplay integration;
     # not shipped product, mirrors the interactive_drive precedent above.
     "integrations/hy_worldplay/tests/parity_check/**",
+    # Triton JIT kernels use a DSL that ty cannot model as normal Python.
+    "integrations/flashvsr/flashvsr/transformer/triton_sparse_attn.py",
```
liruilong940607: What does this mean?

We already have triton kernels here but doesn't seem to need something like this

https://github.com/NVIDIA/flashdreams/blob/main/flashdreams/flashdreams/core/attention/rope_kernel.py

JunchenLiu77:
resolve #232

CONTRIBUTING.md
```diff
@@ -320,8 +320,8 @@ the declared minimums. This means:
 
 ## Speeding up local builds
```
liruilong940607: I think it doesn't trigger TE anymore. And without block sparse attn I think we can get rid of this section entirely ？

docs/source/quickstart/installation.rst
liruilong940607: also here. Could you do a scan on things like this?

integrations/flashvsr/flashvsr/grpc/uplift_server.py
```diff
@@ -98,28 +97,10 @@ def _resolve_scale(scale: int) -> Scale:
...
-        return False
-    return True
-
-
 def _resolve_attention_mode(attention_mode: RequestedAttentionMode) -> AttentionMode:
     if attention_mode == "full":
         return "full"
-    if _block_sparse_attn_available():
-        return "sparse"
-    if attention_mode == "auto":
```
liruilong940607, line 114: This makes auto not useful anymore? But instruction has auto

JunchenLiu77:
/ok to test [457855d](https://github.com/NVIDIA/flashdreams/pull/247/commits/457855d630c740e08296d8734a54f836e8a7d562)

liruilong940607 (approved):
Great!

JunchenLiu77:
/ok to test [8f6cc32](https://github.com/NVIDIA/flashdreams/pull/247/commits/8f6cc32e1da623ba8b4261ee1810c08d03f09b7f)

PR #252, opened 2026-06-02
interactive-drive: materialize lazy RGB frame before JPEG encode

jialiangw219:
## Summary
- Fixes a crash on the first rendered frame of the `interactive-drive` MJPEG stream: `AttributeError: '_LazyRGBFrame' object has no attribute '__array_interface__'`.
- The streaming presenter's no-overlay path (`_with_status_overlay` with `message is None`) forwards the `_LazyRGBFrame` wrapper straight to `_publish`. That wrapper implements `__array__` but not `__array_interface__`, which `PIL.Image.fromarray` requires.
- Coerce with `np.asarray(...)` at the single publish chokepoint: a no-op for plain ndarrays, and it triggers `__array__` to materialize the lazy GPU-backed frame.

## Test plan
- [ ] Launch `interactive-drive --stream-mjpeg 8080` and confirm the stream renders frames (previously crashed at first present).
- [ ] Verify status-overlay frames (idle/loading) still render correctly.

### DIFF
--- a/integrations/omnidreams/omnidreams/interactive_drive/streaming_presenter.py
+++ b/integrations/omnidreams/omnidreams/interactive_drive/streaming_presenter.py
@@ -794,7 +794,10 @@ class MJPEGStreamingPresenter:
 
     def _publish(self, rgb_host_uint8: np.ndarray) -> None:
         buf = io.BytesIO()
-        Image.fromarray(rgb_host_uint8).save(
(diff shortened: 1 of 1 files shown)


wenqingw-nv:
Superseded by #251, which already contains the canonical `_as_rgb_host_uint8` lazy-frame fix applied at all three `Image.fromarray` boundaries (_publish, _publish_bev, and _with_status_overlay). Closing to avoid duplication.

PR #258, opened 2026-06-02
Cleanup for public release readiness by removing internal-only storage, launch, and release-process artifacts.

liruilong940607:
Resolves #168 

## Summary

  Cleanup for public release readiness by removing internal-only storage, launch, and release-process artifacts.

  ## What Changed

  - Removed internal storage toggle/code paths and references:
    - Deleted internal storage helper module and tests.
    - Removed internal storage config/runner wiring.
    - Removed stale `use_internal_storage` imports and comments.

  - Removed internal S3 helper scripts and risky examples:
    - Deleted asset upload/download shell scripts.
    - Removed demo/internal usage blocks from S3 utility modules.
    - Kept generic S3 support functions intact.

  - Removed Slurm and cluster-specific launch scripts:
    - Deleted `launch_collision_slurm.py`.
    - Deleted `tests/run_tests_slurm.sh`.
    - Removed Skippy/Slurm preprocessing launcher and task shim.
    - Updated test docs to only document local and Docker runners.

  - Removed internal build farm references:
    - Deleted `docker/docker_farm_setup.sh`.
    - Removed `dgx-spark`, internal contact, and NVIDIA infra-specific wording from Docker docs/build script
    comments.

  - Removed internal release-process documentation:
    - Deleted the OSRB/NVBugs-oriented agent skill.
    - Scrubbed OSRB/NVBugs/GitLab-internal wording from release lint/docs.

  - Scrubbed internal paths and labels from examples/docs:
    - Replaced `/lustre`, `/home/jseo`, `flashsim`, `team-sil`, `I4/SIL`, and similar internal references.
    - Updated comments to use public/generic wording.

  - Moved the standalone Ludus cache validator out of the removed `skippy/` namespace:
    - `scripts/skippy/validate_cache.py` -> `scripts/validate_cache.py`

PR #259, opened 2026-06-02
build: bump version to 0.1.0rc1 and deploy to PyPi

janickm:
For public release, move to publishing to public PyPi instead of Test PyPi, and update version from `0.1.0a4` to first release-candidate `0.1.0rc1`.

WARNING: this will deploy to public PyPi - let's only merge once we are ready to switch


.github/workflows/ci.yml
janickm, line 1: just a general comment that we should likely merge https://github.com/NVIDIA/flashdreams/pull/258 befor this one

liruilong940607 (approved):

PR #268, opened 2026-06-03
Fix raster cuda-interop on resize

gtong-nv:
This commit fixes raster CUDA interop after HUD window resize by recreating the
shared CUDA/Vulkan interop resources for the resized window instead of falling
back to the host path.


gtong-nv:
/ok to test [73a65ba](https://github.com/NVIDIA/flashdreams/pull/268/commits/73a65ba42e0364a0acbaf00b6c329ae9ff50cb0f)

wlewNV (approved):

PR #282, opened 2026-06-03
Package FlashVSR CUDA source in wheel builds

fallintoplace:
## Summary
- include `flashvsr/csrc/*.cu` as package data in FlashVSR wheel builds

## Why
`FlashVSRColorCorrector` loads `flashvsr/csrc/color_corrector_adain_cuda.cu` from the installed package at runtime. The source file is present in the repository, but wheel builds currently omit it, which breaks the default CUDA color-corrector path for non-editable installs.

This change keeps the runtime behavior the same and only fixes the built wheel contents.

## Validation
- `python3 -m pip wheel integrations/flashvsr --no-deps -w /tmp/flashdreams-wheelcheck`
- inspected the rebuilt wheel and confirmed `flashvsr/csrc/color_corrector_adain_cuda.cu` is included


gtong-nv:
/ok to test [05ed95a](https://github.com/NVIDIA/flashdreams/pull/282/commits/05ed95ae656c8b149c8990ea86c564d9c3cc0183)

jmccaffrey-nv:
Thanks for the focused packaging fix. I rebuilt the wheel from the merged tree and confirmed flashvsr/csrc/color_corrector_adain_cuda.cu is included, which matches the runtime path in corrector.py. I also checked this against the maintaining-oss-state workflow: no new deps, no vendored third-party source, reuse lint passes, and no OSRB/notice updates are triggered.  LGTM

jmccaffrey-nv (approved):

PR #285, opened 2026-06-04
Model Integration Agent Skill

wenqingw-nv:
Add an **agent integration skill** (e.g., HY) so contributors integrate models in their own repos, rather than growing flashdreams with more models.

### DIFF
--- a/agentic/skills/flashdreams-recipes/SKILL.md
+++ b/agentic/skills/flashdreams-recipes/SKILL.md
@@ -7,6 +7,8 @@ description: Navigate the flashdreams package layout and integration architectur
 
 A map of how `flashdreams/` is organized and how a single rollout flows through the framework. Read once before adding a integration under `flashdreams/flashdreams/integrations/` or restructuring an existing one. Keep docstrings consistent with the `python-docstring-style` skill.
 
+> For the **step-by-step procedure** of doing an integration end-to-end (scope → scaffold → recipe → checkpoint remap → conditioners → runner → verify → perf), see the **`integrate-a-model`** skill. This skill is the *map*; that one is the *route*.
+
 > **The fastest way to learn this codebase is to clone the structure of `integrations/template/`.** It is the reference integration — every contract this skill describes is wired up there in its minimal form. Skim it side-by-side with this document.
 
 ## TL;DR
--- a/agentic/skills/integrate-a-model/SKILL.md
+++ b/agentic/skills/integrate-a-model/SKILL.md
@@ -8,6 +8,8 @@ description: End-to-end workflow for porting an external video diffusion model i
 The ordered procedure for binding an external video model to the flashdreams
 framework. Read the **`flashdreams-integrations`** skill first for the architecture
 (layers, contracts, the cache tree) — this skill is the *route*, that one is the *map*.
+(That skill's `name:` is `flashdreams-integrations` but it lives in the directory
+`agentic/skills/flashdreams-recipes/` — mind the dir/name mismatch.)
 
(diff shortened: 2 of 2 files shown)


jmccaffrey-nv:
Feedback with Codex and GPT-5.5 :

Hi @wenqingw-nv , this is a strong skill. The remap/parity section captures hard-won knowledge that saves future contributors time , and the phase ordering is easy to follow.
My main request before merge would be to split the integration path into two lanes: in-tree workspace-member integrations and out-of-tree plugin packages. The current text says it supports downstream repos, but the scaffold mostly teaches the in-tree integrations/<name> shape. A small external pyproject.toml example with the flashdreams.runner_configs entry point would make this simpler for outside public users.

**Findings**
_[P1 nice-to-have]_ agentic/skills/integrate-a-model/SKILL.md (line 51) says this applies to “a downstream repo,” but Phase 1 only describes an in-tree integrations/<name>/ workspace member. For external repos, the key contract is an installable package exposing [project.entry-points."flashdreams.runner_configs"], or FLASHDREAMS_RUNNER_CONFIGS during development. The actual discovery seam is in plugins/registry.py (line 35), and hy_worldplay demonstrates the entry point in pyproject.toml (line 73). I’d add a short “external repo plugin” scaffold beside the in-tree scaffold.

_[P1 nice-to-have]_ SKILL.md (line 76) makes the HY-WorldPlay path sound like the universal pattern. The integrations directory has several lanes: wan22 is config-only, self_forcing/causal_forcing are light Wan variants, lingbot adds serving/runtime surfaces, and flashvsr uses real builder helpers for dynamic resolution variants. I’d add a Phase 0 decision: config-only recipe, flashdreams-run runner plugin, serving adapter, or full native pipeline port.

_[P2 should-have ]_ The checkpoint bijection snippet at SKILL.md (line 121) is close, but still not copy-paste runnable: my_state_dict_transform_keys is undefined, while real state_dict_transform functions operate on tensor dictionaries, not shape dictionaries. Since this is likely the highest-value part of the skill, I’d either include a tiny reusable key-only remap helper or label the snippet as pseudocode.

wenqingw-nv:
Thanks Jon — all three addressed in 389a8d1:

- **Two lanes (P1):** Phase 1 now splits **in-tree workspace member** vs **out-of-tree pip-installable plugin**, with a standalone `pyproject.toml` + `[project.entry-points."flashdreams.runner_configs"]` example and the `FLASHDREAMS_RUNNER_CONFIGS` dev path (discovery seam: `flashdreams/plugins/registry.py`).
- **Lane decision (P1):** Phase 0 adds a table — config-only (`wan22`) / runner plugin (`self_forcing`, `causal_forcing`, `hy_worldplay`) / serving adapter (`lingbot`) / full native port (`flashvsr`); calls out HY as the runner-plugin lane, not the universal pattern.
- **Bijection snippet (P2):** rewritten to feed the *real* `state_dict_transform` a `{name: meta-tensor}` dict (its actual signature, zero memory; shapes from safetensors headers) — dropped the undefined helper. Verified runnable end-to-end (825↔825).

jmccaffrey-nv:
Thanks Wenqing, re-reviewed 389a8d1. This closes my three comments. Approving.

One optional edit : qualify the “no build_* factories” line for the simpler lanes, since flashvsr is now called out as the builder-variant lane.

jmccaffrey-nv (approved):

wenqingw-nv:
Done in 2b38f6f — the 'no `build_*` factories' line now calls out the exception for the full-native-port lane (`flashvsr`).

PR #287, opened 2026-06-04
docs: document Lingbot-World WebRTC media-path network requirements

### LINKED ISSUE #270: Document Lingbot-World WebRTC Network Requirements
FlashDreams Lingbot-World docs now cover asset staging, disk usage, VRAM, and first-run behavior, but they still do not clearly document the WebRTC media-path networking requirements.

On cloud GPU instances where users only have SSH/TCP forwarding, the HTTP viewer page may load, but WebRTC media can still fail because the browser/server peer connection needs appropriate WebRTC connectivity.

Requested work:
- Document that Lingbot-World WebRTC requires a working WebRTC media path, not just access to the HTTP /request_session page.
- Explain limitations of SSH/TCP-only forwarding for WebRTC media.
- Add guidance for supported remote setups, such as direct network access, proper UDP forwarding, VPN, or TURN/STUN if applicable.
- Add a troubleshooting note for “viewer loads but video never appears.”

Acceptance criteria:
- Users can tell before launch whether their remote/cloud setup supports Lingbot-World WebRTC.
- The docs explain what to check when the page loads but media does not stream.

mvanhorn:
## Summary

Lingbot-World users on cloud GPU instances can now tell before launch whether their setup supports the WebRTC video path. This adds a "Network requirements (WebRTC)" section to `docs/source/models/lingbot_world.rst` explaining that reaching the HTTP `/request_session` page is necessary but not sufficient, that WebRTC needs a working media path (typically UDP) for ICE/DTLS/SRTP, and which remote options work (direct network access, UDP forwarding/VPN, or a TURN/STUN relay).

## Why this matters

Issue #270 (https://github.com/NVIDIA/flashdreams/issues/270) describes the "viewer loads but video never appears" symptom: with SSH/TCP-only port forwarding the HTTP viewer page loads while the WebRTC peer connection silently fails. The new section ends with a short troubleshooting note pointing at the media path so users know what to check. Claims stay tied to standard WebRTC behavior; no specific port numbers or a bundled TURN server are asserted since none appear in the repo.

## Testing

Docs-only change covered by the `doc.yml` Sphinx build: the new section renders with no RST syntax or duplicate-label warnings, and the page already in the models toctree continues to build. The file already carries its SPDX header, so reuse-lint stays green. Full build runs in CI.

Fixes #270


jarcherNV:
Review with Codex GPT 5.5 for PR #287 -> Issue #270
Mostly addresses it, but needs one wording fix before closing #270.

Thanks for putting this together. The new WebRTC network section is the right shape for #270: it makes clear that /request_session only proves the HTTP/signaling path, calls out the SSH -L / TCP-only forwarding limitation, and gives users a concrete “viewer loads but video never appears” thing to check.

Before we merge, could you tighten the STUN/TURN wording so it matches what the current LingBot viewer actually supports? The client creates RTCPeerConnection() with default config, and the server path does not expose a user-facing ICE server / TURN configuration. STUN also is not a relay when direct media connectivity is blocked. I’d either frame TURN/STUN as requiring additional implementation/configuration not currently exposed, or limit the supported setup guidance to direct network access, UDP-capable forwarding, or VPN for this PR.

liruilong940607:
Thanks for adding this instruction!

This feels more like a "potential issue and solution" instead of a general instruction that every user need to read. And if  we keep adding things like this the doc page will grow very fast in length.

I would suggest we move this into a "potential issue and solution" section and fold it by default.

liruilong940607:
I just saw this other PR about trouble shooting -- I think this belongs perfectly to the troubleshoting page. And this is a shared troubleshoting between both lingbot and omnidreams [and in the future more models] for webrtc demo

jarcherNV:
I think the core #270 fix is still needed, but now that #286 has merged, this PR should be reshaped before merging.

Remaining work:

- Move most of the WebRTC media-path explanation into `docs/source/troubleshooting.rst`, or fold it into the existing WebRTC troubleshooting entry added by #286. This is shared guidance for LingBot, OmniDreams, and future WebRTC demos.
- Keep only a short note/link in `docs/source/models/lingbot_world.rst` near the SSH/Brev forwarding commands, making clear that those commands expose the HTTP/signaling path only.
- Update the troubleshooting text for “page loads but video never appears” so it covers missing WebRTC media-path connectivity, not just browser ICE/mDNS behavior.
- Tighten the STUN/TURN wording: STUN is not a relay, and LingBot does not currently expose user-facing TURN/ICE-server configuration. Safer guidance is direct network access, UDP-capable forwarding/VPN, or TURN only if additional relay configuration/implementation exists.
- Run the docs/Sphinx validation after the RST changes.

I think the right next step is to integrate this content with the new troubleshooting page and leave the LingBot page with a concise pointer.

mvanhorn:
Reshaped in d9a946d, rebased onto main so it builds on #286's shared entry. All four points: the media-path explanation now lives in the webrtc-troubleshooting section (kept model-agnostic for LingBot/OmniDreams and future demos), lingbot_world.rst keeps only a short note + cross-reference by the SSH/Brev commands making clear those forward the HTTP/signaling path only, the 'page loads but video never appears' entry now covers missing media-path connectivity ahead of the mDNS case, and the STUN/TURN wording is corrected - STUN is not a relay, and TURN is only there if you deploy it. Couldn't run a local sphinx build, so worth a docs CI pass before merging.

liruilong940607:
/ok to test [286537c](https://github.com/NVIDIA/flashdreams/pull/287/commits/286537ccb50f0c931f3c7f6d5e7b2c45260c3e22)

liruilong940607 (approved):

mvanhorn:
Appreciate you merging the WebRTC network-requirements doc, @liruilong940607.

mvanhorn:
Thanks for taking the docs one too, @liruilong940607 - the WebRTC media-path network requirements were a real gap.

PR #297, opened 2026-06-04
Fix REUSE lint SPDX echo parsing

jmccaffrey-nv:
Summary:
- Wrap echoed SPDX header guidance in REUSE ignore markers so reuse lint does not parse example strings as license expressions.

Test:
- reuse lint


jarcherNV (approved):
LGTM

PR #316, opened 2026-06-09
Add VSR video postprocessing pipeline

gtong-nv:
# Add VSR Video Postprocessing Pipeline

## Summary

Adds a generic video post-processing layer that FlashDreams runners can apply after generation and before persisting MP4 output. The chain is a **no-op by default** and activates only when a preset or explicit processor list is configured.

The first concrete backend is **FlashVSR**, registered as lazy, swappable post-processor presets. Runners call a shared `Runner.postprocess_video_tensor()` hook instead of each integration owning bespoke upscaling logic.

OmniDreams gets a dedicated output path: when post-processing is enabled it frees diffusion VRAM, upscales generated RGB per view, and writes postprocessed video instead of the HDMap/RGB stacked canvas.

## Motivation

- Upscale generated video in the same `flashdreams-run` command, without a separate FlashVSR runner or gRPC service.
- Keep a reusable post-processing contract for future backends (denoise, color correction, etc.).
- Preserve existing runner behavior and eval/canary tooling when post-processing is disabled.

### Core post-processing infrastructure

- Added `flashdreams.infra.postprocess` with:
  - `VideoSpec`, `VideoChunk`, processor/session ABCs
  - `VideoPostprocessChainConfig` / `VideoPostprocessChainSession`
  - Layout conversion for common RGB tensor layouts (`tchw`, `bvtchw`, `bcthw`, …)
  - Value-range normalization (`minus_one_one`, `zero_one`, `uint8`)
  - `postprocess_video_tensor()` helper and chunk concatenation utilities
- Added `RunnerConfig.postprocess` (defaults to an empty chain) and `Runner.postprocess_video_tensor()`.

### Preset discovery

- Added `discover_postprocess_presets()` / `resolve_postprocess_preset()` in `flashdreams.plugins.registry`.
- Presets are registered via the `flashdreams.postprocess_presets` setuptools entry-point group.
- FlashVSR registers three presets in `integrations/flashvsr/pyproject.toml`:
  - `flashvsr-v1.1-sparse-2.0` (default stable sparse preset)
  - `flashvsr-v1.1-sparse-1.5` (faster sparse preset)
  - `flashvsr-v1.1-full-attn` (dense attention; suitable for multi-GPU)

### FlashVSR backend

- Added `integrations/flashvsr/flashvsr/postprocess.py`:
  - Lazy FlashVSR pipeline build from first stream dimensions
  - Chunk coalescing into FlashVSR cold-start / steady-state sizes
  - Tail handling via `replicate_pad` (preserve all frames) or `drop`
  - `finalize()` called in `try/finally` on generation failures
- Added CPU tests with mocked pipeline orchestration in `integrations/flashvsr/tests/test_postprocess.py`.

### Runner wiring

Post-processing is applied **before MP4 write** in these offline runners:

| Integration | Notes |
| --- | --- |
| `wan21` | Single-stream T2V output |
| `cosmos_predict2` | Single-stream output |
| `causal_forcing` | Single-stream output |
| `fastvideo_causal_wan22` | Single-stream output |
| `lingbot` | Single-stream output |
| `self_forcing` | Single-stream output |
| `omnidreams` | Multi-view path; frees pipeline VRAM before VSR; writes postprocessed RGB layout when enabled |

When post-processing is disabled, runner outputs are unchanged.

### Tests

- `flashdreams/tests/test_video_postprocess.py` — generic chain/layout/value-range behavior
- `flashdreams/tests/test_postprocess_presets.py` — preset discovery and chain resolution
- `integrations/flashvsr/tests/test_postprocess.py` — FlashVSR chunking/tail/error-path behavior

## CLI

Post-processing is configured through the standard FlashDreams config CLI on `RunnerConfig.postprocess`.

Primary flag:

| Flag | Type | Default | Description |
| --- | --- | --- | --- |
| `--postprocess.preset` | string | *(empty)* | Registered post-processor preset slug from `flashdreams.postprocess_presets` |

Available FlashVSR presets:

- `flashvsr-v1.1-sparse-2.0`
- `flashvsr-v1.1-sparse-1.5`
- `flashvsr-v1.1-full-attn`

Example — upscale Wan2.1 T2V output with the default FlashVSR preset:

```bash
flashdreams-run \
  --postprocess.preset flashvsr-v1.1-sparse-2.0 \
  wan21-t2v-1.3b-480p \
  --prompt "A cat surfing."
```

Inspect the resolved runner config without loading models:

```bash
flashdreams-run --no-instantiate \
  --postprocess.preset flashvsr-v1.1-sparse-2.0 \
  wan21-t2v-1.3b-480p
```


gtong-nv:
/ok to test [91f0c21](https://github.com/NVIDIA/flashdreams/pull/316/commits/91f0c21addbe95a9389c6487453596cb59e64a06)

integrations/lingbot/lingbot/runner.py
```diff
@@ -290,6 +290,12 @@ def run(self) -> None:
         video = torch.cat(chunks, dim=0)  # [T, C, H, W]
         if not self.is_rank_zero:
             return
+        video = self.postprocess_video_tensor(
```
wilsonCernWq, line 293: I think post-processing is more than just upsampling? 

If we want to make post-processing a polymorphic API, then I would prefer user to explicitly specify which post-operation to apply here?

flashdreams/flashdreams/scripts/cli.py
```diff
@@ -104,6 +185,11 @@ def entrypoint() -> None:
                 bool,
                 dataclasses.field(default=False),
             ),
+            (
+                "postprocess",
+                CLIVideoPostprocessConfig,
```
wilsonCernWq: I actually feel postprocessing is better to be an optional step in the pipeline?

@liruilong940607 what do you think?

flashdreams/tests/test_video_postprocess.py
```diff
@@ -0,0 +1,181 @@
...
+    resize_bvtchw,
+    to_bvtchw,
+    to_minus_one_one,
+)
+
+pytestmark = pytest.mark.ci_cpu
+
+
+@dataclass(kw_only=True)
+class _RepeatSpatialConfig(VideoPostProcessorConfig):
```
wilsonCernWq: why we have these tests? Are these example post processing ops? I feel this is way more than needed? 

I usually don't like pre-mature optimization/generation until we really have more than one use-case?

flashdreams/flashdreams/scripts/cli.py
```diff
@@ -65,6 +105,47 @@ def main(config: RunnerConfig, no_instantiate: bool = False) -> None:
...
+    processor = FlashVSRPostProcessorConfig(
+        scale=postprocess.scale,
+        chunk_size=postprocess.chunk_size,
+        sparse_ratio=sparse_ratio,
+        attention_mode=attention_mode,
+        device=postprocess.device,
+        tail_policy=postprocess.tail_policy,
+        compile_network=postprocess.compile_network,
+        use_cuda_graph=postprocess.use_cuda_graph,
+    )
```
wilsonCernWq: here we implicitly allow all user to instantiate flashvsr post processing, won't it create some kind of circular dependency for flashvsr cli?

also, I suggest each integration's cli to explicitly create a flashvsr postprocessing op via a command line string (default to flashvsr)?

integrations/flashvsr/flashvsr/postprocess.py
```diff
@@ -0,0 +1,253 @@
...
+    kv_ratio: int = 3
+    """Number of prior chunks kept by FlashVSR's streaming attention cache."""
+
+    local_range: int = 11
+    """Local window radius used by FlashVSR sparse attention."""
+
+    attention_mode: Literal["sparse", "full"] = "sparse"
+    """Attention backend. Use ``"full"`` for multi-GPU context parallelism."""
+
+    compile_network: bool = False
```
wilsonCernWq: why false?

integrations/flashvsr/flashvsr/postprocess.py
```diff
@@ -0,0 +1,253 @@
...
+    """Enable ``torch.compile`` for FlashVSR components."""
+
+    use_cuda_graph: bool = False
```
wilsonCernWq: like wise?

integrations/lingbot/lingbot/runner.py
gtong-nv: yes, postprocessing can be more than just upsampling. The user specifies that in the config and any those postprocessors will be registered and dispatched from here.

User explicitly specify which post-operation via cmd option or config, not here

flashdreams/tests/test_video_postprocess.py
gtong-nv: I will remove this test, as only omnidreams (which outputs hdmap+rgb stacked video) will benefit from it and not other model

integrations/flashvsr/flashvsr/postprocess.py
gtong-nv: Changed it to True for best performance. Would need to profile. Might makes more sense for one-shot short videos processing rather than long time serving...

integrations/flashvsr/flashvsr/postprocess.py
gtong-nv: Changed to True

flashdreams/flashdreams/scripts/cli.py
gtong-nv: Refactored. It is optional now, if not configured, then the postprocessing is a no-op

flashdreams/flashdreams/scripts/cli.py
gtong-nv: Refactored. Now the runner entry specifies the model and postprocessing is configured via 
`--postprocess.preset`

gtong-nv:
/ok to test [89aab9c](https://github.com/NVIDIA/flashdreams/pull/316/commits/89aab9cdb75d5cbc3d65904c0bd03e7e15c7d746)

gtong-nv:
/ok to test [c140850](https://github.com/NVIDIA/flashdreams/pull/316/commits/c140850a7c6092675db09a8787fb1d7964e20be3)

gtong-nv:
/ok to test [c46b7e8](https://github.com/NVIDIA/flashdreams/pull/316/commits/c46b7e8adb8acc1b0a984021c7645a6a908a2ecc)

flashdreams/flashdreams/infra/postprocess/base.py
```diff
@@ -0,0 +1,399 @@
...
+def postprocess_video_tensor(
+    tensor: Tensor,
+    *,
+    layout: VideoTensorLayout,
+    value_range: VideoValueRange,
+    postprocess: VideoPostprocessChainConfig,
+    fps: float | None = None,
+) -> Tensor:
+    """Run a complete post-processing chain over one in-memory video tensor."""
+    spec = infer_video_spec(tensor, layout=layout, fps=fps)
```
wilsonCernWq, line 239: Even when postprocessing is disabled, this still runs through the full chunk path and concatenates the video again. Since all the runners now call postprocess_video_tensor() unconditionally, default runs get an extra full-video copy/allocation before .cpu(). For large outputs this could be enough to introduce an OOM or a noticeable memory regression.

Can we add a fast path here, e.g. return tensor directly when postprocess.is_enabled() is false?

integrations/wan21/wan21/runner.py
```diff
@@ -192,7 +192,12 @@ def run(self) -> None:
         stats = self.pipeline.finalize(autoregressive_index=0, cache=cache)
         if not self.is_rank_zero:
             return
-        generated = generated.cpu()
+        generated = self.postprocess_video_tensor(
+            generated,
+            layout="tchw",
+            value_range="minus_one_one",
+            fps=config.fps,
+        ).cpu()
```
wilsonCernWq, line 200: I’m worried about the memory behavior for the non-OmniDreams runners. OmniDreams frees the generation pipeline before running FlashVSR postprocess, but WAN/Cosmos/etc. keep their original pipeline alive and then the postprocessor builds a second CUDA pipeline.

Since the CLI example shows FlashVSR postprocess with WAN, this path seems likely to OOM on realistic workloads unless we release the original pipeline/cache first, or document/guard that this requires enough extra VRAM.

integrations/flashvsr/flashvsr/postprocess.py
```diff
@@ -0,0 +1,296 @@
...
+)
+"""Default FlashVSR post-processing preset (stable sparse attention)."""
+
+POSTPROCESS_PRESET_FLASHVSR_V1_1_SPARSE_1_5 = FlashVSRPostProcessorConfig(
+    sparse_ratio=1.5,
+    attention_mode="sparse",
+)
+"""Faster FlashVSR post-processing preset (lower sparse attention budget)."""
+
+POSTPROCESS_PRESET_FLASHVSR_V1_1_FULL_ATTN = FlashVSRPostProcessorConfig(
```
wilsonCernWq, line 293: The full-attention is designed for multi-GPU, but postprocessing currently runs only after the runners return on non-rank-zero ranks. If FlashVSR full-attn enables context-parallel, rank 0 may enter distributed ops while the other ranks have already skipped this path, so this might break?

Can we either make distributed postprocess run on all ranks, or reject the full-attn preset when used through the current rank-zero-only runner output path?

wilsonCernWq (changes requested):

flashdreams/flashdreams/infra/postprocess/base.py
gtong-nv: Right.. I thought it was no-op but some redudant conversion is still involved. Added a postprocess.is_enabled() check

integrations/flashvsr/flashvsr/postprocess.py
gtong-nv: I added a wrap to use FlashVSR full-attn CP when multi-GPU run is lanched. I will test it tomorrow with a node.

integrations/wan21/wan21/runner.py
gtong-nv: good call! I just tested out WAN and didn't see OOM, but yea, this is a legit concern.
Maybe we should just drop the generation pipeline for the current `generate → postprocess → write output` flow?
Let's discuss this tomorrow.

jarcherNV:
This looks close from a triage pass: CI is green and Greptile’s latest summary does not flag functional blockers.

The remaining open items seem to be the two unresolved threads:

- memory behavior for non-OmniDreams runners when FlashVSR postprocess builds a second CUDA pipeline
- whether the full-attn preset is safe in the current rank-zero-only output/postprocess path

@gtong-nv did you get a chance to test or decide on those? Is the current behavior acceptable?

@wilsonCernWq if those two points are addressed, could you re-review or resolve the threads?

wilsonCernWq:
> if those two points are addressed, could you re-review or resolve the threads?

Yes, happy to close this. But idk if @liruilong940607 has anything extra to say? I remember he wanted to further discuss the design a bit more, but I am not sure if how did the discussion go.

gtong-nv:
oh, sorry let's close this one. I discussed with @liruilong940607 and redesigned the API based on his feedback in https://github.com/NVIDIA/flashdreams/pull/362. Thanks for reviewing!

PR #317, opened 2026-06-09
Add manifest-driven video quality regression harness

jarcherNV:
Implements the initial evaluate-only regression infrastructure with a case
manifest, schema validation, deterministic core metrics, synthetic calibration
sentinels, artifact-emitting runner, manual/scheduled GitHub Actions workflow,
and CPU tests.


.github/workflows/video-quality-regression.yml
```diff
@@ -0,0 +1,91 @@
...
+            --manifest configs/video_quality_cases.yml
+            --suite "${{ github.event.inputs.suite || 'calibration' }}"
+            --output-dir artifacts/video_quality
+          )
+          if [ -n "${{ github.event.inputs.case_id }}" ]; then
+            args+=(--case-id "${{ github.event.inputs.case_id }}")
+          fi
+          if [ -n "${{ github.event.inputs.hf_revision }}" ]; then
+            args+=(--hf-revision "${{ github.event.inputs.hf_revision }}")
+          fi
```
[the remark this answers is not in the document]
jarcherNV: Fixed.

tools/video_quality/manifest.py
```diff
@@ -0,0 +1,361 @@
...
+    if not isinstance(out, str) or not out:
+        raise ValueError(f"{key} must be a non-empty string when set")
+    return out
+
+
+def _required_int(value: dict[str, Any], key: str, *, context: str) -> int:
+    out = value.get(key)
+    if not isinstance(out, int):
+        raise ValueError(f"{context}.{key} must be an integer")
+    return out
```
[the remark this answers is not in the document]
jarcherNV: Fixed.

tools/video_quality/manifest.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Compatibility imports for the installable video-quality manifest module."""
+
+from flashdreams.quality.video_quality.manifest import (
```
gtong-nv: Do we really need these wrappers in `tools/quality` and the implementaions in `flashdreams/quality`? Seems redudant to me? Maybe move everything to `tools/` ?

configs/video_quality_cases.yml
```diff
@@ -0,0 +1,75 @@
...
+  - calibration
+  - per_commit
+  - nightly
+  - quarantine
+  - vlm_experimental
+
+default_hf_dataset: nvidia/flashdreams-video-quality-regression
+
+cases:
+  - id: synthetic_core_metric_sentinels
```
gtong-nv, line 16: not sure how can we simplify this manfiest list. It's seems difficult to add another case, given so many fileds and not sure which ones are required or optional.

configs/video_quality_cases.yml
```diff
@@ -0,0 +1,75 @@
...
+    source:
+      type: synthetic
+      fps: 8
+      frames: 16
+      height: 64
+      width: 64
+      seed: 17
+    assets:
+      known_good: synthetic://textured_motion
+      known_bad:
```
gtong-nv, line 32: For the metrics, `known_good` makes sense as the golden reference, what are `known_bad` used for?

configs/video_quality_cases.yml
```diff
@@ -0,0 +1,75 @@
...
+          expected_failures: [stripe_axis_max, stripe_row_autocorr_max]
+      sha256: {}
+    generation:
+      runner: synthetic
+      seed: 17
+      duration_s: 2
+      extra_args: []
+    windows:
+      head: {start_s: 0, end_s: 1}
+      tail: {start_s: 1, end_s: 2}
```
gtong-nv: maybe we should move all these properties from this config file to the implementation of that particular test? I would image not all tests need them

configs/video_quality_cases.yml
```diff
@@ -0,0 +1,75 @@
...
+    thresholds:
+      - id: luma_std_min
```
gtong-nv, line 44: So looks like this just defines the the failing criteria, and those matrics like `luma_std`, `grey_pixel_ratio` will always be computed?
I'm wondering how can we add another case, that might not need those matrics, but another parituclar one (e.g. MUSIQ) that is known to capture that type of regression. 
I was thinkg if we can decopule the test cases (which defines input and selected of metrics) and the metrics can be run.

tools/video_quality/manifest.py
jarcherNV: Agreed that the wrappers were too redundant. I removed the manifest, metrics, and evaluators compatibility wrappers from tools/. I kept the implementation in flashdreams.quality.video_quality because importing top-level tools from tests caused CI collection issues after package install. tools.video_quality.run_regression remains only as the CLI entry point.

configs/video_quality_cases.yml
jarcherNV: Good point. I simplified the starter case and added a manifest comment documenting the minimal case shape: id, description, suites, metrics, assets, and thresholds. generation and windows are now omitted from the synthetic case and only needed when a model runner or windowed metric requires them.

configs/video_quality_cases.yml
jarcherNV: Added an inline comment explaining this. known_bad clips are calibration fixtures, not production golden references. The runner expects them to fail exactly the listed thresholds, which proves the selected metrics catch known failure modes and also catches metrics that become too broad/noisy.

configs/video_quality_cases.yml
jarcherNV: I removed generation and windows from this synthetic manifest case since it does not need them. The schema still supports those fields for real model-generation cases and tail/head metrics, but cases no longer need to carry them by default.

configs/video_quality_cases.yml
jarcherNV: Agreed. I added explicit per-case metrics: [...] selection. Thresholds now define pass/fail criteria, while metrics declares which metric groups the case asks the runner to compute. This gives us a cleaner path to add future metric groups like MUSIQ without forcing every case to run every metric.

gtong-nv (approved):

PR #324, opened 2026-06-10
test(omnidreams): same-seed bitwise-reproducibility GPU test + nightly CI

jmccaffrey-nv:
## What

Adds a GPU test that pins **bitwise reproducibility at the same seed** for the OmniDreams distilled runner, plus a nightly CI job that runs it.

The test (`integrations/omnidreams/tests/test_omnidreams_same_seed_reproducibility.py`) runs the distilled runner (`omnidreams-sv-2steps-chunk2-loc6-lightvae-lighttae`) **twice at the same seed** under PyTorch's strict-determinism flags and asserts the two output MP4s are **byte-identical** (sha256):

- `CUBLAS_WORKSPACE_CONFIG=:4096:8`
- `torch.use_deterministic_algorithms(True, warn_only=True)`
- `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`

(flags set in a fresh subprocess before the first CUDA context, per the PyTorch reproducibility notes)

Unlike a golden-hash regression, it compares the two runs **against each other**, so there is no committed digest to maintain and it stays green across toolchain/hardware changes.

## How it's gated

- Tier-marked `ci_gpu` but **skips unless `OMNIDREAMS_REPRO_RUN` is set**, so the per-PR `pytest -m ci_gpu` job collects-and-skips it instantly.
- New nightly workflow `.github/workflows/determinism.yml` (cron + `workflow_dispatch`) sets `OMNIDREAMS_REPRO_RUN=1` to actually run it on the GPU runner.

(We avoid the `manual` marker on purpose: `pytest-manual-marker` xfails every `manual` test at setup, so it would never execute as a CI guard.)

## Verification

Ran on an H100 (slurm, cu130 container): two `seed=1` rollouts produced an identical MP4 — `1 passed in 384.73s`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/determinism.yml
+++ b/.github/workflows/determinism.yml
@@ -1,0 +1,83 @@
+name: Determinism (periodic)
+
+# Periodic bitwise-reproducibility guard for the omnidreams runner.
+#
+# The same-seed reproducibility test is marked `manual`: it needs a real GPU,
+# downloads the distilled checkpoint + an example HDMap clip from HF, and runs
+# two full rollouts, so it is too heavy for the per-PR `ci_gpu` job. Instead it
+# runs here on a nightly schedule (and on demand via the Actions "Run workflow"
+# button). Scheduled runs execute on the default branch only.
+on:
+  schedule:
+    # 07:00 UTC nightly (~midnight US Pacific).
+    - cron: "0 7 * * *"
+  workflow_dispatch:
+
+jobs:
+  reproducibility:
+    runs-on: linux-amd64-gpu-rtxpro6000-latest-2
+    timeout-minutes: 120
+    container:
+      image: nvidia/cuda:13.2.1-cudnn-devel-ubuntu24.04
+      options: --gpus all
+    env:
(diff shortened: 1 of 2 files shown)


jarcherNV:
Jon is away for a while, so I’m going to take this over rather than leave the PR waiting.

The remaining items look small: refresh the branch against current `main`, fix the workflow comment around the `ci_gpu` / `OMNIDREAMS_REPRO_RUN` gating, and decide whether the internal `setup-proxy-cache@main` action ref should stay floating or be pinned. I’ll recreate this on a fresh branch and reference this PR from the replacement.

jarcherNV:
I have created a new PR for this at https://github.com/NVIDIA/flashdreams/pull/366 to address the remaining issues from here.

Once that is merged, I will close this.

jarcherNV:
https://github.com/NVIDIA/flashdreams/pull/366 is now merged so I am closing this PR.

Issue #276, opened 2026-06-03
Support for Helios


SHYuanBest:
Hi! Thanks for great work! We recently release Helios, a 14B Real-Time Streaming Video Generation, would it be possible integrated into FlashDreams?

Code: https://github.com/PKU-YuanGroup/Helios
Page: https://pku-yuangroup.github.io/Helios-Page/

jmccaffrey-nv:
Hi @SHYuanBest  , thank you for the request and for you'alls great work with Helios, including its easy anti-draft and adversarial hierarchical distillation.  We'll evaluate priority to integrate this video model.

jmccaffrey-nv:
Hi @SHYuanBest , are you interested to try an external integration in Helios for flashdreams inference ?
via a standalone package that pip installs flashdreams and plugs in via the flashdreams.runner_configs entry point. We added an [integrate-a-model](https://github.com/NVIDIA/flashdreams/tree/main/agentic/skills/integrate-a-model)/SKILL.md to assist such integration.
We're managing how many models we integrate into the flashdreams repo, to control long-term maintenance.

jarcherNV:
Thanks again for the request. I’m taking over follow-up here while Jon is unavailable.

We are still interested in Helios support, but we are being careful about adding more first-party model integrations to the main FlashDreams repo because each one adds long-term maintenance cost. For now, the preferred path would be an external/plugin-style integration that depends on `flashdreams` and registers a runner through the `flashdreams.runner_configs` entry point.

The current integration guidance lives here:
https://github.com/NVIDIA/flashdreams/tree/main/skills/integrate-a-model

I’ll keep this open as an enhancement/backlog item, but we do not currently have an internal ETA for implementing Helios support ourselves.

PR #347, opened 2026-06-24
Add Helios 14B streaming T2V integration

shy982:
## Summary

Adds FlashDreams integration for [Helios](https://github.com/PKU-YuanGroup/Helios) 14B real-time streaming text-to-video, addressing #276.

- New plugin under `integrations/helios/` wrapping `HeliosPyramidPipeline` from diffusers
- Three registered runner slugs:
  - `helios-distilled-t2v-14b` — `BestWishYsh/Helios-Distilled`, pyramid `[2,2,2]`
  - `helios-base-t2v-14b` — `BestWishYsh/Helios-Base`, pyramid `[20,20,20]`, CFG 5.0
  - `helios-distilled-t2v-14b-2gpu` — distilled checkpoint with Ulysses context parallelism (`torchrun`)
- Each `generate()` call yields one native 33-frame chunk through the FlashDreams streaming interface
- CPU smoke tests for runner registration and entry-point discovery
- Model gallery docs at `docs/source/models/helios.rst`

## Requirements note

`HeliosPyramidPipeline` requires a recent diffusers build. If the resolved PyPI release does not export it, install from source:

```bash
pip install git+https://github.com/huggingface/diffusers.git
uv sync --project integrations/helios

### CODE REVIEW

File: integrations/helios/helios/encoder.py [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+        negative_prompt: list[str],
+        device: torch.device,
+        *,
+        guidance_scale: float,
+        image: Optional[Image.Image] = None,
+    ) -> HeliosConditionings:
+        do_cfg = guidance_scale > 1.0 and not getattr(
+            self.pipe.config, "is_distilled", False
+        )
+        pe, npe = self.pipe.encode_prompt(
```
liruilong940607, line 44: is Helios using the same text encoder as WAN? if so maybe we should use what is already present in flashdreams to avoid create multiple code path doing the same thing and reduce the diffuser dependency.

https://github.com/NVIDIA/flashdreams/blob/main/flashdreams/flashdreams/infra/encoder/text/umt5.py

integrations/helios/helios/helios_loader.py
```diff
@@ -0,0 +1,91 @@
...
+from __future__ import annotations
+
+from typing import Any, Type
+
+import torch
+
+
+def get_helios_pipeline_class() -> Type[Any]:
+    try:
+        from diffusers import HeliosPyramidPipeline
```
liruilong940607, line 13: We should try using the components carried in flashdreams (text encoder, VAE, transformers) to implement Helios instead of importing the diffuser version so that it can benefit the optimizations in flashdreams.

liruilong940607 (changes requested):

jarcherNV:
Thanks for working on this. I think we should keep this open as the current Helios integration attempt, but it is not ready for merge yet.

The main blocker is the implementation direction Ruilong raised: before we add this as a first-party integration, we should decide whether Helios can reuse existing FlashDreams components for the text encoder / VAE / transformer path instead of depending primarily on the Diffusers pipeline wrapper. That would better match the rest of FlashDreams and avoid creating a parallel implementation path.

There are also a few concrete issues to address before another review pass:

- The branch is currently conflicted with `main`.
- The benchmark path currently appears to crash because it repeatedly measures the same autoregressive index.
- The video-conditioning fallback should not append discontinuous fallback frames into the AR history.
- The frame range normalization should rely on a known output contract rather than inferring `[0, 1]` vs `[-1, 1]` from tensor min/max.
- The unused optimized config should either be wired up, removed, or clearly marked as deferred.

Once those are addressed, we can do a proper review pass.

PR #348, opened 2026-06-24
Add LTX-Video 2B streaming T2V integration

shy982:
## Summary

Adds a first-party FlashDreams integration for [LTX-Video](https://github.com/Lightricks/LTX-Video) 2B causal streaming text-to-video.

- New plugin under `integrations/ltx_video/` wrapping `LTXPipeline` from diffusers
- Three registered runner slugs:
  - `ltx-video-t2v-2b` — streaming wrapper using native `pipe()` per chunk
  - `ltx-video-t2v-2b-optimized` — manual denoise loop with KV-cache, `torch.compile`, and FlashAttention
  - `ltx-video-t2v-2b-taehv` — optimized path plus TAEHV fast decoder
- CPU smoke tests for runner registration and entry-point discovery
- GPU optimization tests (`ci_gpu`) validating each optimization layer in isolation
- Model gallery docs at `docs/source/models/ltx_video.rst`

## Performance note

The optimized runner improves time-to-first-frame after warmup. Steady-state per-chunk throughput vs the streaming runner is still being tuned (KV window size, compile/graph strategy, denoise loop overhead).

## Test plan

- [x] `uv sync --project integrations/ltx_video --extra dev`
- [x] `uv run --project integrations/ltx_video pytest integrations/ltx_video/tests/test_smoke.py -v` (5 passed)
- [x] `pytest integrations/ltx_video/tests/test_optimizations.py -v -m ci_gpu` (7 passed on H100, cu128 env)
- [x] `uv run --project integrations/ltx_video flashdreams-run --help` lists all three LTX slugs
- [ ] End-to-end: `uv run --project integrations/ltx_video flashdreams-run ltx-video-t2v-2b --total-blocks 7 --pixel-height 512 --pixel-width 768` (requires HF weights)


liruilong940607:
Same comments as here: https://github.com/NVIDIA/flashdreams/pull/347#discussion_r3494011336

jarcherNV:
Thanks for working on this. I think this needs another revision before review/merge.

The main open design question is the same as on #347: before we add this as a first-party integration, we should decide whether LTX should reuse existing FlashDreams components where possible rather than primarily wrapping the Diffusers pipeline path.

There are also several concrete blockers:

- The branch is currently conflicted with `main`.
- The OSRB collateral sanity check is failing.
- The KV-cache path appears to duplicate past tokens each AR step, which would corrupt the optimized runner output.
- The KV context should not be a module-level global if this may run in a concurrent/server environment.
- The benchmark path currently appears to crash because it measures a later AR step with a fresh cache.
- The `kv_cache=True` / `cuda_graphs=True` combination should be explicitly guarded or made safe.
- `low_cpu_mem_usage=False` should be reconsidered for the 2B loader.

One possible path forward is to narrow this PR to the baseline non-KV LTX runner first, then add the optimized/KV-cache/TAEHV paths in follow-up PRs once the cache correctness and benchmark coverage are solid.

PR #349, opened 2026-06-24
Add CUDA 12.8 resolver constraints for uv pip workflows

wilsonCernWq:
## Summary                                                                    
                                                                  
  - Add a `cuda12` dependency group as a Linux opt-in for CUDA 12.8 torch
    wheels (`uv sync --group cuda12 --extra runners`). Default sync still     
    installs the CUDA 13 stack — PyPI on Linux, NVIDIA's cu130 index on         
    Windows.                                                                  
  - Reintroduce the Windows-only cu130 source binding that the previous         
    attempt removed. PyPI's Windows torch wheel is the CPU build, so CUDA
    extensions (gated on `torch.version.cuda is not None`) need NVIDIA's        
    index.                                                                      
  - Drop `torch>=2.9` from the `docs-ci` group and mock `torch` /             
    `torchvision` in `autodoc_mock_imports`, so                                 
    `uv sync --only-group docs --only-group docs-ci --python 3.12`            
    resolves cleanly.                                                           
                                                                              
  ## Why these pieces have to land together                                     
                                                                  
  The CUDA-flavor selection is built from three coupled uv mechanisms,          
  each unlocking the next:                                                    
                                                                                
  1. **cu128 source bound to `cuda12` (Linux marker).** Routes torch /        
     torchvision to the pytorch-cu128 index when the group is active.           
     The `sys_platform != 'win32'` marker is what makes the cu128 / cu130
     source domains physically disjoint.                                        
  2. **cu130 source bound unconditionally on Windows.** uv's                    
     `default-groups` does not propagate per-group source bindings, so a      
     group-gated cu130 binding would not fire on plain `uv sync` on             
     Windows. Group-free + Windows marker is the only shape that gives        
     Windows default the cu130 wheel without requiring `--group cuda13`.        
  3. **`cuda13` placeholder group + `conflicts` peer.** Without the           
     conflicts declaration uv rejects the mixed group-conditional cu128         
     + unconditional cu130 sources as "conflicting indexes in all marker      
     environments", even though their `sys_platform` markers are                
     disjoint.                                                                
                                                                                
  The `docs-ci` change is required because the resulting workspace lock
  holds three torch candidates (PyPI / +cu128 / +cu130) and uv can't pick       
  one when only `docs` / `docs-ci` groups are active — no source binding        
  fires for torch in that scope.                                              
                                                                                
  ## File layout                                                  
                                                                                
  - `pyproject.toml` (root): unchanged for torch wiring. Only the 
    `docs-ci` group lost its `torch>=2.9` entry, and its comment now          
    notes that torch is mocked rather than installed.                           
  - `flashdreams/pyproject.toml`: owns the two `[[tool.uv.index]]`            
    entries, the `[tool.uv.sources]` bindings, the `cuda13` / `cuda12`          
    dependency groups, `default-groups`, and `conflicts`. Colocated with      
    the `torch>=2.9` `[project]` dep.                                           
  - `docs/source/conf.py`: `torch` and `torchvision` added to                   
    `autodoc_mock_imports`, alongside the existing GPU-package mocks.


jarcherNV (approved):
LGTM

PR #356, opened 2026-06-29
feat(wan-vae): default Wan 2.2 TI2V-5B VAE to the native .pth (no remap)

wenqingw-nv:
Supersedes #223 

Addresses review on #223 (name the component the same way it is named in the pth).

## What changed
- Renames the Wan 2.2 residual down/up blocks' grouped `resnets` ModuleList + separate `downsampler` / `upsampler` into a single flat `downsamples` / `upsamples` `Sequential` (residual blocks followed by the optional resample). This matches upstream's native `Wan2.2_VAE.pth` key layout **one-to-one**.
- **Removes** `_WAN22_TI2V_5B_VAE_PTH_KEY_REMAP` and `wan22_ti2v_5b_vae_pth_state_dict_transform` entirely — they're no longer needed. The Wan 2.2 VAE configs now load the native `.pth` with `state_dict_transform=None` (genuinely no remap runs).
- The diffusers `AutoencoderKLWan` remap is retargeted onto the new flat layout and kept as an opt-in fallback.

## Tests
- `test_wan22_vae_native_pth_loads_without_remap` (ci_cpu): the native layout matches the module tree 1:1 with no leftover grouped names.
- `test_wan22_vae_diffusers_remap_still_targets_real_params` (ci_cpu): the diffusers fallback still targets real module params after the rename.
- `test_wan22_vae_native_pth_and_diffusers_weights_identical` (manual): both load paths yield bit-identical weights.

Verified on CPU: 196-param module tree matches the native keys 1:1, forward smoke passes, `ruff check` + `format` clean (pinned v0.12.7).

### DIFF
--- a/flashdreams/flashdreams/recipes/wan/autoencoder/vae.py
+++ b/flashdreams/flashdreams/recipes/wan/autoencoder/vae.py
@@ -52,23 +52,18 @@ from flashdreams.infra.encoder import (
     StreamingVideoEncoder,
 )
 
-# Wan 2.2 TI2V 5B's VAE ships in the diffusers Wan-AI repo. The
-# loader pulls the diffusers safetensors shard and remaps keys via
-# :func:`wan22_ti2v_5b_vae_state_dict_transform` (``encoder.conv_in``
-# / ``quant_conv`` / ``post_quant_conv`` etc. -> our internal layout).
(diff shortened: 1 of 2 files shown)


liruilong940607:
/ok to test [8bf2c0f](https://github.com/NVIDIA/flashdreams/pull/356/commits/8bf2c0f6e0c88d8de8d9ef798da1983197606008)

liruilong940607 (approved):

PR #358, opened 2026-06-30
fix hy worldplay model card instructions

liruilong940607:
- fix hy worldplay model card instructions
- fix the omnidreams typing issues
- improve the plugin importing logic to avoid warning spam.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,7 +28,7 @@ command:
 
 ```bash
 # from the repo root
-uv run --group docs sphinx-build -b html docs/source docs/_build/html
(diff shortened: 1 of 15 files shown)


wenqingw-nv (approved):
LGTM. Docs-build invocation + autodoc mock-imports, model-card instructions matching the distilled-ckpt default, `from __future__ import annotations` across core, the plugin-discovery warning-spam fix (nicely scoped to ModuleNotFoundError for the plugin's own namespace, with tests), and the scipy Rotation typing guards all look good. Minor nit (non-blocking): `_compose_rotations` is duplicated across a couple of omnidreams modules — could live in a shared helper later.

PR #360, opened 2026-07-03
fix: cap numpy below 2.5 until mediapy supports PEP 695 NDArray

### LINKED ISSUE #355: Runner discovery fails on Python 3.12: mediapy 1.2.6 incompatible with numpy 2.5.0
## Summary

After `uv sync --extra runners` on Python 3.12, `flashdreams-run --help` logs warnings for 14 runner entry points that fail to import. The root cause is that **mediapy 1.2.6** (latest on PyPI, locked in `uv.lock`) crashes on import when **numpy 2.5.0** is installed (the version `uv.lock` resolves for Python >= 3.12).

## Environment

- Python 3.12
- `numpy==2.5.0` (from `uv.lock`)
- `mediapy==1.2.6` (from `uv.lock`, latest on PyPI)
- Install: `uv sync --extra runners`

## Reproduction

```bash
uv sync --extra runners
uv run python -c "import mediapy"
```

Fails with:

```
TypeError: typealias() takes exactly 2 positional arguments (3 given)
```

Traceback points to mediapy:

```python
class _VideoArray(npt.NDArray[Any]):
```

NumPy 2.5 redefined `numpy.typing.NDArray` as a PEP 695 type alias, which can no longer be subclassed this way.

Running `flashdreams-run --help` then emits warnings like:

```
Failed to load flashdreams runner entry point 'cosmos2-t2v-2b-720p' -> cosmos_predict2.config:RUNNER_COSMOS2_T2V_2B_720P
```

Any integration whose `config.py` imports `runner.py` at module level (which does `import mediapy`) is affected. In a fresh sync this includes 14 runners (cosmos2, wan21, causal_forcing, self_forcing, flashvsr, fastvideo, etc.). Omnidreams runners still appear because omnidreams lazy-imports mediapy.

## Impact

- Affected runners are silently skipped during plugin discovery (`flashdreams/flashdreams/plugins/registry.py` catches the import error and logs a warning).
- Users following the README quickstart may not notice until they try to run a non-omnidreams model.
- Core inference is unaffected; mediapy is only used for runner I/O (`read_image`, `read_video`, `write_video`).

## Proposed fix

Pin numpy below 2.5 until mediapy releases a compatible version, using the existing workspace override pattern in root `pyproject.toml`:

```toml
[tool.uv]
override-dependencies = [
    ...
    "numpy>=1.24,<2.5",
]
```

Verified workaround: `numpy==2.4.6` + `mediapy==1.2.6` imports cleanly and all runners register.

## Upstream

- mediapy: https://github.com/google/mediapy (no release fixing numpy 2.5 NDArray subclassing yet; 1.2.6 is latest on PyPI)
- numpy 2.5 typing change: `NDArray` is now a `TypeAliasType` (PEP 695)

mvanhorn:
## Summary

Caps numpy to `>=1.24,<2.5` in the uv override list so `import mediapy` works again on Python 3.12+; a fresh sync currently crashes at import time on every entry point that touches video I/O.

## Why this matters

numpy 2.5.0 turned `numpy.typing.NDArray` into a PEP 695 `TypeAliasType`, and mediapy 1.2.6 subclasses that alias during import, which raises `TypeError` the moment mediapy loads. Issue #355 reports exactly this on a clean `uv sync`: the resolver picks numpy 2.5.0 for Python 3.12+ (the 3.10/3.11 markers already resolved to 2.2.x/2.4.x, which is why older interpreters were unaffected). The cap pins the 3.12+ branch to numpy 2.4.6 in `uv.lock`, alongside the existing overrides in `pyproject.toml`, with a comment linking the issue so the cap can be dropped once mediapy ships a release compatible with numpy 2.5.

## Testing

`uv lock --check` passes (lock consistent with the new override, 251 packages resolved). The lock diff shows every Python 3.12+ numpy resolution moving from 2.5.0 to 2.4.6 and no other dependency movement. No numpy 2.5.0 references remain in `uv.lock`.

Fixes #355


jatentaki (approved):
LGTM, thanks a lot! Will try to follow up with hardening of CI to prevent this kind of breakages from being undetected.

mvanhorn:
Thanks for the quick turnaround, @jatentaki - the numpy cap keeps installs green until mediapy catches up.

mvanhorn:
Thanks for merging, @jatentaki - capping numpy below 2.5 should keep mediapy installs green until PEP 695 support lands.

PR #361, opened 2026-07-06
ci: validate locked Python 3.12 environment

jatentaki:
https://github.com/NVIDIA/flashdreams/pull/360 fixes a py312 crash due to a recent `numpy` version breaking compatibility, however it doesn't provide a means to detect such breakages in CI. This MR attempts that by

* Running CPU CI on Python 3.12 (previously: 3.10)
* Retaining Python 3.10 for GPU coverage
* Requiring all workflow syncs to use the committed `uv.lock` (no silent `uv.lock` upgrades in CI)
* Adding a `pre-commit` lockfile freshness check to alert developers of what would come up as CI breakage

By running CPU jobs with `py312` and GPU with `py310` we keep testing the lowest supported Python version but also have a way to catch simple regressions in a more modern version without increasing CI burden. This is a judgement call, we could choose to e.g. run GPU@310 and multiple CPU jobs @310,311,312,313.

Overall, this MR should make it much harder to have a similar breakage reappear on `main` in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,9 @@ jobs:
     env:
       UV_PROJECT_ENVIRONMENT: /tmp/flashdreams-venv
       UV_LINK_MODE: copy
-      UV_PYTHON: "3.10"
+      # Exercise the dependency branch used by current Python releases.
+      # GPU CI below remains on 3.10, the minimum supported version.
+      UV_PYTHON: "3.12"
     steps:
       - name: Get PR info
         if: startsWith(github.ref_name, 'pull-request/')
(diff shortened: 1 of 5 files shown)


jatentaki:
Pushed a commit addressing `uv run ...` (previously only `uv sync`); remove a redundant pre-publish step.

jatentaki:
Note the original CI run (before rebasing on #360) failed as expected: https://github.com/NVIDIA/flashdreams/actions/runs/28793756754/job/85378793414?pr=361#step:13:110

jarcherNV:
This looks good to me. The CI lockfile hardening is scoped and the current checks are passing.

One non-blocking follow-up: the WorldLens canary still uses `uv run --locked --with ...`, so those transient evaluator dependencies are not fully governed by the workspace lockfile. I do not think that should block this PR since that pattern predates this change, but if we want fully locked canary dependencies later we should probably move those into a dependency group or otherwise lock the evaluator environment explicitly.

jarcherNV (approved):
LGTM

PR #362, opened 2026-07-08
Postprocess Stream and FlashVSR Presets

gtong-nv:
# Postprocess Stream and FlashVSR Presets

## Summary

This PR adds a runner-level video post-processing path and wires FlashVSR as the first concrete post-processor preset. The main goal is to let existing autoregressive video runners inject streaming post-processing after decode without moving that concern into `StreamInferencePipeline`.

At a high level:

- `Runner` creates one `VideoPostprocessStream` per rollout.
- Each AR chunk still flows through `pipeline.generate(...)` and `pipeline.finalize(...)` exactly as before.
- The decoded chunk is handed to the stream with `stream.process(...)`.
- The stream applies the configured post-processing chain, collects emitted frames on rank zero, and flushes any buffered tail at end-of-stream.
- If post-processing is disabled, runners use an identity stream so their output collection code stays simple.

## Motivation

The existing pipeline timing covered encode, diffuse, decode, and finalize, but not work inserted after decode. FlashVSR post-processing can be substantial, especially when it buffers several small AR chunks before running an upsample window. Keeping it outside the pipeline preserves the core pipeline contract while making the extra work explicit and measurable.

The implementation also needed to solve a few runner-facing problems:

- Avoid duplicating collection and flush logic in every integration runner.
- Keep rank-zero persistence behavior unchanged for normal multi-GPU generation.
- Allow context-parallel post-processors to run on every rank when required.
- Preserve streaming post-processors that may consume an AR chunk without emitting frames yet.
- Keep the public postprocess API small enough for future integrations to implement.

## Public API

The public postprocessing API is exported from `flashdreams.infra.postprocess`:

- `VideoPostprocessChainConfig`
- `VideoPostprocessStream`
- `create_runner_postprocess_stream`
- `VideoPostProcessorConfig`
- `VideoPostProcessor`
- `VideoPostProcessorSession`
- `VideoChunk`
- `VideoSpec`
- `VideoTensorLayout`
- `to_bvtchw`

The runner config surface is intentionally small:

- `postprocess`: ordered postprocessing chain plus optional preset name.
- `postprocess_output_layout`: decoded runner output layout, required when output collection or postprocessing is used.
- `postprocess_per_view`: create one session per view for multi-view tensors.

Preset names are resolved from the `flashdreams.postprocess_presets` entry-point group. The FlashVSR integration registers:

- `flashvsr-v1.1-sparse-2.0`
- `flashvsr-v1.1-sparse-1.5`
- `flashvsr-v1.1-full-attn`

## Runner Flow

The intended runner shape is now:

```python
output_stream = self.create_postprocess_stream(fps=config.fps)

for i in range(config.total_blocks):
    video_chunk = self.pipeline.generate(autoregressive_index=i, cache=cache)
    stats = self.pipeline.finalize(autoregressive_index=i, cache=cache)
    output_stream.process(video_chunk, autoregressive_index=i)

generated = output_stream.finish()
if generated is None:
    return
```

This folds postprocess, buffering, flush, rank-zero collection, and CPU transfer into `VideoPostprocessStream`. Runners no longer need to manually maintain a `chunks` list or remember to append the postprocess tail.

## Call Graph

```mermaid
flowchart TD
    Runner["integration runner"]
    Create["Runner.create_postprocess_stream()"]
    Factory["create_runner_postprocess_stream()"]
    Stream["VideoPostprocessStream"]
    Process["VideoPostprocessStream.process()"]
    Apply["apply_video_postprocess()"]
    Session["VideoPostprocessChainConfig.setup()"]
    Chain["VideoPostprocessChainSession.process()"]
    ProcessorSession["VideoPostProcessorSession.process()"]
    FlashVSR["FlashVSR session buffer + pipeline.generate/finalize"]
    Finish["VideoPostprocessStream.finish()"]
    Flush["flush_video_postprocess()"]

    Runner --> Create --> Factory --> Stream
    Runner --> Process --> Apply
    Apply --> Session --> Chain --> ProcessorSession --> FlashVSR
    Runner --> Finish --> Flush --> Chain
```

## Chain and Session Model

`VideoPostProcessor` is the stateless factory object created from config. It owns configuration and starts a per-rollout `VideoPostProcessorSession`.

`VideoPostProcessorSession` owns stream state. For FlashVSR, this includes the lazily constructed FlashVSR pipeline, cache, frame buffer, AR index, and metadata spans.

`VideoPostprocessChainConfig.setup(spec)` resolves configured processors and preset processors, instantiates each processor, starts one session per processor, and returns a chain session. The chain session runs chunks through each processor in order. During `flush()`, tail output from an earlier processor is fed only into downstream processors, because it has already passed the earlier stages.

## Streaming and Buffering

The postprocess boundary uses `VideoChunk` objects. Each chunk contains a tensor, its declared layout, and optional metadata. Runners pass decoded AR chunks into the stream; the postprocessor may return zero, one, or multiple output chunks.

Returning zero chunks is important for FlashVSR. Generation AR chunks can be smaller than the FlashVSR window. In that case the FlashVSR session synchronously consumes the input, appends it to its internal buffer, and returns `[]`. A later AR chunk or `flush()` supplies enough frames to produce output.

To keep `VideoPostprocessStream.process()` tensor-only, empty processor output is represented as a zero-frame tensor in the caller's layout. The stream checks the layout-specific time dimension and skips collection for zero-frame tensors. Flush is different: it may legitimately have no tail, so `finish()` treats `None` as "nothing to append."

## Layout and Range Handling

Postprocessing expects RGB video in `[-1, 1]`. The supported layouts are:

- `tchw`
- `btchw`
- `bcthw`
- `bvtchw`

The canonical conversion helper is `[B, V, T, C, H, W]`. It is used only at postprocess boundaries and for concatenating chunks with potentially different source layouts. Shape inference does not convert tensors; `infer_video_spec_from_tensor_shape(...)` reads dimensions directly from the declared layout.

FlashVSR internally consumes contiguous `[B, C, T, H, W]`. The FlashVSR session converts each incoming chunk with:

```python
canonical[:, 0].permute(0, 2, 1, 3, 4).contiguous()
```

That `.contiguous()` can copy for time-major or view-major runner layouts. The tradeoff is intentional: the copy happens once at the FlashVSR boundary, after which buffering, concatenation, and FlashVSR model calls stay in FlashVSR's native layout. For large upscale runs this copy is expected to be small relative to FlashVSR inference, but it is still part of the postprocess timing logged by the stream profiler.

## Distributed Behavior

Generation and pipeline finalize still run on every distributed rank. Persistence remains rank-zero only.

For postprocessing:

- Disabled postprocessing uses an identity stream. Every rank can keep the same runner shape, but only rank zero collects and persists output.
- A normal postprocessor that does not require all ranks is created only on rank zero under multi-GPU execution.
- A context-parallel postprocessor returns `requires_all_ranks=True`; in that case every rank creates and runs the postprocessor, and rank zero remains the only collector.

FlashVSR sparse presets reject `world_size > 1`, because sparse FlashVSR postprocess is not distributed. The `flashvsr-v1.1-full-attn` preset opts into the all-rank context-parallel path.

## Profiling

When `config.pipeline.enable_sync_and_profile` is enabled, `VideoPostprocessStream` measures postprocess work with `EventProfiler` and logs:

- per-AR postprocess elapsed time
- input shape
- emitted output shape
- flush time and flush output shape

For all-rank processors the profiler synchronizes distributed ranks before summarizing, matching the execution contract for context-parallel postprocessing.

## FlashVSR Preset Behavior

FlashVSR is loaded lazily on the first chunk. The input frame size is validated against the first stream spec, the FlashVSR pipeline is moved to the configured device, and a FlashVSR cache is initialized for the postprocess session.

FlashVSR chunking follows its native first-chunk and steady-state target sizes. A partial final buffer is handled by `tail_policy`:

- `replicate_pad`: pad with the last frame, run FlashVSR, then trim back to the real tail length.
- `drop`: discard the incomplete tail.

The output spec reports the spatially upsampled size rounded down to FlashVSR's 128-pixel alignment.

## Tests and Validation

After rebasing onto `origin/main` at `87639f08`, the focused CPU validation passed:

```text
uv run pytest tests/test_agent_skills.py \
  flashdreams/tests/test_runner_postprocess.py \
  flashdreams/tests/test_postprocess_stream.py \
  flashdreams/tests/test_video_postprocess.py \
  flashdreams/tests/test_postprocess_presets.py \
  integrations/flashvsr/tests/test_postprocess.py
```

Result:

```text
29 passed
```

Previously captured GPU smoke runs for this branch:

- Self-forcing TAEHV 256x448 with `flashvsr-v1.1-sparse-2.0`: completed, emitted 21 frames at 512x896, peak allocated VRAM about 21.294 GiB.
- OmniDreams 704p with `flashvsr-v1.1-sparse-2.0`: completed, emitted 13 frames at 1408x2560, wall time 128.790s, peak allocated VRAM about 94.025 GiB and reserved VRAM about 120.705 GiB.

First GPU runs include compile and CUDA graph warmup, so those numbers should be treated as smoke/perf documentation rather than final benchmark claims.

## Compatibility and Migration Notes

Existing runners keep their generate/finalize behavior. The main migration is replacing manual chunk collection with `self.create_postprocess_stream(...)`, `stream.process(...)`, and `stream.finish()`.

Runners must declare `postprocess_output_layout` so the stream knows where the time axis is and how to infer video metadata. Multi-view runners that want independent view processing set `postprocess_per_view=True` and use `bvtchw`.

The postprocess implementation deliberately avoids exporting internal state classes. The public API describes configs, processor factories, processor sessions, chunks, specs, layouts, and the stream facade.

## Risks and Follow-Ups

- Layout conversion at the FlashVSR boundary can copy. The design contains this cost to the boundary and includes it in postprocess profiling.
- FlashVSR sparse presets are single-GPU only. Multi-GPU postprocessing requires the full-attention preset.
- The current per-view path is generic, but OmniDreams is the primary multi-view runner today. Future multi-view postprocessors should validate that they need generic per-view handling rather than model-specific logic.
- Additional postprocessors should follow `skills/flashdreams-postprocessing/SKILL.md` so the config/session split and runner stream behavior remain consistent.


liruilong940607 (approved):
LGTM!

gtong-nv:
/ok to test [07b80b3](https://github.com/NVIDIA/flashdreams/pull/362/commits/07b80b3d978e63c017fac1547fa4c3d391f53a48)

gtong-nv:
/ok to test [6c68d52](https://github.com/NVIDIA/flashdreams/pull/362/commits/6c68d5234414419fcba04475ce2eec705d4ca142)

gtong-nv:
/ok to test [84278a5](https://github.com/NVIDIA/flashdreams/pull/362/commits/84278a5b1a033f3fc88f535aeaa77079aec61b50)

gtong-nv:
/ok to test [84278a5](https://github.com/NVIDIA/flashdreams/pull/362/commits/84278a5b1a033f3fc88f535aeaa77079aec61b50)

gtong-nv:
/ok to test [4531fe9](https://github.com/NVIDIA/flashdreams/pull/362/commits/4531fe960a7ab0134f04873bc03fe7bd535c62e4)

PR #367, opened 2026-07-11
fix: add cuDNN-only build option that skips the SageAttention-3 shim

### LINKED ISSUE #334: Exclude unused kernels from compilation, when cuDNN backend is selected?
Some kernels are very heavy (seem to require >64GB of memory), and they may not be used if some manifest selects cuDNN backend.

Could we excluding the Sage Blackwell source from the build, if cuDNN is selected?

Excuses if above assumption are wrong, not very familiar with how things work, it could be more complicated:

```
Unfortunately it may not be a quick flag. The entire 558-byte shim is:

#undef PYBIND11_MODULE
#define PYBIND11_MODULE(name, variable) static void native_sage3_blackwell_pybind_stub(...)
#include "sageattention3_blackwell/sageattn3/blackwell/api.cu"   // ← the 62 GB compile

That one #include pulls in all of SageAttention‑3's Blackwell api.cu (the full CUTLASS sm_120 kernel library) and instantiates it. 
That's your 62 GB cudafe++.

```

mvanhorn:
## Summary
Adds a build-time option to produce a cuDNN-only OmniDreams single-view native extension that skips the SageAttention-3 Blackwell shim. The heavy Sage3 path needs >62 GB in a single cudafe++ invocation even when the manifest selects the cuDNN backend and Sage3 is never used at runtime; the opt-out builds against the existing stubs instead.

## Why this matters
Reported in #334. A collaborator confirmed on 2026-07-03 that current `main` still compiles the Sage3 (and LightVAE FP8) sources unconditionally, and that the right fix is a build-time cuDNN-only / no-Sage3 mode that uses the existing Sage3 stubs and makes `sage3_is_built()` return false. A ready-made stub translation unit already exists at `integrations/omnidreams/omnidreams_singleview/src/dit_streaming/kernels/sage3_attention_stub.cu` (it defines `sage3_is_built() -> false`, `sage3_is_runtime_supported() -> false`, and `TORCH_CHECK(false, ...)` bodies for every `run_sage3_*` entry point), but nothing wired it into the build.

## Changes
- Adds a build flag that swaps `sage3_blackwell_api_shim.cu` for the existing `sage3_attention_stub.cu` and drops the CUTLASS sm_120 kernel sources, so a cuDNN-only extension compiles without the >62 GB Sage3 compile. Default behavior (Sage3 built) is unchanged; the flag is opt-in.

## Testing
Built the extension with the opt-out flag set and confirmed the Sage3 shim and CUTLASS sm_120 sources are excluded, `sage3_is_built()` returns false, and the cuDNN attention path runs. Default build still compiles Sage3.

Fixes #334

### DIFF
--- a/integrations/omnidreams/omnidreams/native/omnidreams_singleview.py
+++ b/integrations/omnidreams/omnidreams/native/omnidreams_singleview.py
@@ -53,6 +53,7 @@ _VAE_STREAMING_DIR = _SOURCE_DIR / "vae_streaming"
 _PYTORCH_MAX_JOBS_ENV = "MAX_JOBS"
 _DEFAULT_MAX_JOBS_CAP = 8
 _NATIVE_CUDA_ARCH_LIST_ENV = "OMNIDREAMS_SINGLEVIEW_CUDA_ARCH_LIST"
+_DISABLE_SAGE3_ENV = "OMNIDREAMS_SINGLEVIEW_DISABLE_SAGE3"
 _PYTORCH_CUDA_ARCH_LIST_ENV = "TORCH_CUDA_ARCH_LIST"
 _DEFAULT_CUDA_ARCH_LIST = "12.0a"
 
@@ -365,7 +366,12 @@ def _file_sha256(path: Path) -> str:
     return digest.hexdigest()
 
 
+def _sage3_disabled() -> bool:
+    return os.environ.get(_DISABLE_SAGE3_ENV, "").strip().lower() in {"1", "true"}
+
+
 def _extension_sources() -> list[Path]:
+    disable_sage3 = _sage3_disabled()
     return [
         _EXTENSION_SOURCE,
         _NATIVE_PRIMITIVES_SOURCE,
@@ -378,8 +384,14 @@ def _extension_sources() -> list[Path]:
         _VAE_STREAMING_DIR / "lightvae_fp8_warp_mma_stages.cu",
         _VAE_STREAMING_DIR / "lightvae_fp8_attention.cu",
         _DIT_STREAMING_PYEXT_DIR / "streaming_dit_bridge.cu",
-        _DIT_STREAMING_PYEXT_DIR / "sage3_blackwell_api_shim.cu",
-        _DIT_STREAMING_PYEXT_DIR / "sage3_fp4_quant_shim.cu",
+        *(
+            []
+            if disable_sage3
+            else [
+                _DIT_STREAMING_PYEXT_DIR / "sage3_blackwell_api_shim.cu",
+                _DIT_STREAMING_PYEXT_DIR / "sage3_fp4_quant_shim.cu",
+            ]
+        ),
         _DIT_STREAMING_KERNEL_DIR / "attention.cu",
         _DIT_STREAMING_KERNEL_DIR / "block_quant.cu",
         _DIT_STREAMING_KERNEL_DIR / "cosmos_adaln_lora.cu",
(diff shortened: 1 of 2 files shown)


jarcherNV:
Thanks, this looks like the right direction overall. The source selection, compile flags, extension name isolation, and Sage3 availability checks all make sense to me.

I think we need one cache fix before merging, though. `load_extension()` still returns the first loaded module from the module-level `_extension` cache before checking the current `OMNIDREAMS_SINGLEVIEW_DISABLE_SAGE3` mode. That means a long-running process that loads the Sage3 build first, then sets `OMNIDREAMS_SINGLEVIEW_DISABLE_SAGE3=1`, will keep using the Sage3-enabled binary. The reverse case has the same problem.

Could you update the cache to be keyed by the extension name/build mode, or store the loaded extension’s cache key and only return it when it matches the current mode? It would also be good to add a regression test that loads once in one mode, toggles the env var, calls `load_extension()` again, and verifies the second mode is loaded separately.

I think once that's done along with a rebase, we can trigger CI and attempt to merge.

mvanhorn:
Good catch — fixed in e1bf2e0. The `_extension` cache is now keyed on the resolved `OMNIDREAMS_SINGLEVIEW_DISABLE_SAGE3` mode, so toggling it at runtime loads the correct binary instead of returning the first-loaded module. Added a test covering both modes.

jarcherNV (approved):
Sorry, somehow this PR slipped off my radar. This change LGTM now.

gtong-nv:
Hi @mvanhorn , looks there is a CI test failure - 
```
tmp_path = PosixPath('/tmp/pytest-of-root/pytest-0/test_load_extension_caches_sep0')
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x75dd94f81370>

    @pytest.mark.ci_cpu
    def test_load_extension_caches_separate_sage3_modes(
        tmp_path: Path,
        monkeypatch: pytest.MonkeyPatch,
    ) -> None:
        import torch.utils.cpp_extension as cpp_extension
    
        extensions: list[ModuleType] = []
    
        def fake_load_torch_extension(**_: object) -> ModuleType:
            extension = _fake_extension_module()
            extensions.append(extension)
            return extension
    
        monkeypatch.setattr(native, "_extension", {})
        monkeypatch.setattr(native, "_extension_load_error", None)
        monkeypatch.setattr(
            native, "validate_thirdparty", lambda: _fake_thirdparty_info(tmp_path)
        )
        monkeypatch.setattr(cpp_extension, "load", fake_load_torch_extension)
        monkeypatch.setattr(native, "_python_package_dir", lambda package: None)
        monkeypatch.delenv("OMNIDREAMS_SINGLEVIEW_DISABLE_SAGE3", raising=False)
    
        sage3_extension = native.load_extension(build_root=tmp_path / "native-build")
        monkeypatch.setenv("OMNIDREAMS_SINGLEVIEW_DISABLE_SAGE3", "1")
        stub_extension = native.load_extension(build_root=tmp_path / "native-build")
    
        assert sage3_extension is extensions[0]
>       assert stub_extension is extensions[1]
                                 ^^^^^^^^^^^^^
E       IndexError: list index out of range

integrations/omnidreams/tests/test_omnidreams_singleview_native.py:388: IndexError
```

Can you address it? We can merge it once it's fixed

mvanhorn:
Good catch, thanks. That test was failing on its own bug, not on the extension change.

`_fake_thirdparty_info` does `mkdir(parents=True)` with no `exist_ok`, and it was being called from inside the `validate_thirdparty` lambda, so the second lookup raised `FileExistsError`. `load_extension` catches that, records it in `_extension_load_error` and returns `None`, so the stub load never reached the fake builder and `extensions[1]` was missing, which is the `IndexError` you saw.

Now built once and closed over, same shape as `test_extension_name_isolated_by_sage3_build_opt_out`. Locally the test fails before the change and passes after.

mvanhorn:
Thank you for merging the cuDNN-only build option, @jarcherNV. Should make builds without SageAttention much easier.

PR #371, opened 2026-07-13
fix(webrtc): reap checkpoint download workers

gtong-nv:
## Summary

- Prevent WebRTC shutdown from leaving checkpoint-download worker processes running after `Ctrl+C`.
- Terminate and reap multiprocessing children owned by each WebRTC server rank during shutdown.
- Escalate workers that do not exit within five seconds from `terminate()` to `kill()`.
- Add a CPU regression test covering interruption while a checkpoint-download-style child process is active.

## Problem

WebRTC startup preloads model checkpoints before the server begins accepting sessions. Sharded Hugging Face checkpoints are downloaded using multiprocessing workers. If the server is interrupted during this startup phase, the server rank can exit while those workers continue downloading as orphan processes re-parented to PID 1.

## Implementation

The shared WebRTC bootstrap now performs child-process cleanup in its shutdown `finally` path for both the serving rank and distributed worker ranks. Cleanup:

1. Finds active multiprocessing children owned by the current rank.
2. Sends each child a termination request.
3. Waits up to five seconds for the children to exit and reaps them.
4. Kills and reaps any remaining children.

Rank 0 still sends the distributed exit signal even if child-process cleanup raises an error.

## Testing

- `pytest flashdreams/tests/test_webrtc_serving.py flashdreams/tests/test_webrtc_manager.py flashdreams/tests/test_webrtc_bootstrap.py -q` — 10 passed
- Ruff check and format verification for the changed files — passed
- Targeted `ty` check for the changed files — passed

The repository-wide pre-commit type hook still reports the existing unresolved optional `flip_evaluator` import in `flashdreams/flashdreams/quality/clip_compare.py`; this is unrelated to this change.

### DIFF
--- a/flashdreams/flashdreams/serving/webrtc/bootstrap.py
+++ b/flashdreams/flashdreams/serving/webrtc/bootstrap.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 
 import gc
 import logging
+import multiprocessing
+import time
 from typing import Protocol
 
 import torch
@@ -15,6 +17,8 @@ from aiohttp import web
 from loguru import logger
 
 from flashdreams.core.distributed import configure_loguru_for_distributed
+
+_CHILD_PROCESS_TERMINATION_TIMEOUT_S = 5.0
 
 
 class WebRTCServerLifecycle(Protocol):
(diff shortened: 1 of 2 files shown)


gtong-nv:
/ok to test [5c22865](https://github.com/NVIDIA/flashdreams/pull/371/commits/5c228656846979325d6bb4b6c1d592197ddd0227)

PR #376, opened 2026-07-14
Add FlashDreams performance workflow skills

jarcherNV:
Create reusable agent skills for profiling model performance, applying
inference optimizations, and validating benchmark/quality results. Update the
agent skill map so future agents can discover the new performance workflow
guidance.


gtong-nv (approved):
LGTM!

PR #377, opened 2026-07-14
fix(lingbot): Route LingBot v2 examples to the v2 repository

gtong-nv:
# Route LingBot v2 examples to the v2 repository

## Summary

- Associate each LingBot runner preset with its matching upstream example-data
  repository.
- Download v2 example assets from `Robbyant/lingbot-world-v2` instead of
  `Robbyant/lingbot-world`.
- Store v1 and v2 examples in separate cache directories so assets at the same
  example index cannot overwrite or mask one another.
- Apply the version-specific source to both offline runners and the WebRTC
  runtime, including its remote fallback URLs.
- Document the new download and cache behavior and add regression coverage for
  every shipped LingBot slug.

## Motivation

LingBot v1 and v2 can provide different prompts for the same numbered example.
Previously, all LingBot slugs downloaded examples from the v1 repository and
shared the same `lingbot_world/<NN>` cache directory. A v2 run could therefore
receive the v1 prompt for an otherwise matching example index, including when a
stale v1 asset was already cached.

This change treats the example-data repository as part of the runner preset and
separates the caches by model version. The selected model now consistently uses
the prompt and assets published for that version.

## Behavior

| Runner presets | Upstream examples | Local cache |
| --- | --- | --- |
| `lingbot-world-fast` and `lingbot-world-fast-taehv-window15-sink3` | `Robbyant/lingbot-world` | `$FLASHDREAMS_CACHE_DIR/example_data/lingbot_world/<NN>/` |
| `lingbot-world-v2-14b-causal-fast` and `lingbot-world-v2-14b-causal-fast-taehv-window15-sink3` | `Robbyant/lingbot-world-v2` | `$FLASHDREAMS_CACHE_DIR/example_data/lingbot_world_v2/<NN>/` |

The v1 cache path remains unchanged for backward compatibility. Existing v1
downloads are not moved or invalidated.

## Testing

```bash
.venv/bin/pytest -q \
  integrations/lingbot/tests/test_smoke.py \
  integrations/lingbot/tests/test_distributed_server_main.py
```

Result: `26 passed`.

The changed files also pass the repository pre-commit hooks, including Ruff
formatting/lint and the `ty` type check.

### DIFF
--- a/integrations/lingbot/README.md
+++ b/integrations/lingbot/README.md
@@ -185,12 +185,16 @@ Then open:
   HuggingFace on first run and cached under `$HF_HOME`.
 - ~200 GB free disk for the model + HF cache.
 - Example assets (`image.jpg`, `intrinsics.npy`, `poses.npy`, and a prompt when
-  available) auto-download from the upstream
+  available) for v1 presets auto-download from the upstream
   [`Robbyant/lingbot-world`](https://github.com/Robbyant/lingbot-world/tree/main/examples)
-  examples folder into `assets/example_data/lingbot_world/<NN>/` on
-  first launch (`<NN>` is the `--example-idx`: `00` through `05`). Examples
-  `03` and `04` use an empty prompt because they do not provide their own
-  upstream `prompt.txt`.
+  examples folder into
+  `$FLASHDREAMS_CACHE_DIR/example_data/lingbot_world/<NN>/`. A v2 preset instead
+  downloads from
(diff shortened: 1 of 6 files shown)


gtong-nv:
/ok to test [1f57f95](https://github.com/NVIDIA/flashdreams/pull/377/commits/1f57f954067c1613e5e704329a30e247a9faadde)

liruilong940607 (approved):

PR #382, opened 2026-07-17
[Interactive Drive] Annotate nvtx and Optimize

ArielG-NV:
Changes:

Reduce BEV waiting by delaying BEV by 1 chunk. Visually this looks better since it looks like GoogleMaps in a car (slightly behind real life, floating feeling). This is better for perf since we don't need to wait/worry on our BEV, we had a whole 8 frames to build the BEV+HUD

Delete things in a different order due to nuance in how they cause syncing (overlap)

Remove some unneeded cuda-stream passing (which introduced long cuda sync points which don't do anything)

Remove the python-GIL on code that does not need it

Add NVTX annotations & nsys guide so that we have an easier time in exploring profiler driven optimizations in the future


ArielG-NV:
Will make a new PR that is cleaned up and does not include nvtx/profiler (auxiliary changes)

PR #383, opened 2026-07-17
Standardize workspace on headless OpenCV

### LINKED ISSUE #363: Standardize workspace on opencv-python-headless to prevent missing cv2 installs
## Summary

The workspace resolves both `opencv-python` and `opencv-python-headless` into the same shared virtual environment. These are separate Python distributions but install the same top-level `site-packages/cv2/` files. Installing or removing one variant can therefore delete files owned by the other while leaving its `.dist-info` metadata behind.

The resulting environment appears valid to `uv pip list`, but `import cv2` fails:

```text
ModuleNotFoundError: No module named 'cv2'
```

This was encountered while launching the OmniDreams WebRTC server from PR #359.

## Current dependency sources

OmniDreams and the core package use the headless distribution:

- `integrations/omnidreams/pyproject.toml`: `opencv-python-headless>=4.5`
- `flashdreams/pyproject.toml`: `opencv-python-headless>=4.5`

The workspace also introduces the GUI distribution:

- `integrations/flashvsr/pyproject.toml`: `opencv-python>=4.8`
- `integrations/omnidreams/ludus-renderer/pyproject.toml` dev extra: `opencv-python>=4.5`

The root `uv.lock` consequently contains both distributions.

A source scan found no `cv2` imports in FlashVSR. Ludus Renderer uses `cv2.VideoCapture` in an example, which works with the headless wheel and does not require GUI functionality.

## Reproduction

From the workspace root on Windows:

```bash
uv sync --all-packages --all-extras
uv pip list | grep opencv
```

The shared environment can contain both:

```text
opencv-python
opencv-python-headless
```

After an environment transition that removes or replaces one variant, for example syncing a narrower workspace package:

```bash
uv sync --package flashdreams-omnidreams --extra dev
uv pip list | grep opencv
uv run --package flashdreams-omnidreams python -c "import importlib.util; print(importlib.util.find_spec('cv2'))"
uv run --package flashdreams-omnidreams python -c "import cv2"
```

Observed state:

```text
opencv-python-headless 4.13.0.92
None
ModuleNotFoundError: No module named 'cv2'
```

The headless distribution's `.dist-info` directory and `RECORD` remained, but the actual `site-packages/cv2/` directory was absent. Because the requested version was still recorded as installed, a normal `uv run` did not reinstall it.

The environment can be repaired manually with:

```bash
uv sync \
  --package flashdreams-omnidreams \
  --extra dev \
  --reinstall-package opencv-python-headless
```

That workaround restores `cv2`, but it does not remove the workspace-level collision.

## Proposed fix

1. Remove the unused `opencv-python>=4.8` dependency from `integrations/flashvsr/pyproject.toml`. FlashVSR currently has no direct `cv2` imports.
2. Change the Ludus Renderer dev extra from `opencv-python>=4.5` to `opencv-python-headless>=4.5`. Its observed `VideoCapture` use does not require GUI support.
3. Regenerate the root lockfile with `uv lock` and verify the root workspace no longer resolves `opencv-python`.
4. Add a `ci_cpu` dependency-policy test that scans workspace-member `pyproject.toml` files and rejects GUI/contrib OpenCV distributions:
   - `opencv-python`
   - `opencv-contrib-python`
   - `opencv-contrib-python-headless`

The shared workspace should use only `opencv-python-headless`.

## Acceptance criteria

- Root `uv.lock` contains `opencv-python-headless` but not `opencv-python`.
- `uv sync --all-packages --all-extras` installs only one OpenCV distribution.
- Switching between full-workspace and OmniDreams-only syncs does not break `import cv2`.
- `uv run --package flashdreams-omnidreams python -c "import cv2"` succeeds.
- A CPU test prevents workspace members from reintroducing another OpenCV wheel variant.

## OSS/dependency notes

This proposal does not add a new dependency. `opencv-python-headless` is already an approved direct dependency and already appears in `THIRD-PARTY-NOTICES`. The change removes `opencv-python` from the root workspace resolution and updates `uv.lock`; no new OSRB dependency approval should be required.

Arison591:
## Summary

- remove FlashVSR's unused direct dependency on the GUI-enabled `opencv-python` distribution
- switch ludus-renderer's development extra to `opencv-python-headless`
- regenerate both the workspace and standalone ludus-renderer lockfiles
- add a CPU dependency-policy test that prevents workspace packages from reintroducing conflicting OpenCV distributions

## Why

The `opencv-python` and `opencv-python-headless` distributions install the same `cv2` namespace and should not coexist in one environment. FlashDreams already standardizes on the headless distribution for server and CI use, but FlashVSR and ludus-renderer's development extra could still bring the GUI distribution into the shared workspace.

FlashVSR does not import `cv2` directly, so its extra direct dependency is unnecessary. The renderer's development dependency still provides the same Python API through the headless build.

This does not add a new third-party component: `opencv-python-headless` is already used throughout the workspace and recorded in `THIRD-PARTY-NOTICES`.

## Validation

- `PYTHONPATH=flashdreams python -m pytest tests/test_dependency_policy.py -m ci_cpu -q` — 1 passed
- `uv lock --check` — passed
- `uv lock --check --project integrations/omnidreams/ludus-renderer` — passed
- `ruff check tests/test_dependency_policy.py` — passed
- `ruff format --check tests/test_dependency_policy.py` — passed
- `uvx ty check tests/test_dependency_policy.py` — passed
- `uvx --from reuse reuse lint` — REUSE 3.3 compliant, 508/508 files covered

GPU inference was not run because this change only affects dependency metadata and its CPU policy test.

The commit is signed off under the Developer Certificate of Origin.

### DIFF
--- a/integrations/flashvsr/pyproject.toml
+++ b/integrations/flashvsr/pyproject.toml
@@ -27,7 +27,6 @@ dependencies = [
     "flashdreams",
     "grpcio>=1.80",
     "mediapy>=1.1",
-    "opencv-python>=4.8",
     "Pillow>=10",
     "protobuf>=6.31.1",
 ]
--- a/integrations/omnidreams/ludus-renderer/pyproject.toml
+++ b/integrations/omnidreams/ludus-renderer/pyproject.toml
@@ -45,7 +45,7 @@ video-codec = [
 dev = [
     "pytest>=8.0",
     "imageio-ffmpeg>=0.5",
-    "opencv-python>=4.5",
+    "opencv-python-headless>=4.5",
 ]
 
 [build-system]
--- a/integrations/omnidreams/ludus-renderer/pyproject.toml
+++ b/integrations/omnidreams/ludus-renderer/pyproject.toml
@@ -45,7 +45,7 @@ video-codec = [
 dev = [
     "pytest>=8.0",
     "imageio-ffmpeg>=0.5",
-    "opencv-python-headless>=4.5",
+    "opencv-python-headless>=4.5,<5",
 ]
 
 [build-system]
--- a/integrations/omnidreams/ludus-renderer/uv.lock
+++ b/integrations/omnidreams/ludus-renderer/uv.lock
@@ -28,7 +28,7 @@ name = "cuda-bindings"
 version = "13.3.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "cuda-pathfinder", marker = "(python_full_version < '3.11' and sys_platform == 'emscripten') or (python_full_version < '3.11' and sys_platform == 'win32') or (sys_platform != 'emscripten' and sys_platform != 'win32')" },
+    { name = "cuda-pathfinder" },
 ]
 wheels = [
     { url = "https://files.pythonhosted.org/packages/b6/f7/a4a9491d73ea5622da0e66d3a2bd581b2b6e7456a3faa871835f47155616/cuda_bindings-13.3.0-cp310-cp310-manylinux_2_24_aarch64.manylinux_2_28_aarch64.whl", hash = "sha256:326b31d88314068f0ead48141222d7ada094f4013723dd924af9db10150fa472", size = 6071925, upload-time = "2026-05-27T03:58:55.336Z" },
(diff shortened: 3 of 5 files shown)


gtong-nv:
Hi @Arison591, thanks for the PR. Can you mark it "Ready for review" so that we can land it?
This PR should fix https://github.com/NVIDIA/flashdreams/issues/363

gtong-nv:
/ok to test [55980b2](https://github.com/NVIDIA/flashdreams/pull/383/commits/55980b2df76a11074f3d244b95e18084d095f95d)

jarcherNV:
Thanks, this looks directionally good to me. Standardizing on `opencv-python-headless` makes sense, and the workspace CI passing gives good coverage for the main FlashDreams path.

One thing I think we should address before merge: the standalone `integrations/omnidreams/ludus-renderer/uv.lock` now resolves `opencv-python-headless` to `5.0.0.93`, while the workspace lock remains on `4.13.0.92`. Since the PR is intended to switch GUI OpenCV to headless rather than introduce an OpenCV major-version upgrade, could we either pin the standalone ludus-renderer dev extra to `<5` and regenerate the lockfile, or run/record a standalone ludus-renderer smoke/test pass against the OpenCV 5 lock?

Minor follow-up: the new dependency policy test currently scans workspace member `pyproject.toml` files, but not the root `pyproject.toml`, where shared dependency groups live. It would be useful to include the root file too so future root-level dependency additions cannot reintroduce `opencv-python`.

jarcherNV (approved):
LGTM

PR #384, opened 2026-07-18
Extract shared runner I/O utilities

jarcherNV:
Extract shared runner I/O utilities

Add flashdreams.infra.runner_io for common prompt resolution, output paths,
stats JSON, image/video loading, normalization, and MP4 writing with lazy
cv2/mediapy imports.

Migrate matching public runners to use the shared helpers while preserving
model-specific download, crop, camera, and writer behavior. Add CPU tests for
the new helper contracts.

### DIFF
--- a/flashdreams/flashdreams/infra/runner_io.py
+++ b/flashdreams/flashdreams/infra/runner_io.py
@@ -1,0 +1,296 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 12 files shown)


flashdreams/flashdreams/infra/runner_io.py
```diff
@@ -0,0 +1,353 @@
...
+    path: str | Path,
+    *,
+    fps: int | float,
+    layout: VideoTensorLayout,
+    install_hint: str = DEFAULT_RUNNER_INSTALL_HINT,
+) -> Path:
+    """Write a ``[-1, 1]`` video tensor as an MP4 and return the path."""
+    media = _import_mediapy("Writing videos", install_hint=install_hint)
+    path = Path(path)
+    media.write_video(str(path), video_tensor_to_uint8(video, layout=layout), fps=fps)
```
gtong-nv, line 320: Not for this PR, but we should provide a `stream_video_tensor_to_file` function to avoid OOM for long video output by appending the tensor to the output during the rollout, rather than dumping in the end at once.

integrations/causal_forcing/causal_forcing/runner.py
```diff
@@ -17,21 +17,26 @@
...
 from flashdreams.infra.runner import Runner, RunnerConfig
+from flashdreams.infra.runner_io import (
+    ensure_output_dir,
+    load_first_frame_tensor,
+    read_image_rgb,
+    resolve_input_path,
+    resolve_prompt_value,
+    runner_artifact_path,
+    write_runner_stats,
+    write_video_tensor,
```
gtong-nv, line 38: nit: instead of importing all the functions, maybe better organize them in a single class?

gtong-nv (approved):
LGTM. 
Left a few comments but OK to address in the future.

jarcherNV:
Thanks for the comments. I have added them to my TODO list.

PR #385, opened 2026-07-19
Extract realtime input and media primitives

jarcherNV:
Add a transport-neutral serving.realtime package for keyboard/input state, latest-frame publishing, tensor/array RGB conversion, and lazy JPEG encoding. Refactor WebRTC controls and media to consume the shared primitives while preserving existing import compatibility and BufferedVideoTrack behavior.

Add CPU tests covering input snapshots, frame-bus shutdown semantics, media conversion parity, and JPEG encoding.

### DIFF
--- a/flashdreams/flashdreams/serving/realtime/media.py
+++ b/flashdreams/flashdreams/serving/realtime/media.py
@@ -38,6 +38,8 @@ def _as_numpy(value: object, *, sync_device: bool) -> np.ndarray:
                 "host/device synchronization is explicit at the call site."
             )
         tensor = tensor.cpu()
+    if tensor.is_floating_point():
+        tensor = tensor.float()
     return tensor.numpy()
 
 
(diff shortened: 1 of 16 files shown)


flashdreams/flashdreams/serving/realtime/media.py
```diff
@@ -0,0 +1,189 @@
...
+    image_module = importlib.import_module("PIL.Image")
+    image_from_array = getattr(image_module, "fromarray")
+    rgb_uint8 = rgb_frame_to_uint8(
+        frame,
+        layout=layout,
+        value_range=value_range,
+        sync_device=sync_device,
+    )
+    output = io.BytesIO()
+    image_from_array(rgb_uint8).save(output, format="JPEG", quality=quality)
```
gtong-nv, line 188: maybe we can add the nvjpeg path here if that HW unit is available ?

gtong-nv (approved):
LGTM!

flashdreams/flashdreams/serving/realtime/media.py
jarcherNV: Good idea, I will add this to my TODO list.

PR #387, opened 2026-07-19
Extract shared realtime timing helpers

jarcherNV:
Extract shared realtime timing helpers

Move the OmniDreams realtime timing records, trace context, video-model
stage tracing, and input-to-present profile window into
flashdreams.serving.realtime.timing. Add shared duration and rolling
summary helpers, optional decode/cache stage timings, and migrate the
OmniDreams interactive demo and tests to consume the shared module while
keeping the old import path as a compatibility shim.

### DIFF
--- a/flashdreams/flashdreams/serving/realtime/timing.py
+++ b/flashdreams/flashdreams/serving/realtime/timing.py
@@ -1,0 +1,543 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Transport-neutral realtime timing records and trace helpers."""
+
+from __future__ import annotations
+
+from collections import defaultdict, deque
+from collections.abc import Iterable, Mapping
+from dataclasses import dataclass
+from threading import Lock
+from typing import Protocol
+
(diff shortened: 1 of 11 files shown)


gtong-nv:
I think we might need to revisit this profiling logic later. Currently only the omnidream-interactive demo uses this logic. Technically other interactive demo can re-use it, but the interface is not very clear.

gtong-nv (approved):
LGTM

jarcherNV:
> I think we might need to revisit this profiling logic later. Currently only the omnidream-interactive demo uses this logic. Technically other interactive demo can re-use it, but the interface is not very clear.

Yeah I imagine @ArielG-NV may want to update our timing metrics as part of his benchmark work this week, so this one may go through a bunch of updates.

ArielG-NV:
I am unsure on if I would want to use the current tooling here for timing or to eventually replace it all. Unlikely I use it in its current state though.

Eventually, either we will add to what we have or replace it so that any timing metrics we collect give us an annotated [nvtx](https://nvtx.readthedocs.io/en/latest/index.html) timeline for agentic development and profiling sanity reasons.

We can still merge the changes though with/without the existing code - if I won't use it, it will just be code I strip out naturally when I propose the benchmark/regression-test changes.

PR #388, opened 2026-07-19
Extract shared realtime presentation helpers

jarcherNV:
Extract shared realtime presentation helpers

Add transport-neutral presentation queue, pacing, frame materialization, and
MJPEG multipart helpers under flashdreams.serving.realtime. Migrate the
OmniDreams interactive runtime loop and MJPEG presenter to use the shared
queue/pacing and latest-frame MJPEG utilities while keeping demo-specific
routes, controls, overlays, and scene selection local.

### DIFF
--- a/flashdreams/flashdreams/serving/realtime/__init__.py
+++ b/flashdreams/flashdreams/serving/realtime/__init__.py
@@ -1,0 +1,4 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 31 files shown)


gtong-nv:
This is also used only the omnidreams interactive demo. Do we really need it to be shared with others? 
I think we should default to use the webRTC path for streaming?

jarcherNV:
> This is also used only the omnidreams interactive demo. Do we really need it to be shared with others? I think we should default to use the webRTC path for streaming?

Yeah, I think webrtc should be the default. What do you think? Should we not worry about refactoring this for potential future demos to use?

jarcherNV:
I am abandoning this change as we have decided to remove the interactive drive demo, and will default to using the webrtc path for all demos going forward.

PR #393, opened 2026-07-21
 Stabilize Omnidreams WebRTC Manifest Demo

gtong-nv:
# Stabilize Omnidreams WebRTC Manifest Demo

Before:
Browser can only present at ~20 FPS even the server can produces 30 FPS steady.

After:
Both Browser and server can maintain 30+FPS.

after_30fps

## Summary

Adds manifest-driven Omnidreams WebRTC session setup and stabilizes the interactive browser playback path. The branch also brings in the NVENC H.264 WebRTC encoder path, richer runtime/browser profiling, and fixes manifest override tracking for abbreviated long CLI options.

The WebRTC demo can now prefer hardware H.264 encode when the browser negotiates it, fall back or fail loudly according to the selected encoder mode, and expose enough server/client timing to distinguish model, encode, transport, decode, and compositor bottlenecks.

## Changes

- Add Omnidreams WebRTC manifest support and split shared interactive-drive CLI parsing helpers into `cli_args.py`.
- Track explicit CLI overrides when users pass unambiguous abbreviated long options, preventing manifest defaults from replacing values supplied by the caller.
- Add a `VideoEncoder` abstraction with `DefaultRTCEncoder` and PyNvVideoCodec-backed NVENC H.264 support.
- Wire Omnidreams WebRTC session initialization to select `auto`, `nvenc`, or `default` encoder backends, with `--prefer_sw_encoder` and `--require_nvenc` CLI behavior.
- Preflight SDP video codecs so NVENC is only used when the browser offers H.264; auto mode falls back to the software encoder and required-NVENC mode raises clearly.
- Preserve encoded packet order with backpressured NVENC queue delivery instead of silently dropping packets in the production path.
- Add chunk-level runtime, delivery, encode, queue, lag, and Omnidreams wrapper timing logs.
- Add native extension build/load status logs so first-run compiler activity is visible to users.
- Defer Omnidreams per-chunk finalization so model cache maintenance is attributed separately from the visible generation call.
- Add browser-side WebRTC profiling for receive, decode, drop, jitter, bitrate, presentation, and callback FPS.
- Simplify the live WebRTC UI during playback by hiding nonessential panels, removing expensive blur/shadow effects, throttling metric DOM updates, and reducing log churn.

## Notes

- NVENC is H.264-only in this path; browsers that do not offer H.264 use the software encoder in auto mode.
- The HUD now separates `Media FPS` from `Presented` FPS. `Media FPS` reflects receive/decode throughput; `Presented` reflects browser compositor cadence.
- Extra server-side buffering was not made the default because profiling showed Chrome was already receiving and decoding 30 FPS; the observed slowdown was in presentation/compositing.

### DIFF
--- a/flashdreams/flashdreams/serving/webrtc/encoders.py
+++ b/flashdreams/flashdreams/serving/webrtc/encoders.py
@@ -27,7 +27,7 @@ import time
 from collections.abc import Callable
 from dataclasses import dataclass
 from fractions import Fraction
-from typing import TYPE_CHECKING, Literal, Protocol, runtime_checkable
+from typing import TYPE_CHECKING, Any, Literal, Protocol, runtime_checkable
 
 import torch
 from aiortc import MediaStreamTrack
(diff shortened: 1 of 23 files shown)


gtong-nv:
Close this and the use the cleaned PR  https://github.com/NVIDIA/flashdreams/pull/409

PR #394, opened 2026-07-21
Make WebRTC serving reusable across demos

jarcherNV:
Add shared WebRTC runtime and data-channel message contracts, plus common packaged-app and CUDA/distributed bootstrap helpers. Migrate the LingBot and OmniDreams WebRTC servers onto the shared helpers while keeping their model-specific routes, runtime config, and preflight logic local.


gtong-nv (approved):
LGTM

PR #397, opened 2026-07-22
Support split WebRTC timing and THWC frame chunks

jarcherNV:
Add shared WebRTC runtime hooks for integrations whose input sampling cadence differs from their output video cadence. The session manager now uses optional input FPS, input frame count, and steady output frame count hooks while keeping the legacy runtime methods as defaults.

Add periodic WebRTC performance logging, support THWC uint8 realtime frame chunks, and relax block KV cache rolling so cache sizes do not need to be divisible by chunk size.

Cover the shared behavior with focused realtime, WebRTC manager, and KV cache tests.


gtong-nv (approved):
LGTM

PR #399, opened 2026-07-29
Add local benchmark harness with quality reports

jarcherNV:
Add a local-first benchmark CLI for selected FlashDreams runner/demo scenarios that captures MP4 outputs, logs, structured metrics, environment metadata, and a static HTML report.

Add shipped LingBot and Omnidreams benchmark scenario configs for deterministic 30-second quality comparisons plus one-minute performance and manual-review clips.

Support baseline/candidate MP4 comparison quality metrics for short seeded clips, and optional PAI-Bench-Long scoring for one-minute review scenarios only.

Add an aarch64-friendly local PAI-Bench adapter with dependency preflight diagnostics, staged-video cleanup, score parsing, and report integration.

Document the standard local benchmark commands and cover the harness behavior with focused tests.


ArielG-NV (approved):
I suppose I will leave all comments that we will address in future revisions here (since these are likely to happen soon but better to have benchmarks sooner rather than later):
* Need a clear guide where at the top is 1 clear command to run/execute a config with a demo/benchmark-tool. By clear I mean concise and straight to the point (something useful for a developer), not verbose.
* Need a clear doc for "how to create a config" and a doc for the `schema` of our configs. By clear I mean concise, not verbose.
* Probably want YAML instead of JSON for configs in the future since JSON's are hard to modify and wasteful (`"` everywhere, organization is not enforced via indentation)
* Current config design is a "hack" since we require setting explicit `command` without internal system knowledge of parameters to tune (cannot fuzz parameters programmatically)
* Current config system assumes that model+checkpoint+model_config are going to be tightly integrated together

jarcherNV:
> * Need a clear guide where at the top is 1 clear command to run/execute a config with a demo/benchmark-tool. By clear I mean concise and straight to the point (something useful for a developer), not verbose.
> * Need a clear doc for "how to create a config" and a doc for the `schema` of our configs. By clear I mean concise, not verbose.

For these, the guide is in docs/source/developer_guides/local_benchmarks.rst, there are two commands given which you can directly copy-paste, one is for creating the baseline, the other for the comparison. There are instructions included for creating a config included.

> * Probably want YAML instead of JSON for configs in the future since JSON's are hard to modify and wasteful (`"` everywhere, organization is not enforced via indentation)

We could change this to yaml if people prefer, I don't mind too much.

> * Current config design is a "hack" since we require setting explicit `command` without internal system knowledge of parameters to tune (cannot fuzz parameters programmatically)
> * Current config system assumes that model+checkpoint+model_config are going to be tightly integrated together

I guess maybe that's true? Basically the config just uses whatever can be run from the command line. We can update it when we have an API.

ArielG-NV:
> For these, the guide is in docs/source/developer_guides/local_benchmarks.rst, there are two commands given which you can directly copy-paste, one is for creating the baseline, the other for the comparison. There are instructions included for creating a config included.

Ah, I see what you mean (I skipped over the first section when reading, apologies!).

For the schema I understand your point-- I suppose the config is quite "simple" minus choosing a `command` to run portion. No separate doc is needed to describe this in its current state.

jarcherNV:
Ok, I will merge this now so we can switch to the API work. We can update this to use yaml later if that is preferred.

PR #403, opened 2026-07-31
Add experimental inference runtime API envelope

jarcherNV:
Define the initial flashdreams.runtime package with minimal T1 boundaries for runtime config, user/model inputs, schemas, input mapping, model adapters, runtime/session protocols, output targets, and metrics. Add focused CPU tests for the new API surface without migrating existing runners.

### DIFF
--- a/flashdreams/flashdreams/runtime/config.py
+++ b/flashdreams/flashdreams/runtime/config.py
@@ -5,9 +5,12 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Literal, Mapping
+from typing import Any, Literal
+
+from flashdreams.runtime._utils import freeze_mapping
 
 ExecutionBackend = Literal["local", "local-distributed", "external", "hosted"]
 """Execution backend families the v0 envelope leaves room for."""
@@ -21,8 +24,12 @@ class InferenceConfig:
     """Model/runtime execution settings.
 
     Prompts, user controls, browser settings, output paths, and benchmark
-    directories intentionally live outside this object.
+    directories intentionally live outside this object. The typed optimization
+    fields cover common cross-backend knobs; adapter-specific choices belong in
+    :attr:`runtime_options`.
     """
+
+    __hash__ = None
 
     model_id: str
     """Stable model or adapter identity."""
@@ -63,3 +70,7 @@ class InferenceConfig:
     def __post_init__(self) -> None:
         if not self.model_id.strip():
             raise ValueError("InferenceConfig.model_id must be non-empty.")
+        object.__setattr__(
+            self, "runtime_options", freeze_mapping(self.runtime_options)
+        )
+        object.__setattr__(self, "resource_hints", freeze_mapping(self.resource_hints))
--- a/flashdreams/flashdreams/runtime/config.py
+++ b/flashdreams/flashdreams/runtime/config.py
@@ -1,0 +1,65 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Runtime-facing configuration envelope."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Literal, Mapping
+
+ExecutionBackend = Literal["local", "local-distributed", "external", "hosted"]
+"""Execution backend families the v0 envelope leaves room for."""
+
+Precision = Literal["auto", "fp32", "fp16", "bf16"]
+"""Coarse runtime precision choices."""
+
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class InferenceConfig:
+    """Model/runtime execution settings.
+
+    Prompts, user controls, browser settings, output paths, and benchmark
+    directories intentionally live outside this object.
+    """
+
+    model_id: str
+    """Stable model or adapter identity."""
+
+    preset_id: str | None = None
+    """Optional preset identity under :attr:`model_id`."""
+
+    checkpoint: str | Path | None = None
+    """Optional checkpoint or model-asset selector understood by the adapter."""
+
+    backend: ExecutionBackend = "local"
+    """Runtime backend family."""
+
+    device: str | None = None
+    """Optional device selector such as ``cuda`` or ``cuda:0``."""
+
+    precision: Precision = "auto"
+    """Preferred compute precision."""
+
+    compile: bool | None = None
+    """Whether model compilation is requested, disabled, or left to the adapter."""
+
+    cuda_graph: bool | None = None
+    """Whether CUDA graph capture is requested, disabled, or left to the adapter."""
+
+    attention_backend: str | None = None
+    """Optional attention implementation selector."""
+
+    cache_policy: str | None = None
+    """Optional cache policy selector."""
+
+    runtime_options: Mapping[str, Any] = field(default_factory=dict)
+    """Adapter/backend-specific runtime options."""
+
+    resource_hints: Mapping[str, Any] = field(default_factory=dict)
+    """Cheap resource hints for launchers, schedulers, or hosted backends."""
+
+    def __post_init__(self) -> None:
+        if not self.model_id.strip():
+            raise ValueError("InferenceConfig.model_id must be non-empty.")
--- a/flashdreams/flashdreams/runtime/config.py
+++ b/flashdreams/flashdreams/runtime/config.py
@@ -13,7 +13,7 @@ from typing import Any, Literal
 from flashdreams.runtime._utils import freeze_mapping
 
 ExecutionBackend = Literal["local", "local-distributed", "external", "hosted"]
-"""Execution backend families the v0 envelope leaves room for."""
+"""Where and how inference compute is run."""
 
 Precision = Literal["auto", "fp32", "fp16", "bf16"]
 """Coarse runtime precision choices."""
@@ -21,18 +21,18 @@ Precision = Literal["auto", "fp32", "fp16", "bf16"]
 
 @dataclass(frozen=True, kw_only=True, slots=True)
 class InferenceConfig:
-    """Model/runtime execution settings.
+    """Runtime settings that affect model execution.
 
     Prompts, user controls, browser settings, output paths, and benchmark
     directories intentionally live outside this object. The typed optimization
-    fields cover common cross-backend knobs; adapter-specific choices belong in
-    :attr:`runtime_options`.
+    fields cover common cross-backend knobs; open-ended adapter-specific choices
+    can use :attr:`runtime_options`.
     """
 
     __hash__ = None
 
     model_id: str
-    """Stable model or adapter identity."""
+    """Stable identity for the model adapter or runtime integration."""
 
     preset_id: str | None = None
     """Optional preset identity under :attr:`model_id`."""
@@ -41,10 +41,10 @@ class InferenceConfig:
     """Optional checkpoint or model-asset selector understood by the adapter."""
 
     backend: ExecutionBackend = "local"
-    """Runtime backend family."""
+    """Execution placement and backend family for inference compute."""
 
     device: str | None = None
-    """Optional device selector such as ``cuda`` or ``cuda:0``."""
+    """Optional device selector such as ``cuda`` or ``cuda:0``; ``None`` leaves placement to the adapter/backend."""
 
     precision: Precision = "auto"
     """Preferred compute precision."""
@@ -56,10 +56,10 @@ class InferenceConfig:
     """Optional - Whether CUDA graph capture is requested or disabled. `None` means left to the adapter to decide."""
 
     attention_backend: str | None = None
-    """Optional attention implementation selector."""
+    """Optional attention implementation selector; ``None`` leaves the choice to the adapter."""
 
     cache_policy: str | None = None
-    """Optional cache policy selector."""
+    """Optional cache policy selector; ``None`` leaves the choice to the adapter."""
 
     runtime_options: Mapping[str, Any] = field(default_factory=dict)
     """Adapter/backend-specific runtime options."""
--- a/flashdreams/flashdreams/runtime/config.py
+++ b/flashdreams/flashdreams/runtime/config.py
@@ -50,7 +50,7 @@ class InferenceConfig:
     """Preferred compute precision."""
 
     compile: bool | None = None
-    """Whether model compilation is requested, disabled, or left to the adapter."""
+    """Optional - Whether model compilation is requested or disabled. `None` means left to the adapter to decide."""
 
     cuda_graph: bool | None = None
     """Whether CUDA graph capture is requested, disabled, or left to the adapter."""
--- a/flashdreams/flashdreams/runtime/config.py
+++ b/flashdreams/flashdreams/runtime/config.py
@@ -65,7 +65,7 @@ class InferenceConfig:
     """Adapter/backend-specific runtime options."""
 
     resource_hints: Mapping[str, Any] = field(default_factory=dict)
-    """Cheap resource hints for launchers, schedulers, or hosted backends."""
+    """Resource hints for launchers, schedulers, or hosted backends."""
 
     def __post_init__(self) -> None:
         if not self.model_id.strip():
--- a/flashdreams/flashdreams/runtime/config.py
+++ b/flashdreams/flashdreams/runtime/config.py
@@ -53,7 +53,7 @@ class InferenceConfig:
     """Optional - Whether model compilation is requested or disabled. `None` means left to the adapter to decide."""
 
     cuda_graph: bool | None = None
-    """Whether CUDA graph capture is requested, disabled, or left to the adapter."""
+    """Optional - Whether CUDA graph capture is requested or disabled. `None` means left to the adapter to decide."""
 
     attention_backend: str | None = None
     """Optional attention implementation selector."""
--- a/flashdreams/flashdreams/runtime/inputs.py
+++ b/flashdreams/flashdreams/runtime/inputs.py
@@ -59,6 +59,16 @@ class UserInputSchema:
             return True
         return requested.issubset(self.event_kinds)
 
+    def missing_snapshot(self, inputs: "UserInputs") -> tuple[str, ...]:
+        """Return required snapshot fields absent from ``inputs``."""
+        return _missing_required(self.snapshot_fields, inputs.snapshot)
+
+    def require_snapshot(self, inputs: "UserInputs") -> None:
+        """Raise if required snapshot fields are absent."""
+        missing = self.missing_snapshot(inputs)
+        if missing:
+            raise ValueError(f"Missing required user snapshot field(s): {missing}")
+
 
 @dataclass(frozen=True, kw_only=True, slots=True)
 class ModelInputSchema:
@@ -118,7 +128,9 @@ class UserInputs:
         """Return inputs with events filtered to ``time_window``."""
         return UserInputs(
             events=tuple(
-                event for event in self.events if time_window.contains(event.timestamp_s)
+                event
+                for event in self.events
+                if time_window.contains(event.timestamp_s)
             ),
             snapshot=self.snapshot,
             metadata=self.metadata,
@@ -141,4 +153,6 @@ class ModelInputs:
 def _missing_required(
     fields: tuple[InputField, ...], payload: Mapping[str, Any]
 ) -> tuple[str, ...]:
-    return tuple(field.name for field in fields if field.required and field.name not in payload)
+    return tuple(
+        field.name for field in fields if field.required and field.name not in payload
+    )
--- a/flashdreams/flashdreams/runtime/inputs.py
+++ b/flashdreams/flashdreams/runtime/inputs.py
@@ -6,13 +6,16 @@
 from __future__ import annotations
 
 import math
+from collections.abc import Iterable, Mapping
 from dataclasses import dataclass, field
-from typing import Any, Iterable, Mapping
+from typing import Any
+
+from flashdreams.runtime._utils import freeze_mapping
 
 
 @dataclass(frozen=True, kw_only=True, slots=True)
 class TimeWindow:
-    """Half-open time window in canonical seconds."""
+    """Half-open time window in seconds since session start."""
 
     start_s: float
     end_s: float
@@ -101,7 +104,9 @@ class ModelInputSchema:
 
 @dataclass(frozen=True, kw_only=True, slots=True)
 class UserInputEvent:
-    """Timestamped user-facing input event."""
+    """User-facing input event timestamped in seconds since session start."""
+
+    __hash__ = None
 
     timestamp_s: float
     kind: str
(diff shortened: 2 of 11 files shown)


flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class InferenceConfig:
+    """Model/runtime execution settings.
+
+    Prompts, user controls, browser settings, output paths, and benchmark
+    directories intentionally live outside this object. The typed optimization
+    fields cover common cross-backend knobs; adapter-specific choices belong in
+    :attr:`runtime_options`.
+    """
```
ArielG-NV:
```suggestion
    """Model/runtime inference config.

    Controls all knobs that functionally affect the inference of a model: Technique to host inference compute, `model_id` we are running inference on, `checkpoint` of model to load, whether we run CUDA via a CUDA Graph or not, etc...
    """
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+from collections.abc import Mapping
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Literal
+
+from flashdreams.runtime._utils import freeze_mapping
+
+ExecutionBackend = Literal["local", "local-distributed", "external", "hosted"]
+"""Execution backend families the v0 envelope leaves room for."""
```
ArielG-NV: We should define what each "ENUM" literal means.

```suggestion
"""Techniques to host inference compute."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+    """Preferred compute precision."""
+
+    compile: bool | None = None
+    """Whether model compilation is requested, disabled, or left to the adapter."""
+
+    cuda_graph: bool | None = None
+    """Whether CUDA graph capture is requested, disabled, or left to the adapter."""
+
+    attention_backend: str | None = None
+    """Optional attention implementation selector."""
```
ArielG-NV:
```suggestion
    """Optional - attention implementation selector. `None` means unused."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    cache_policy: str | None = None
+    """Optional cache policy selector."""
```
ArielG-NV:
```suggestion
    """Optional - cache policy selector. `None` means unused."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    runtime_options: Mapping[str, Any] = field(default_factory=dict)
+    """Adapter/backend-specific runtime options."""
+
+    resource_hints: Mapping[str, Any] = field(default_factory=dict)
+    """Cheap resource hints for launchers, schedulers, or hosted backends."""
```
ArielG-NV:
```suggestion
    """Resource hints for launchers, schedulers, or hosted backends."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+    Prompts, user controls, browser settings, output paths, and benchmark
+    directories intentionally live outside this object. The typed optimization
+    fields cover common cross-backend knobs; adapter-specific choices belong in
+    :attr:`runtime_options`.
+    """
+
+    __hash__ = None
+
+    model_id: str
+    """Stable model or adapter identity."""
```
ArielG-NV:
```suggestion
    model_id: str
    """Model architecture used to host the selected `checkpoint`."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    preset_id: str | None = None
+    """Optional preset identity under :attr:`model_id`."""
```
ArielG-NV, line 38: Why is this optional? To identify a model given a YAML we need an identity string to refer to

```suggestion
    preset_id: str
    """Unique preset identity"""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    checkpoint: str | Path | None = None
+    """Optional checkpoint or model-asset selector understood by the adapter."""
```
ArielG-NV, line 41: Why is this optional? this would mean we hard-code default checkpoint locations somewhere.
```suggestion
    checkpoint: str | Path
    """Checkpoint understood by the adapter. URL to fetch from or Path to local file. Ensure `env` contains `HF_TOKEN` if fetching from hugging-face."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    backend: ExecutionBackend = "local"
+    """Runtime backend family."""
```
ArielG-NV:
```suggestion
    backend: ExecutionBackend = "local"
    """Technique to host inference compute."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    device: str | None = None
+    """Optional device selector such as ``cuda`` or ``cuda:0``."""
```
ArielG-NV:
```suggestion
    device: str = "cuda"
    """Device selector such as ``cuda`` or ``cuda:0``."""
```

Why does not specifying a device make sense to include?

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    precision: Precision = "auto"
+    """Preferred compute precision."""
+
+    compile: bool | None = None
+    """Whether model compilation is requested, disabled, or left to the adapter."""
```
ArielG-NV:
```suggestion
    compile: bool | None = None
    """Optional - Whether model compilation is requested or disabled. `None` means left to the adapter to decide."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    cuda_graph: bool | None = None
+    """Whether CUDA graph capture is requested, disabled, or left to the adapter."""
```
ArielG-NV:
```suggestion
    cuda_graph: bool | None = None
    """Optional - Whether CUDA graph capture is requested or disabled. `None` means left to the adapter to decide."""
```

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+    cache_policy: str | None = None
+    """Optional cache policy selector."""
+
+    runtime_options: Mapping[str, Any] = field(default_factory=dict)
+    """Adapter/backend-specific runtime options."""
+
+    resource_hints: Mapping[str, Any] = field(default_factory=dict)
+    """Cheap resource hints for launchers, schedulers, or hosted backends."""
+
+    def __post_init__(self) -> None:
```
ArielG-NV, line 70: We likely need more validation on parameter values and checks if types assigned were correct & if elements like `runtime_options` are valid

maybe we stub these checks?

flashdreams/flashdreams/runtime/inputs.py
```diff
@@ -0,0 +1,189 @@
...
+        missing = self.missing_snapshot(inputs)
+        if missing:
+            raise ValueError(f"Missing required user snapshot field(s): {missing}")
+
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class ModelInputSchema:
+    """Minimal metadata for model-facing initial and per-step inputs."""
+
+    initial_fields: tuple[InputField, ...] = ()
```
ArielG-NV, line 80:
```suggestion
    initial_fields: tuple[InputField, ...] = ()
    """Inputs needed to run first generation."""
```

flashdreams/flashdreams/runtime/inputs.py
```diff
@@ -0,0 +1,189 @@
...
+    step_fields: tuple[InputField, ...] = ()
```
ArielG-NV, line 83:
```suggestion
    step_fields: tuple[InputField, ...] = ()
        """Inputs needed to run generation at an arbitrary step after the initial generation."""

```

flashdreams/flashdreams/runtime/inputs.py
```diff
@@ -0,0 +1,189 @@
...
+class UserInputEvent:
+    """User-facing input event timestamped in seconds since session start."""
+
+    __hash__ = None
+
+    timestamp_s: float
+    kind: str
+    payload: Mapping[str, Any] = field(default_factory=dict)
+    source: str | None = None
+    event_id: str | None = None
```
ArielG-NV: why string and not `int`?

also- how do we have no id?

flashdreams/flashdreams/runtime/inputs.py
```diff
@@ -0,0 +1,189 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ArielG-NV, line 1: We should at some point doc-comment these members in the file to better explain purpose (and add a requirement to CI so that our API does not fall apart over time)

ArielG-NV (commented):
We should likely stub `__post_init__` validation of types so that agents don't forget to include the validation.

I commented rather-heavily-- feel free to ignore a suggestion if it seems rather-nonsense

flashdreams/flashdreams/runtime/config.py
fangjunzhou-nv: What're available `attention_backend` exposed here? Is this field going to be used to choose

```
        sdpa_backend = {
            "math": torch.nn.attention.SDPBackend.MATH,
            "efficient": torch.nn.attention.SDPBackend.EFFICIENT_ATTENTION,
            "cudnn": torch.nn.attention.SDPBackend.CUDNN_ATTENTION,
            "flash": torch.nn.attention.SDPBackend.FLASH_ATTENTION,
        }[self.backend]
```

One of the suggestions I have is we remove all these implementation details selector out of `InferenceConfig`, or we create a `ModelConfig` class to hold all the implementation details in a model. For now, we can store one `ModelConfig` per `InferenceConfig`. Later when we start building auto tune system we can store a list of candidate ModelConfig for the auto tuner to pick.

See Triton autotune system for reference: https://triton-lang.org/main/python-api/generated/triton.autotune.html

flashdreams/flashdreams/runtime/config.py
fangjunzhou-nv: What are some available options here? We should make a Enum instead of str.

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Literal
+
+from flashdreams.runtime._utils import freeze_mapping
+
+ExecutionBackend = Literal["local", "local-distributed", "external", "hosted"]
+"""Execution backend families the v0 envelope leaves room for."""
+
+Precision = Literal["auto", "fp32", "fp16", "bf16"]
```
fangjunzhou-nv, line 18:
```suggestion
class Precision(Enum):
  AUTO = "auto"
  FP32 = "fp32"
  FP16 = "fp16"
  BF16 = "bf16"
```

We should use enum.Enum for all enums, this should make LSP happier.

flashdreams/flashdreams/runtime/inputs.py
gtong-nv: do we mean event_type here? otherwise sounds __hash__ should just be the id

flashdreams/flashdreams/runtime/config.py
```diff
@@ -0,0 +1,76 @@
...
+    cuda_graph: bool | None = None
+    """Whether CUDA graph capture is requested, disabled, or left to the adapter."""
+
+    attention_backend: str | None = None
+    """Optional attention implementation selector."""
+
+    cache_policy: str | None = None
+    """Optional cache policy selector."""
+
+    runtime_options: Mapping[str, Any] = field(default_factory=dict)
```
fangjunzhou-nv, line 64: Should we use TypedDict over dict for all dictionary? Again, this should make LSP happier.

flashdreams/flashdreams/runtime/inputs.py
gtong-nv: hmm, but we already have event.kind

flashdreams/flashdreams/runtime/inputs.py
fangjunzhou-nv, line 1: Should we split user input and model input into two source files?

flashdreams/flashdreams/runtime/metrics.py
```diff
@@ -0,0 +1,124 @@
...
+    Timing samples should use seconds as their canonical unit.
+    """
+
+    __hash__ = None
+
+    name: str
+    value: float | int
+    unit: str = "s"
+    step_index: int | None = None
+    category: str = "runtime"
```
gtong-nv, line 29: `category` should be enum?

flashdreams/flashdreams/runtime/interfaces.py
fangjunzhou-nv, line 1: I'm not sure if the current "pipeline" way of interface definition is a good idea or not. This requires users to compose the application in a sequential way. They need to choose the session, an adapter, then a model.

I think a better to structure the application is providing a model interface and a session interface. The model interface specifies model input schema. The session interface specifies user input schemas. The user will implement their Session declaratively. The input mapping happens inside a Session implementation.

flashdreams/flashdreams/runtime/inputs.py
```diff
@@ -0,0 +1,189 @@
...
+                for event in self.events
+                if time_window.contains(event.timestamp_s)
+            ),
+            snapshot=self.snapshot,
+            metadata=self.metadata,
+        )
+
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class ModelInputs:
```
fangjunzhou-nv, line 174: As commented in flashdreams/flashdreams/runtime/interfaces.py, I'd like a declarative way of defining model inputs. The user should be able to build a Model that expose the input scheme directly. Instead of using a dedicated ModelInput object.

flashdreams/flashdreams/runtime/mapping.py
fangjunzhou-nv, line 1: Similar to the comments on ModelInput, if we design the pipeline with declarative primitives the user should be able to design the mapping stage inside Session

flashdreams/flashdreams/runtime/metrics.py
fangjunzhou-nv, line 1: Is there any chance we can have Python decorator style metrics recorder like nvtx? We need more ergonomic profiling/benchmark tooling.

fangjunzhou-nv:
I'm going to make an alternative API proposal over the weekend.

flashdreams/flashdreams/runtime/config.py
jarcherNV: The model_id is defined earlier, which is non-optional, so this would be an optional extra thing, like if you're using the lingbot model, it could have some optional preset things that you could use instead, like a lingbot+wan or lingbot+taev. Unless you think that should not be optional?

flashdreams/flashdreams/runtime/config.py
jarcherNV: I was thinking here that for some models, the checkpoint may just be implicit and the ModelAdapter would be able to handle it just based on the model_id, or that could be overridden by this checkpoint str. But maybe that's wrong and we should just always require this? What do you think?

flashdreams/flashdreams/runtime/config.py
jarcherNV: What about "Execution placement and backend family for inference compute." instead?

flashdreams/flashdreams/runtime/config.py
jarcherNV: What about "Stable identity for the model adapter or runtime integration."

jarcherNV:
> I'm going to make an alternative API proposal over the weekend.

Sounds good. I’ll keep this PR scoped as the minimal T1 envelope for now, and we can compare it against your proposal. But we should not deviate from what reactor specifically liked about the design unless we have a good reason to, this will be their API too after all and ideally should be a joint effort with them.

flashdreams/flashdreams/runtime/config.py
jarcherNV: Yeah this could probably be clearer, but maybe we should avoid ‘all knobs’ because prompts/output/browser settings are intentionally outside InferenceConfig. What about ‘Runtime settings that affect model execution.’?

flashdreams/flashdreams/runtime/config.py
jarcherNV: What about ‘Where and how inference compute is run’ rather than ‘host’, since local backends are not hosted services.

flashdreams/flashdreams/runtime/config.py
jarcherNV: Yeah that probably makes sense. The fields like attention_backend and cache_policy are not really universal inference config and if we keep adding stuff like that then InferenceConfig could probably become bloated pretty quickly. I'll see if I can split this stuff out into a ModelConfig thing.

flashdreams/flashdreams/runtime/config.py
jarcherNV: Yeah a free string is probably ambiguous if this is a shared field. I'd prefer not to define the canonical cache policies now though, it's probably better to leave that for the T9 task.

flashdreams/flashdreams/runtime/config.py
jarcherNV: device=None lets the adapter/backend choose placement, which might matter for hosted/external runs and existing local configs. Adapters that require explicit placement can reject None in a validate step.

flashdreams/flashdreams/runtime/config.py
jarcherNV: Yeah that sounds reasonable, I'll see what checks I can add.

flashdreams/flashdreams/runtime/config.py
jarcherNV: Regarding the suggested wording change, what about "Optional attention implementation selector; None leaves the choice to the adapter."

flashdreams/flashdreams/runtime/inputs.py
jarcherNV: What about "These are model inputs required before starting the initial generation/session."

flashdreams/flashdreams/runtime/inputs.py
jarcherNV: What about phrasing it as "per-step model inputs required after the session starts."

flashdreams/flashdreams/runtime/inputs.py
jarcherNV: Yeah we should do that at some point before we merge all this stuff to main. I guess once the API is more settled we can throw that in.

flashdreams/flashdreams/runtime/metrics.py
jarcherNV: Yeah it probably should be once we know the canonical categories. For the T1 task I’d keep it open and we can revisit it in the future T9 task when benchmark/profiling categories are defined.

flashdreams/flashdreams/runtime/config.py
jarcherNV: Yeah there may be an LSP benefit. I’d keep literal for the T1 task and revisit enum when we actually implement this part of the API.

flashdreams/flashdreams/runtime/config.py
jarcherNV: For runtime_options/resource_hints, a shared TypedDict would be too generic to help much because the keys are adapter/backend-specific. It's probably better to use adapter-owned typed configs once those shapes are known when we implement them.

flashdreams/flashdreams/runtime/inputs.py
jarcherNV: Yeah that might make sense. We could probably do that in the T2/T3 tasks, I'll leave that up to whoever works on those tasks to decide.

flashdreams/flashdreams/runtime/interfaces.py
jarcherNV: Yeah that might be a useful design direction, we should probably compare it before committing T2-T4. My intent for this PR is the lower-level T1 envelope rather than the final user-facing authoring API. I think the current primitives can still sit underneath a declarative model/session API: a model can expose ModelInputSchema, a session can expose UserInputSchema, and the session can own or wrap an InputMapping. I’d prefer not to do a full pivot in this PR until we have the alternative proposal and can check it against LingBot, OmniDreams, the-internal-demo-that-shall-not-be-named and Reactor requirements.

flashdreams/flashdreams/runtime/inputs.py
jarcherNV: Yeah, I think part of the confusion is naming. `kind` is meant to be the event type, e.g. `move`, `look`, `reset`, etc. `event_id` is not meant to be the primary identity/hash of the event; it is only an optional source/correlation ID from a transport, replay trace, or external system. Many runtime-stamped events will not need one.

Maybe we just rename `event_id` to something like `source_event_id` or `correlation_id` and maybe rename `kind` to `event_type`.

What do you guys think?

flashdreams/flashdreams/runtime/inputs.py
jarcherNV: Yeah a declarative Model API could expose its schema directly. I still think we need a transport-neutral ModelInputs payload underneath for mapping, replay, tests, and benchmarks. The later declarative layer can wrap or produce ModelInputs rather than removing that boundary.

flashdreams/flashdreams/runtime/mapping.py
jarcherNV: A declarative session can own or wrap the mapping, but I’d keep the UserInputs -> InputMapping -> ModelInputs boundary explicit. Reactor specifically liked that separation, and it helps replay/benchmark/runtime overrides without changing the model step.

flashdreams/flashdreams/runtime/metrics.py
jarcherNV: Yes, I like that direction. For T1 I’d keep MetricsRecorder as the low-level sink, then add decorator/context-manager helpers in the metrics/profiling work once benchmark needs are clearer.

flashdreams/flashdreams/runtime/inputs.py
```diff
@@ -0,0 +1,196 @@
...
```
fangjunzhou, line 174: One new issue I found when reviewing this: How is the `ModelInputSchema` being used to constraint `ModelInputs`? One thing that confused me a lot is the schema and input are very separated and it feels like it requires a lot of hard coded constraint to validate schema. A good schema should be defined through some kind of typing system and being validated automatically. I suggest we should even look into some existing solutions like Pydantic: https://github.com/pydantic/pydantic

jarcherNV: Design-wise, I agree the current T1 schema is only lightweight field-presence validation. I’ll leave the decision on how to validate up to whoever implements the T2/T3 tasks.

PR #404, opened 2026-08-01
Fix benchmark video codec handling

jarcherNV:
Avoid deadlocks when writing MP4s by draining ffmpeg stderr while streaming frames to stdin. Add imageio-ffmpeg as a runner/dev fallback for systems without host ffmpeg, use it for clip comparison reads when MediaPy cannot find ffmpeg, and make optional baseline quality comparison skip cleanly when no decoder is available.


jarcherNV:
I am closing this PR as I think we will need to address these issues separately, along with some other changes.

PR #405, opened 2026-08-01
Refine benchmark reports and fix FFmpeg stderr draining

jarcherNV:
Fold derived benchmark summary metrics into their base metric rows so the HTML report shows the median-focused view developers expect. Constrain the scenario table layout and collapse long command strings behind a details control to avoid wide commands crowding adjacent columns.

Drain FFmpeg stderr concurrently while streaming video frames to stdin to avoid subprocess pipe-buffer deadlocks during MP4 writing.


gtong-nv (approved):
LGTM

PR #409, opened 2026-08-03
Adds manifest-driven Omnidreams WebRTC and stabilizes the interactive browser playback path

gtong-nv:
This PR adds 2 changes

1. Adds manifest parsing to the Omnidreams WebRTC, to be consistent with the local window mode. So now it takes `--manifest example_world_model_perf.yaml` to use the native kernel path.
2. The WebRTC demo can now prefer hardware H.264 encode when the browser negotiates it. This PR also preserve encoded packet order with backpressured NVENC queue delivery instead of silently dropping packets in the production path.

This combined can deliver 30 FPS when omnidream demo is running in a RTX6000 server, and viewed from the browser with H.264 decoding capability.

Test with 
```
uv run --package flashdreams-omnidreams torchrun --standalone --nproc_per_node 1 \
  -m omnidreams.webrtc.server \
  --host 0.0.0.0 \
  --port 8080 \
  --manifest example_world_model_perf.yaml \
  --scene-uuid 0d404ff7-2b66-498c-b047-1ed8cded60d4 \
  --warmup_chunks 5 \
  --warmup_timeout_s 1200
```

### DIFF
--- a/flashdreams/flashdreams/serving/webrtc/media.py
+++ b/flashdreams/flashdreams/serving/webrtc/media.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from fractions import Fraction
 from typing import TYPE_CHECKING
 
@@ -145,8 +145,9 @@ class NVENCVideoTrack(MediaStreamTrack):
     ``RTCRtpSender`` routes through ``H264Encoder.pack()`` for RTP
     fragmentation only. The encoder sets ``pts`` and ``time_base`` on
     each packet before enqueueing; this track only paces delivery to
-    ``fps`` and applies a drop-oldest overflow policy so a slow
-    consumer cannot stall the encode worker.
+    ``fps``. The async enqueue path applies backpressure so the browser
+    receives every frame in timestamp order instead of seeing silent
+    server-side drops.
     """
 
     kind = "video"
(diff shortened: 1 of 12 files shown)


gtong-nv:
/ok to test [5f117b7](https://github.com/NVIDIA/flashdreams/pull/409/commits/5f117b73dfdc1eb53a816127739f07c743ee0694)

ArielG-NV (approved):
Code looks good to me, before merging:
1. Address greptile review (explain why it is invalid or implement)
2. Fix CI failing

Notes: I think we should be merging CI arguments across different ways to serve the same demo (less disjoint between jpeg-stream vs. webrtc), but given that we will be refactoring the entire cmdline argument system later due to a larger refactor I don't think its worth addressing here.

gtong-nv:
> Code looks good to me, before merging:
> 
> 1. Address greptile review (explain why it is invalid or implement)
> 2. Fix CI failing
> 
> Notes: I think we should be merging CI arguments across different ways to serve the same demo (less disjoint between jpeg-stream vs. webrtc), but given that we will be refactoring the entire cmdline argument system later due to a larger refactor I don't think its worth addressing here.

Thanks! Addressed the greptile review - PyNvHardwareEncoder.deliver_chunk() now streams each encoded packet to NVENCVideoTrack as soon as the NVENC callback emits it

also fixed the linter issue

gtong-nv:
/ok to test [4ec2917](https://github.com/NVIDIA/flashdreams/pull/409/commits/4ec291787556a2fac02cda0aebddd2e0dc06c39c)

gtong-nv:
/ok to test [9fb5a84](https://github.com/NVIDIA/flashdreams/pull/409/commits/9fb5a843b0283b4cfe2345fe0c04efd6f139300f)

PR #411, opened 2026-08-04
Stream safetensors checkpoints into models

gtong-nv:
## Summary

- Stream LingBot’s sharded safetensors checkpoint directly into the materialized model.
- Skip creation of the duplicate merged safetensors cache for streamed loads.
- Retain the state-dict loading path for non-safetensors checkpoints and models that require whole-state-dict transforms.
- Enable streaming by default for LingBot presets.

## Resource savings

LingBot’s 16 transformer shards total approximately **69 GiB**. Previously, non-streaming loads also created a merged safetensors cache, requiring an additional ~70 GiB of disk space.

Streaming avoids this duplicate cache and avoids retaining a complete checkpoint state dict during loading. This reduces peak host-memory pressure and prevents out-of-memory failures on RAM-constrained systems.

## Loading-time delta

| Mode | Ready time (2 runs) | Median |
| --- | --- | ---: |
| Non-streaming | 20.708 s, 20.306 s | 20.507 s |
| Streaming | 23.576 s, 23.694 s | 23.635 s |

Benchmarked on an NVIDIA GB300 with cached LingBot checkpoints, streaming adds **3.13 s (15.3%)** to startup time. This ~3-second cost is an acceptable tradeoff for the 14B model’s reduced disk usage and host-memory pressure.

`stream_checkpoint` is enabled by default for LingBot and remains disabled by default for other models until their checkpoint layouts and transform requirements are validated.


ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/411/commits/b0c0d2f5b9480426bbc3b5a4abfef02085f1ea81

flashdreams/flashdreams/recipes/wan/transformer/wan21.py
```diff
@@ -157,6 +157,9 @@ class Wan21TransformerConfig(TransformerConfig):
     """Pre-load state-dict remap (e.g. Self-Forcing's
     ``generator_ema.model.…`` layout)."""
 
+    stream_checkpoint: bool = False
```
ArielG-NV, line 160: If `safetensors` are available (can be automatically detected from a checkpoint file afaik), shouldn't we by default always load them?

integrations/lingbot/lingbot/config.py
```diff
@@ -76,6 +76,7 @@
                 in_dim=16 + 4 + 16,
             ),
             checkpoint_path=LINGBOT_WORLD_V1_CHECKPOINT_PATH,
+            stream_checkpoint=True,
```
ArielG-NV, line 79: We likely want this config option for all demos. Unsure if we should make this change now or if this change will be something we implement during a larger refactor.

ArielG-NV (commented):
1 suggestion
1 question

integrations/lingbot/lingbot/config.py
gtong-nv: For smaller model or system with large RAM, this option is not helpful and will add some overhead when loading.
I'm testing the overhead to see if the difference matters or not. Will enable by default if not.

gtong-nv:
Mode | Ready time (2 runs) | Median
-- | -- | --
Non-streaming | 20.708 s, 20.306 s | 20.507 s
Streaming | 23.576 s, 23.694 s | 23.635 s

Benchmarked on GB300, lingbot checkpoints streaming is 3.13 s / 15.3% slower on cached startup
I think we should still keep streaming enabled for LingBot. It avoids the non-streaming path’s additional 70 GB merged safetensors cache and host-residency risk; the ~3 s startup cost is a reasonable tradeoff for this 14B model.

gtong-nv:
/ok to test [2aaeb74](https://github.com/NVIDIA/flashdreams/pull/411/commits/2aaeb74c7b2f404a74b434aef4a776fe6077d307)

ArielG-NV:
> Benchmarked on GB300, lingbot checkpoints streaming is 3.13 s / 15.3% slower on cached startup
I think we should still keep streaming enabled for LingBot. It avoids the non-streaming path’s additional 70 GB merged safetensors cache and host-residency risk; the ~3 s startup cost is a reasonable tradeoff for this 14B model.

Is this because we are issuing 1 big copy (`_copy_checkpoint_tensor`), because `tensor_names` is very large (code below), or something-else (a collection of costs)?
```
                for name in tensor_names:
                    tensor = source.get_tensor(name)
                    try:
                        tensor_bytes = _copy_checkpoint_tensor(
                            model_state[name], tensor
                        )
                        shard_copied_bytes += tensor_bytes
                        total_copied_bytes += tensor_bytes
                    finally:
                        del tensor
```
                        

Is this something we can potentially improve upon in your opinion?

PR #413, opened 2026-08-04
Implement T2, T3, and part of T4 from API refactor plan

aidanfnv:


flashdreams/flashdreams/runtime/interfaces.py
```diff
@@ -42,8 +42,8 @@ def close(self) -> None:
 class InferenceRuntime(Protocol):
```
fangjunzhou-nv, line 42: This is not necessary related to the input system design in this PR, but I suggest we remove InferenceRuntime as it's only job is to spin up and tear down InferenceSession. We should just leave that to the Application.

flashdreams/flashdreams/runtime/interfaces.py
```diff
@@ -56,10 +56,10 @@ def close(self) -> None:
 class ModelAdapter(Protocol):
```
fangjunzhou-nv, line 56: We should probably rename ModelAdapter to InferenceSessionSchema or something. It's just advertising the schemas for InferenceSession

flashdreams/tests/test_inference_runtime_api.py
```diff
@@ -328,18 +356,24 @@ def _drive_two_step_session(
...
     mapping.validate(
-        user_schema=adapter.user_input_schema,
-        model_schema=adapter.model_input_schema,
+        canonical_schema=adapter.canonical_input_schema,
+        inference_input_schema=adapter.inference_input_schema,
     )
-    initial_inputs = mapping.map_initial_inputs(
-        user_inputs=user_inputs,
-        model_inputs=model_inputs,
+    initial_inputs = mapping.map_global_inputs(
```
jarcherNV, line 370: I'm wondering if this is how we want it to behave.

The reference loop advances stateful input converters before the session starts. I think this canonicalizes the full session horizon for global inputs before runtime/session creation? But InputCanonicalizer converters are explicitly stateful and must be fed windows in session order.

KeyboardToDriverCommand.convert() mutates held-key state while reading events at canonical.py, line 177. If a real demo copied this loop, future key events could affect the initial state. Maybe the global conditioning should bypass canonicalization, or use a separate non-live startup path?

Not too sure about this one. Maybe @ArielG-NV or @fangjunzhou-nv may have an opinion. Or just throw this comment at codex and see what it thinks.

docs/inference_runtime_api_design.md
```diff
@@ -305,33 +311,63 @@ User inputs are not model inputs. A keyboard event does not have one universal
...
+UserInputs  ->  CanonicalInputs  ->  InferenceInput
+   raw           canonicalized          encoded
+```
+
+Raw device events are canonicalized into device-independent modalities before an
+application sees them, so adding a keyboard, gamepad, or wheel is a converter
+registration rather than an application change. `InferenceInput` is what an
+`InferenceSession` actually receives.
+
+`InferenceInput` describes the data the model or inference pipeline actually
```
jarcherNV, line 328: I think the implementation and the docs may be disagreeing with each other a little here, maybe just tell codex to do another once-over on the docs before you merge.

flashdreams/flashdreams/runtime/mapping.py
```diff
@@ -60,26 +67,315 @@ class IdentityInputMapping:
...
+        mapping_schema=reported_schema,
+        missing_modalities=tuple(missing_modalities),
+        missing_required_model_fields=missing_required,
+        satisfied_required_model_fields=satisfied_required,
+        available_optional_model_fields=available_optional,
+        unavailable_mapping_schemas=tuple(unavailable),
+    )
+
+
+def check_mapping_compatibility(
```
jarcherNV, line 316: The compatibility checking seems good, but not sure if it's wired into the primary adapter/mapping path yet?

This function looks useful, but I think ModelAdapter.default_input_mapping() returns only an InputMapping at interfaces.py line 80, and schema declaration is an optional refinement at.

Before demo migration, maybe the standard loop needs a clear way to get mapping schema(s) and run compatibility before loading checkpoints?

Not necessarily something that has to be fixed here, but we should keep a note of it when we get to demo migration maybe.

flashdreams/flashdreams/runtime/canonical.py
```diff
@@ -0,0 +1,387 @@
...
+            key for keys in self._bindings.values() for key in keys
+        )
+        self._state = KeyboardState(supported_keys=self._supported_keys)
+        self._schema = DeviceConverterSchema(
+            name=name,
+            produces=DRIVER_COMMAND,
+            device_kind="keyboard",
+            priority=priority,
+            consumes=(
+                UserInputCapability(
```
jarcherNV, line 159: Looks like this consumes key_down / key_up. I think the older realtime helper uses keydown / keyup at input.py line 77 (flashdreams/flashdreams/serving/realtime/input.py:77), and one runtime test uses keyboard.keydown at test_inference_runtime_api.py, line 44.

I'm wondering if we need to standardise the names yet, or leave that for later. I imagine the demos will need a minimal event catalog or something.

docs/inference_runtime_supported_inputs_inventory.md
```diff
@@ -0,0 +1,321 @@
...
+        InputField(name="camera_intrinsics", lifecycle="rollout_binding"),
+        InputField(name="memory_config", lifecycle="rollout_binding"),
+    ),
+)
+```
+
+```python
+sana_wm_model = InferenceInputSchema(
+    description="sana-wm",
+    global_fields=(
```
ArielG-NV, line 297: these fields should potentially have a semantic type as well?

example: a "negative prompt" may be treated differently than a positive prompt, the underlying model_arch needs support for the semantic type of each (since each are technically processed uniquely depending on the underlying model_arch)

docs/inference_runtime_supported_inputs_inventory.md
```diff
@@ -0,0 +1,321 @@
...
+        InputField(name="prompt", lifecycle="cache_init"),
```
ArielG-NV, line 298: why are we using new terminology for the frequency of an input?

I suppose lifescycle is fine

but cache_init should be something like "global_conditioning"?

docs/inference_runtime_supported_inputs_inventory.md
```diff
@@ -0,0 +1,321 @@
...
+        InputField(name="negative_prompt", required=False, lifecycle="cache_init"),
+        InputField(name="global_conditioning_frame", lifecycle="cache_init"),
+        InputField(
+            name="camera_trajectory_c2w",
+            semantic_type="c2w_sequence",
+            lifecycle="rollout_binding",
```
ArielG-NV, line 304: what is a rollout_binding, why not something like`per_step` or `every_step`?

ArielG-NV (commented):
Initial review; questions and suggestions

docs/inference_runtime_inputs_implementation.md
```diff
@@ -0,0 +1,287 @@
...
+```python
+from flashdreams.runtime import SESSION_START_ONLY, InferenceInputSchema, InputField
+
+schema = InferenceInputSchema(
+    global_fields=(
+        InputField(name="prompt", update_policy="step_boundary"),
+        InputField(name="scene_id", update_policy=SESSION_START_ONLY),
+    )
+)
+schema.unsupported_global_updates(
```
gtong-nv, line 99: what is `unsupported_global_updates` ?

aidanfnv: `unsupported_global_updates()` is a helper on `InferenceInputSchema` for mid-rollout global-conditioning update requests. It looks at the non-empty `InferenceInput.global_conditioning` payload passed to a step/update and returns the field names whose schema marks them `SESSION_START_ONLY`, meaning they can be supplied at session start but should not be swapped while the rollout is active.

docs/inference_runtime_supported_inputs_inventory.md
aidanfnv: Somehow in the process of implementing this, the global conditioning inputs got split into 5 "lifecycles", where `rollout_binding` represents "values bound after cache initialization but before AR steps, such as HY-WorldPlay action labels, camera tensors, and memory state".

That is not useful to specify, so `rollout_binding` is indeed a per-step global conditioning input. I will change this to `per_step`.

docs/inference_runtime_supported_inputs_inventory.md
aidanfnv: For "lifecycle", there was some partial attempt at separating the timeframe/frequency of input, frequency of consumption, and scope of usage (global vs step) and sometimes those share a label, sometimes they don't, which is needlessly confusing. "cache_init" describes a one-time global conditioning input.

`global_fields` is already the inputs for global conditioning (I will rename to `global_conditioning_fields`), so we do not need to specify "global_conditioning" again, and what makes "cache_init" different from other one-time global conditioning inputs is irrelevant here.

Honestly, "lifecycle" is not really fine here, it really confused myself and any agent I used here. Since the global vs step distinction is already handled elsewhere, I will rename it to `frequency_consumed`, where this example would have the value `"once"`.

aidanfnv: See https://github.com/NVIDIA/flashdreams/pull/421 for the change

docs/inference_runtime_supported_inputs_inventory.md
aidanfnv: See https://github.com/NVIDIA/flashdreams/pull/421 for the change

docs/inference_runtime_inputs_implementation.md
aidanfnv: It will be removed in https://github.com/NVIDIA/flashdreams/pull/421

docs/inference_runtime_supported_inputs_inventory.md
aidanfnv: Turns out that part of the doc is just wrong about what a `semantic_type` entails. 
`name="negative_prompt"` and `name="prompt"` would provide the negative vs positive distinction.
I will be renaming `semantic_type` to `input_modality`, and have corrected the docs, see https://github.com/NVIDIA/flashdreams/pull/421

aidanfnv: `name="negative_prompt"` would distinguish a negative prompt from a positive `name="prompt"`.
The docs describe semantic_type incorrectly, and included stuff like "path" which are not even supposed to be in that field.

I have changed "semantic type" to "input modality" and corrected the doc, see https://github.com/NVIDIA/flashdreams/commit/e1fa7d0da7113a0aebfad3d7a70b542b31710b2c (accidentally committed that a bit prematurely).

docs/inference_runtime_api_design.md
aidanfnv: Addressed this after the fact in e1fa7d0da7113a0aebfad3d7a70b542b31710b2c

flashdreams/tests/test_inference_runtime_api.py
aidanfnv: Bypassed canonicalization for global conditioning in cced4c743443f58bef16702647e4ea45c9478f75

PR #415, opened 2026-08-04
`flashdreams.runtime` inference session

fangjunzhou-nv:


fangjunzhou-nv:
Updated to make it clear that InputSystem doesn't handle key bind, event pulling, etc. That should be handled by the upstream input system (another library implemented in python, or an external game engine in the future).

PR #424, opened 2026-08-06
Fix CUDA context leak during NVENC probing

gtong-nv:
## Summary

Restore PyTorch's CUDA device context after probing PyNvVideoCodec for NVENC support.

`GetEncoderCaps` can replace the calling thread's active CUDA context when the probe fails. Lingbot then falls back to aiortc, but subsequent Triton kernels run with the wrong context and fail with `invalid resource handle`. Retrying the partially advanced rollout produces the secondary `AR step out of order` error.

The encoder-selection boundary now restores the original PyTorch device context in a `finally` block for every hardware-probe outcome.


jarcherNV (approved):
LGTM

gtong-nv:
/ok to test [d4f3532](https://github.com/NVIDIA/flashdreams/pull/424/commits/d4f353242196ab8b87faf56763f5210501f1be2c)

PR #426, opened 2026-08-07
Share demo WebRTC plumbing across models (#424)

jarcherNV:
Create a follow-up branch to consolidate the demo WebRTC path so OmniDreams and LingBot can use the same shared server/session construction, with model-specific behavior supplied through small adapter hooks. Cherry-pick the NVENC probing fix needed before refactoring the WebRTC stack.


jarcherNV:
Closing this PR as this work is no longer needed.

PR #430, opened 2026-08-08
[Interactive Drive] Add physics via Physx

### LINKED ISSUE #313: [interactive-drive] better physics
Currently real physics doesn't exist in interactive-drive and it's really easy to run out of bounds.

ArielG-NV:
Fixes: #313
Changes:
* Added physx to drive the backend of physics. This requires maintaining our own physx python bindings.
* Removed Vulkan backend from ludus-renderer due to significant overhaul causing this backend to be a maintenance burden that requires a complete rewrite. If there is need for the backend in the future we will just rewrite the code then.
* Physics are generally obeying physx as the main-controller, but some hacks are added to prevent out-of-distribution behavior (yaw-lock on vehicles, speed lock on non-ego vehicles, etc...) 
* Changed the BEV render to be an orthographic top-down view for reasons that it makes driving+navigation easier. Added a `green` car to identify the ego on the BEV.
* Made all collision logic (and other game-like-frills) under the `--game-mode` flag for the interactive-drive demos
* Added a new view when clicking `3`, a physx-debug view. This view enabled viewing the colliders we setup on the backend of physx: vehicles, pedestrians, road-boundaries as walls, etc....
* Added a visual-flare when hitting/getting-hit by objects with enough force.
* Tried to fit all objects into a very 'hacked-up' scene graph structure for ease of future feature additions


integrations/omnidreams/ludus-renderer/ludus_renderer/_physx_native.py
```diff
@@ -0,0 +1,231 @@
...
+            try:
+                stale = time.time() - lock_path.stat().st_mtime > 1_800.0
+            except FileNotFoundError:
+                continue
+            if stale:
+                lock_path.unlink(missing_ok=True)
+                continue
+            if time.monotonic() >= deadline:
+                raise TimeoutError(
+                    f"timed out waiting for the PhysX build lock at {lock_path}"
```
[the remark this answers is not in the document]
ArielG-NV: " is not considered stale until thirty minutes"

10 minutes is enough to know if a process is likely stale.

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/430/commits/e425c34208e62ddf6eaeb5271545ec18bd841b85

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/430/commits/c97940aafbc9a65aaa7765f4d5a9ee3ffc4907cc

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/430/commits/5210855f73f06c9ef0dceb34190adab8c2b1bf0c

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/430/commits/4dbe02485689d24be10d626b446c0dfc5ef08dbb

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/430/commits/493592d4f616a4194bdf5c26f2ea4ef9123d5828

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/430/commits/cb82d2cbc33b8f53c525971003d03d0c2d229346

integrations/omnidreams/ludus-renderer/README.md
```diff
@@ -51,60 +44,45 @@ Dependencies installed:
...
+- CMake 4+ for the one-time native module build
+- NVIDIA PhysX 5.9.0, fetched from the official repository at a pinned commit
+  and verified by SHA-256
+
+Prebuild PhysX during environment setup so interactive startup never pays the
+one-time compile cost:
+
+```bash
+uv run --package ludus-renderer python -c "from ludus_renderer import prepare_physx; prepare_physx()"
+```
```
gtong-nv, line 56: wonder if we should add it to omnidreams, triggered by something like  
```
uv run --package flashdreams-omnidreams omnidreams-prepare --physx
```

ArielG-NV: the whole "magically compiles for you" thing is quite convenient, I would actually prefer to remove all `prepare` commands and just add "single methods" that set things up for users. If a system is known to be compatible, it should 'just work' if we call a cmd.

ArielG-NV: at the end of the day:
* if missing a file we need to detect and warn, letting users know to call `prepare`
* this means we already need technology to "know when to recompile/compile" (why would we not just compile then?)
* we want automatic recompiles in-case file-hash changes and build-system needs a rebuild (manually calling prepare to rebuild is a messy practice)

gtong-nv (approved):
I have not checked each line of code, but tested it locally with RTX 6000 pro and no regression.
LGTM

PR #440, opened 2026-08-11
[Running Flashdreams] Fix newer aarch64 container support

ArielG-NV:
Currently, `uv sync --extra dev --extra runners` will fail on newer containers (`manylinux_2_39_aarch64`+) with `transformer-engine-cu13` incompatibility.

This PR changes the utilized version of `transformer-engine-cu13` into a supported version.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -4778,10 +4778,10 @@ wheels = [
 
 [[package]]
 name = "transformer-engine"
-version = "2.17.0"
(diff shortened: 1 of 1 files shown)


ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/440/commits/2036ade82f33b3bc6ac50a8f52c7747e8ca0991a

PR #441, opened 2026-08-11
Update docs workflow setup-uv action

jarcherNV:
Bump the docs workflow from astral-sh/setup-uv@v4 to @v6 so it matches the rest of CI and avoids the Node certificate failure during action setup.


ArielG-NV (approved):
LGTM

PR #445, opened 2026-08-11
Fix OmniDreams native CUDA architecture selection on GB300

fangjunzhou-nv:
## Summary

Previously, the native extension defaulted to 12.0a whenever the user did not explicitly set a CUDA architecture. GB300 devices report compute capability 10.3 and require an architecture-specific 10.3a build. As a result, the extension could be compiled for an incompatible target and fail to load or execute on GB300.

This PR fixed the OmniDreams single-view native extension to compile for the current GPU architecture instead of always defaulting to 12.0a.

## Background

The native extension includes SageAttention-3 by default. Its FP4 path uses architecture-specific PTX instructions:

https://github.com/NVIDIA/flashdreams/blob/ac214dde70455c1b92f5b0a99e447253b568f484/integrations/omnidreams/omnidreams_singleview/src/dit_streaming/kernels/sage3_attention.cu#L85-L107

SageAttention’s own build configuration compiles compute capability 12.0 devices with:

`-gencode=arch=compute_120a,code=sm_120a`

The a suffix enables the architecture-specific features required by these FP4 kernels.


ArielG-NV:
This fix looks like a hack. If PyTorch does not recognize/list a CUDA arch something is going very wrong (something is already a hack?).

I prefer if we do not merge this and instead figure out the root cause of the problem and address it in the issue description for future maintainers of the code before we consider one off hacks.

fangjunzhou-nv:
> This fix looks like a hack. If PyTorch does not recognize/list a CUDA arch something is going very wrong (something is already a hack?).
> 
> I prefer if we do not merge this and instead figure out the root cause of the problem and address it in the issue description for future maintainers of the code before we consider one off hacks.

The omnidreams singleview was using PyTorch CUDA extension that compiled specifically targeting CUDA architecture 12.0a (which is the arch of RTX PRO 6000). What I'm adding in this PR is using `torch.cuda.get_device_capability()` to detect the corresponding arch when compiling PyTorch extension.

integrations/omnidreams/omnidreams/native/omnidreams_singleview.py
```diff
@@ -463,19 +470,38 @@ def _resolved_max_jobs(max_jobs: int | str | None) -> str | None:
     return str(min(os.cpu_count() or 1, _DEFAULT_MAX_JOBS_CAP))
 
 
-def _resolved_cuda_arch_list() -> str | None:
-    if os.environ.get(_PYTORCH_CUDA_ARCH_LIST_ENV):
+def _format_cuda_arch_list(capability: tuple[int, int]) -> str:
+    major, minor = capability
+    suffix = "a" if major in _ARCH_SPECIFIC_CUDA_MAJORS else ""
+    return f"{major}.{minor}{suffix}"
```
ArielG-NV: It looks like all sm_12X will become sm_12Xa

Same for sm11X

Is this intended behavior?

ArielG-NV (commented):
Question

integrations/omnidreams/omnidreams/native/omnidreams_singleview.py
fangjunzhou-nv: The original omnidreams CUDA extension targets 12.0a here:

https://github.com/NVIDIA/flashdreams/pull/445/changes#diff-cc80dd3171bfab8f5a0cf63db3d8e1057937e414f74b6ffc93895157f489de2aL58

I haven't tried to compile the kernel with CUDA 10.3 without a so I don't know if the extension can even be compiled without a.

Ideally we want to have more different hardwares running CI to varify this. But if we just want to get omnidreams singleview CUDA extension demo working on GB300 (which is what I'm trying to do to get some perf numbers). This should be all we need.

I guess this can also be another reason Triton is preferred over CUDA for accelerated kernels.

jarcherNV:
Regarding Ariel's comments, I dug into this a bit.

From what I can tell, major in {10, 12} → always "a" is actually correct for every chip that exists today. So this isn't a blind guess, it matches NVIDIA's published tables. That said, it's not actually guaranteed for a future chip NVIDIA ships.

Perhaps we swap _ARCH_SPECIFIC_CUDA_MAJORS for an explicit table of the exact (major, minor) pairs we've verified need a (i.e. {(10,0), (10,3), (12,0), (12,1)} or whatever's actually been tested), and raise a clear error for anything outside that table? That keeps today's fix working exactly as-is for GB300, which I think addresses the "this feels like an unbounded hack" concern without requiring a full root-cause investigation before this can land.

This has the downside of needing to be updated when we want to support newer hardware, but after checking, we seem to already have that requirement, just much more silently, and this would instead give a more clear, exact, error message.

Separately, non-blocking nit: _resolved_cuda_arch_list() (around line 499) looks like dead code after this refactor — its only caller, _scoped_cuda_arch_list, is now always invoked with an explicit cuda_arch_list argument from load_extension, so the None-fallback branch that calls it never executes, and no test references it either. Worth deleting as part of this PR so the next reader isn't left wondering which of the two "resolve the arch" functions is actually live.

Maybe there is a better way of handling this? What do you guys think @ArielG-NV and @fangjunzhou-nv ?

ArielG-NV:
I think its about time to overhaul and unhackify our cuda compilation path (flashdreams has enough hacks as is)

fangjunzhou-nv:
I spent some time actually dig into the root cause and it turns out the sage attention implementation in the CUDA kernel used some intrinsics that are **ONLY** available on sm120a machine (see the PR desc for detail).

At this point I've just spent too much time on this. My understanding is we want to ditch the custom CUDA kernel for omnidreams interactive drive at some point. We either use the new `flashdreams.accelerated` Triton kernel library I'm currently building or even if we want to use CUDA we want to do it properly.

I was originally submitting this PR just as a in place minor fix to get omnidreams CUDA kernel working on GB300 to get some perf numbers.

I do think if we also want to gather the perf numbers for omnidreams CUDA kernel in our incoming ci benchmark we need this fix to at least collect the numbers.

fangjunzhou-nv:
Since I've found the root cause is sage3 requiring sm120a to compile, I've updated the omnidreams native to disable sage3 when targeting other platforms. This helps the omnidreams CUDA extension to compile on all platforms while keeping the SAGE3 impl compiled on sm120a platforms.

ArielG-NV:
It looks like (given the PR desc provided) that the correct solution is to remove all hacky CUDA-Arch detection code and instead modify the Sage3 kernel (that we have inside the flashdreams code-base).

We should modify it such that there is an impl for CUDA arch sm120+ and sm100<= (macro should allow detecting at compile time the target arch being compiled for).

This way we may even get perf benefits from SageAttention-3 while removing hacks in the code, although I do not know if this is the case (I just prefer we don't start injecting hacks into code when the correct solution is cleaner).

ArielG-NV (commented):
(Refer to comment above)

PR #448, opened 2026-08-12
Add unified inference runtime and demo API

jarcherNV:
Introduce shared runtime/session contracts, demo specs, run contexts, session drivers, step pipelines, model input providers, output sinks, metrics, benchmark hooks, and compatibility wrappers for the old runner surface.

Migrate OmniDreams, LingBot, and text-to-video demo paths onto the unified runtime across replay, null, MP4, WebRTC, and realtime modes. Move model-specific input preparation into providers, including replay traces, keyboard-driven controls, text events, WebRTC input catch-up, and Ludus replay inputs.

Unify WebRTC viewer and serving plumbing around shared session management, browser event handling, realtime input sources, session validation, transport edges, and packaged viewer resources while keeping legacy launch paths available for later cleanup.

Add focused CPU/GPU tests, demo-runtime CI workflows, artifact validation, README commands, and migration-plan tracking for remaining legacy runner, mapping, and WebRTC compatibility debt.


configs/launch_manifest/omnidreams_mp4.yaml
```diff
@@ -0,0 +1,13 @@
...
+mode: mp4
+
+scenario:
+  example_data: true
+  total_blocks: 60
+  pixel_width: 1280
+  pixel_height: 704
+
+output:
+  path: ../../outputs/omnidreams-replay.mp4
```
jarcherNV, line 12: Are we sure we want to specify paths this way? For example, I typically just run this stuff from the root folder, so this output for example would not work for me. I know people did not like the idea of providing an output path on the command line, I would suggest that if no output path is specified, we just use a default filename of output.mp4 or something, with no path.

jarcherNV, line 12: Also leaving a note that this style would not be useful to me, I would need to be able to do multiple different benchmarks of different scenes and different output locations, I personally think that specifying an output path on the command line is more convenient than creating a different yml file for each. As it stands, I would use the old style commands for every demo, and so would the CI. So my question is, do people prefer specifying a config approach for different runs? Or just have standard command line args with default values? I feel like that latter is more convenient for cli apps right?

configs/launch_manifest/omnidreams_webrtc.yaml
```diff
@@ -0,0 +1,21 @@
+schema_version: 1
+runner: omnidreams
+mode: webrtc
+
+runner_overrides:
+  device: cuda:0
+
+scenario:
+  scene_uuid: 0d404ff7-2b66-498c-b047-1ed8cded60d4
```
jarcherNV, line 9: Would we need to create a separate config file for every possible scene like this? There are a lot of scenes in omnidreams.

integrations/omnidreams/omnidreams/demo/README.md
```diff
@@ -0,0 +1,92 @@
+<!--
```
jarcherNV, line 1: I feel like the command line arg variant should also be included, it is much more convenient for ad-hoc testing than having to update a manifest file each time.

apps/t2v_demo/backends.py
```diff
@@ -0,0 +1,88 @@
...
+        name = preset_name or self.default_preset_name
+        if name not in self.preset_names:
+            raise ValueError(
+                f"Unknown {self.key} preset {name!r}. Available presets: "
+                f"{', '.join(self.preset_names)}."
+            )
+        return PRESETS[name]
+
+
+BACKENDS: dict[str, T2VBackend] = {
```
ArielG-NV, line 34: This should not exist. We should be auto-detecting from a list of presets defined (model_arch+check_point+model_checkpoint_options)?

apps/t2v_demo/backends.py
```diff
@@ -0,0 +1,88 @@
...
+        preset_names=(
+            "self-forcing-wan2.1-t2v-1.3b",
+            "self-forcing-wan2.1-t2v-1.3b-taehv",
+            "self-forcing-wan2.1-t2v-1.3b-sink5-window7-rerope",
+        ),
+    ),
+}
+
+
+def resolve_backend(value: str) -> T2VBackend:
```
ArielG-NV, line 63: Incorrect logic; we are supposed to "automatically detect" if a model-arch chosen is supported based on the Input/Output schema associated with the model.

apps/t2v_demo/backends.py
```diff
@@ -0,0 +1,88 @@
...
+    """Resolve a CLI/UI backend key."""
+    try:
+        return BACKENDS[value]
+    except KeyError as exc:
+        raise ValueError(
+            f"Unknown backend {value!r}. Available backends: {', '.join(BACKENDS)}."
+        ) from exc
+
+
+def backend_choices() -> tuple[str, ...]:
```
ArielG-NV, line 73: Incorrect logic; we are supposed to "automatically detect" if a model-arch chosen is supported based on the Input/Output schema associated with the model.

apps/t2v_demo/backends.py
```diff
@@ -0,0 +1,88 @@
...
+    """Return stable CLI choices."""
+    return tuple(BACKENDS)
+
+
+def backend_metadata() -> list[dict[str, Any]]:
```
ArielG-NV, line 78: Why is this in `t2v_demo`. At the very least this should just be something our cmdline API can report when choseing configs to run (`flashdreams-run <demo> --model-arch <model arch>` ==> `flashdreams-run <demo> --model-arch ?` and other commands)

apps/t2v_demo/launch.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class T2VLaunchCapability:
+    """Expose replay and persistent WebRTC modes for the app-owned T2V demo."""
+
+    def supported_modes(
+        self, config: RunnerConfig, options: LaunchOptions
+    ) -> tuple[LaunchMode, ...]:
+        del config, options
+        return ("mp4", "null", "webrtc")
```
ArielG-NV, line 26: should not need to specify explicitly; should be inferred based on outputs of demo.

more importantly to note: we will provide an *option* to override default behavior for how "webrtc" displays results

the default will likely be "we expect ONLY a video frame as output ==> this just gets plastered on the web)

apps/t2v_demo/launch.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def resolve(
```
ArielG-NV, line 28: A user should not have to see this, this is something we 100% could hoist out into a layer the user never sees

apps/t2v_demo/launch.py
```diff
@@ -0,0 +1,88 @@
...
+            launch=partial(
+                _launch,
+                config=t2v_config,
+                mode=t2v_mode,
+                options=options,
+            ),
+        )
+
+
+def _launch(
```
ArielG-NV, line 56: A user should not have to see this, this is something we 100% could hoist out into a layer the user never sees

integrations/omnidreams/omnidreams/demo/README.md
jarcherNV, line 1: Proposed alternative: as part of providing higher level functions, we have the flashdreams API support command line args for a bunch of its own stuff, with reasonable defaults, then an app can specify any additional ones it wants to add and their own defaults.

This could also be specified by a manifest file of some kind.

In this alternative, the command line args would come from the manifest file, rather than the command line specifying different manifest files.

ArielG-NV (commented):
some review

ArielG-NV:
> So we decided that command line args are too hard? I forget what that discussion was about. Currently as it stands, you'd need about 20,000 copies of the same manifest file if you wanted to be able to load the different omnidreams scenes from huggingface.

This is not great currently, but we should be supporting cmdline_args + yaml via a transparent interface that isn't too messy

so for this:
* we will provide some default pairs of `model_arch+checkpoint` via a '<model_arch>/presets/preset.yaml'. This is bare-minimum preset of only mandatory args, nothing optional goes here. Tag mandatory args with `@mandatory_arg(some_default, "arg_name")`. Some prototype members are mandatory like "checkpoint" for every model_arch.
* model_arch all have tunable knobs, and these will "have defaults". All tunable members require a `@optional_arg(some_default, "arg_name")` annotation. This would also tell our architecture "who is an arg and for what type?". If nothing is specified further CI will just use defaults when running (for every defined `preset.yaml`). If user wants CI to autotune, they must provide a second `@autotune_range(something_trivially_composable_into_a_generator_with_a_start_and_end_and_non_infinite)`.
* demo mandatory bits will have defaults, just make it so they are exposed via the same `@mandatory_arg` annotation again I suppose. 
* demo optional bits for perf can follow the design as well. Supports `@autotune_range`.

All demos accept a manifest for the model+checkpoint+mandatory_args. All demos then accept either: 1. nothing 2. a manifest or 3. cmdline args; this does not matter since it all decomposes into same thing via reflection on the annotations we enforced.

and via reflection on annotations we can collect all `encoder`/`decoder` args from components of our `model_arch` as well, stacking them on top of the model_arch args via a naming scheme like `decoder.arg_1` and `decoder.arg2`

^^^
Potential design

apps/t2v_demo/presets.py
```diff
@@ -0,0 +1,243 @@
...
+from flashdreams.recipes.wan import (
+    Wan21TransformerConfig,
+    WanDiTNetwork1pt3BConfig,
+    WanInferencePipelineConfig,
+    WanVAEDecoderConfig,
+)
+
+
+@dataclass(frozen=True, slots=True)
+class T2VPreset:
```
ArielG-NV, line 35: this is not universally signaling to any external system via reflection that "it is a list of arguments a user can adjust". May be issue later on.

apps/t2v_demo/presets.py
```diff
@@ -0,0 +1,243 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ArielG-NV, line 1: This config file has a lot of dependencies... this is why I prefer yaml's, no dependencies and forces us to either:
1. hack everything together
2. implement solid reflection

apps/t2v_demo/README.md
```diff
@@ -0,0 +1,22 @@
...
+uv run python -m apps.t2v_demo.app replay --backend causal-forcing --output outputs/t2v.mp4
+```
+
+Serve streamed WebRTC output:
+
+```bash
+uv run python -m apps.t2v_demo.app webrtc --backend self-forcing
+```
+
+`--backend` accepts `causal-forcing`, `cosmos-predict2`, and `self-forcing`.
```
ArielG-NV, line 19: we should not specify this, this will code-rot or convince AI to add things that may or may not be true.

Just tell the user about the `?` option to query available options.

jarcherNV:
> > So we decided that command line args are too hard? I forget what that discussion was about. Currently as it stands, you'd need about 20,000 copies of the same manifest file if you wanted to be able to load the different omnidreams scenes from huggingface.
> 
> This is not great currently, but we should be supporting cmdline_args + yaml via a transparent interface that isn't too messy
> 
> so for this:
> 
> * we will provide some default pairs of `model_arch+checkpoint` via a '<model_arch>/presets/preset.yaml'. This is bare-minimum preset of only mandatory args, nothing optional goes here. Tag mandatory args with `@mandatory_arg(some_default, "arg_name")`. Some prototype members are mandatory like "checkpoint" for every model_arch.
> * model_arch all have tunable knobs, and these will "have defaults". All tunable members require a `@optional_arg(some_default, "arg_name")` annotation. This would also tell our architecture "who is an arg and for what type?". If nothing is specified further CI will just use defaults when running (for every defined `preset.yaml`). If user wants CI to autotune, they must provide a second `@autotune_range(something_trivially_composable_into_a_generator_with_a_start_and_end_and_non_infinite)`.
> * demo mandatory bits will have defaults, just make it so they are exposed via the same `@mandatory_arg` annotation again I suppose.
> * demo optional bits for perf can follow the design as well. Supports `@autotune_range`.
> 
> All demos accept a manifest for the model+checkpoint+mandatory_args. All demos then accept either: 1. nothing 2. a manifest or 3. cmdline args; this does not matter since it all decomposes into same thing via reflection on the annotations we enforced.
> 
> and via reflection on annotations we can collect all `encoder`/`decoder` args from components of our `model_arch` as well, stacking them on top of the model_arch args via a naming scheme like `decoder.arg_1` and `decoder.arg2`
> 
> ^^^ Potential design

Sounds like you are volunterring to setup the new CI system to me.

wlewNV (approved):

ArielG-NV (commented):
another short review I had time to add.

PR #451, opened 2026-08-12
Update benchmark configs for migrated LingBot launch

jarcherNV:


configs/deterministic_quality_benchmarks.json
```diff
@@ -122,18 +138,27 @@
...
+        "--output.fps",
+        "16",
+        "--output.path",
+        "{output_dir}/lingbot-world-fast-taehv-one-minute-review.mp4",
+        "--output.stats-path",
+        "{output_dir}/stats_lingbot_world_fast_taehv_one_minute_review.json"
       ],
-      "warmup_steps": 1,
+      "output_dir_arg": null,
+      "warmup_steps": 6,
```
ArielG-NV, line 160: warmup steps went up a lot, any reason in particular for the tests in general?

wlewNV (approved):

flashdreams/flashdreams/runtime/demo/outputs.py
```diff
@@ -267,6 +422,139 @@ def _result_record(result: StepResult) -> Mapping[str, object]:
...
+                unit=unit,
+                category=category,
+                step_index=result.step_index,
+                metadata=metadata,
+            )
+        )
+    return tuple(samples)
+
+
+def _normalize_metric_sample(
```
ArielG-NV, line 510: Looks a bit odd that we need this...

flashdreams/tools/benchmarks/metrics.py
```diff
@@ -470,6 +612,66 @@ def _summary_stats(values: list[float]) -> dict[str, float | int]:
...
+    if measured_steps == 0 or total_s <= 0:
+        return None
+    generated_fps = total_frames / total_s
+    return {
+        "count": measured_steps,
+        "min": generated_fps,
+        "max": generated_fps,
+        "mean": generated_fps,
+        "median": generated_fps,
+        "p90": generated_fps,
```
ArielG-NV: we have samples closer to frames per second via `results`, so why are we reporting p90 as equal to the mean FPS instead of p90 based on the collected samples in `results`?

At a minimum we should likely report N/A or Unsupported here since p90 of mean FPS for runtime is likely misleading.

ArielG-NV (commented):
Most of the AI tests are quite fragile and should be removed. 

1 Metric may be reported incorrectly

Some suggestions.

configs/deterministic_quality_benchmarks.json
jarcherNV: I don't think so, I was probably just eating the slow initial steps previously.

flashdreams/tools/benchmarks/metrics.py
jarcherNV: Sounds fair, I'll look into this.

PR #452, opened 2026-08-13
Add shared native-window demo backend

wlewNV:
Provide a reusable realtime input/output path and a split triangle app/model example so package discovery, model execution, and every presenter mode can be reviewed end to end.


jarcherNV (approved):
LGTM

flashdreams/flashdreams/runtime/demo/application.py
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
csyonghe, line 1: should this file be right under `/runtime/` instead of `/runtime/demo`?

flashdreams/flashdreams/runtime/demo/application.py
```diff
@@ -0,0 +1,33 @@
...
+class FlashDreamsApplication(DemoAdapter, Protocol):
+    application_name: str
+    description: str
+    config: InferenceConfig
+    scenario: object
+    fps: int
+    video_width: int
+    video_height: int
+    output_layout: VideoTensorLayout
+    default_mode: ApplicationMode
```
csyonghe, line 28: ideally, this should be  `presentationLayer` object that provides the input/output surface, so adding a new kind of implementation would not need modify this file (i.e. ApplicationMode definition)

flashdreams/flashdreams/runtime/demo/spec.py
csyonghe, line 1: what is `spec`?

flashdreams/flashdreams/serving/native_window/demo.py
```diff
@@ -0,0 +1,337 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
csyonghe, line 1: why is this file named `demo`?

integrations/triangle_model/triangle_model/application.py
```diff
@@ -0,0 +1,132 @@
...
+import argparse
+from collections.abc import Sequence
+from typing import cast
+
+import torch
+from flashdreams.runtime import InferenceConfig, InferenceInput, StepRequest, StepResult
+from triangle_app import DEFAULT_TRIANGLE_COLOR, TriangleApp, TriangleScenario
+
+
+class TriangleModel(TriangleApp):
```
csyonghe, line 17: yeah, my thinking is that if `TriangleApp` is something like t2v, it should only carve out the model inferencing logic, and you only need to provide the checkpoint loading and inferencing logic in this file, without needing to define any session/runtime bloat.

PR #453, opened 2026-08-13
Add modular FlashDreams application runner

gtong-nv:
## Architecture

```text
application module              flashdreams-runner                 I/O mode
+------------------+       +--------------------------+       +---------------+
| create_runtime() | ----> | initialize Runtime       | ----> | Replay / MP4  |
|                  |       | create Session           |       | WebRTC        |
| Runtime          |       |                          |       | None          |
|  model weights   |       | loop:                    |       +-------+-------+
|  global state    |       |   input = mode.read()    |               |
|                  |       |   output = Session.      |<--------------+
| RunnerInference  |<------|       step(input)        |
| Session          |       |                          |
|  prompt/cache    |       |   mode.write(output)     |-------------->+
|  game state      |       | destroy Session/Runtime  |
+------------------+       +--------------------------+
```


apps/flashdreams_app/flashdreams_app/__init__.py
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
csyonghe: I feel like `flashdreams_app` shouldn't be called `app`, it is the runner that can run other apps. and it shouldn't be located in a place parallel to `t2v_app`. The runner is the "host", "t2v" is the thing that gets run by the host.

apps/flashdreams_app/flashdreams_app/cli.py
```diff
@@ -0,0 +1,243 @@
...
+    if not isinstance(spec, PipelineAppSpec):
+        raise TypeError(
+            f"Provider {args.provider!r} create_app_spec() returned "
+            f"{type(spec).__name__}, expected PipelineAppSpec."
+        )
+    runtime = PipelineAppRuntime(
+        spec=spec,
+        device=environment.device,
+    )
+    if args.mode == "webrtc":
```
csyonghe: so the main loop is sort of duplicated and implemented differently in the webrtc path and mp4 path. This is fine for now but could be cleaned up by abstracting native window/web rtc into objects, so you make the main loop implemented just once.

apps/flashdreams_app/flashdreams_app/contracts.py
```diff
@@ -0,0 +1,146 @@
...
+    initialize_cache: PipelineCacheInitializer
+
+    def __post_init__(self) -> None:
+        if not callable(self.initialize_cache):
+            raise TypeError("PipelineContract.initialize_cache must be callable.")
+
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class PipelineAppSpec:
+    """Declarative application definition consumed by the generic host runtime."""
```
csyonghe: I don't think you can possibly express all forms of applications in a declarative way. An application should be defined by an object that provides `init()` and `generate()`, not a metadata/manifest.

apps/flashdreams_app/flashdreams_app/__init__.py
gtong-nv: yea, once the basic logic is implemented, this is easy to change.

apps/flashdreams_app/flashdreams_app/cli.py
gtong-nv: with the current runtime API, looks like we have to separate it into branches

PR #454, opened 2026-08-13
Add public demo application contracts

jarcherNV:
Introduce the flashdreams.demo facade for application-facing demo protocols and thin adapters over the existing runtime demo stack. Cover the Phase 1 contract shape with focused CPU tests.


jarcherNV:
Closing this as it is no longer needed.

PR #455, opened 2026-08-13
[Demo Runner/Application Framework] T2V Cleanup

ArielG-NV:
Currently our method to designing demos is quite messy. We should amend this by stopping to use bits of code that are "bloated wrappers" and focus more on moving towards a formalized design.

The additions in this PR focus on reusing existing code (WebRTC session manager, output sinks, input schema, ...), explicitly not using code that seems like bloat (`[OutputTarget]Spec`), organizing the folder/file structure (a bit, not too much since that would make an already hard code-review even harder, example would be not renaming everything to `runner`), and providing interfaces to develop applications "fast" (relative to our previous approach, this is likely not a final design for how we will empower users to write demos/apps).

Additionally, we try to formalize application design to help move our code towards (specifically-towards, this is not a final design) a simple contract between FlashDreams [io, runtime, presentation, ...]<->Model. This is to better enable a work flow of "Model makers make model" and "Demo makers make demo" (providing initial structure for a general framework where any properly demo can be run on a compatible model capable of requisite input/output modalities.

Core of the Application framework is as follows:
* `IOFactory` Generates OutputSink & InputHandler object per `Output Target` (LocalWindow, WebRTC, MP4, ...). These objects are protocols that focus on giving applications an interface for sending outputs and receiving inputs in a not platform/presentation-engine specific way.
*  `IFlashdreamsApplication` is a prototype for the 'global state' of a world-model/program (which in most cases is almost everything we need to launch a demo application).
* `IFlashdreamsApplicationSession` is a prototype for the 'per step state' of a world-model/program.
* Demos implementations for a particular model integration must provide a `createApp` method that produces an impl of `IFlashdreamsApplication`. This is our entry-point contract.


integrations/causal_forcing/causal_forcing/t2v/app.py
```diff
@@ -0,0 +1,52 @@
...
+            defaults=T2VApplicationDefaults(
+                pipeline_config=PIPELINE_WAN21_T2V_1PT3B_CHUNKWISE,
+                total_blocks=60,
+                pixel_height=480,
+                pixel_width=832,
+            )
+        )
+
+
+def createApp() -> IFlashDreamsApplication:
```
csyonghe: naming convention should be `create_app()`?

integrations/causal_forcing/causal_forcing/t2v/app.py
```diff
@@ -0,0 +1,52 @@
...
+
+"""Causal-Forcing text-to-video application factory."""
+
+from t2v import (
+    T2VApplication,
+    T2VApplicationDefaults,
+    T2VApplicationSession,
+)
+
+from causal_forcing.config import PIPELINE_WAN21_T2V_1PT3B_CHUNKWISE
```
gtong-nv: To integrate a new model, we still need to define the config.py just like before, right? 
`integrations/causal_forcing/causal_forcing/config.py` we don't consider this hard-coded?

jarcherNV:
Copying my slack comments here as a list of things to fix before merging (some of these would be critical):

- Model init and model stepping can run on different threads, this should happen on the same thread
- Cleanup in _run_session leaks on the first failure
- Generation failures are swallowed
- Model setup now happens inside the SDP offer handler, expect ICE and HTTP timeouts on a cold start. This is a behavioral regression against the preload path
- The new bridge blocks model-thread writes, and treats a transient stall as fatal
- Flow control moved out of the host and into every application, not necessarily bad
- It adds a second CLI rather than unifying, not sure if this matters

ArielG-NV:
(rebase on top of main was done - I will still address review)

flashdreams/flashdreams/serving/webrtc/manager.py
```diff
@@ -119,6 +136,290 @@
...
+
+    def _create_initialized_session(
+        self,
+    ) -> tuple[
+        IFlashDreamsApplicationSession,
+        SessionInfo,
+        CanonicalInputSchema,
+    ]:
+        application, slug_args = create_application(self._application_slug)
+        application.init([*slug_args, *self._commandline_args])
```
gtong-nv: hmm, webRTC manager calls the the `create_application` and owns the `session`, I guess this is intended?

integrations/causal_forcing/causal_forcing/t2v/app.py
ArielG-NV: This is fine "for now", but we likely will want to expand/clean-up the definition of model+checkpoint+model_checkpoint_args later on

flashdreams/flashdreams/serving/webrtc/manager.py
ArielG-NV: This is a mistake and requires fixing.

apps/t2v/t2v.py
```diff
@@ -0,0 +1,289 @@
...
+
+    output_layout: VideoTensorLayout
+    """Layout emitted by the integration pipeline."""
+
+
+class T2VApplication(IFlashDreamsApplication):
+    """Reusable text-to-video application configured by one integration."""
+
+    session_type: type["T2VApplicationSession"]
+
```
gtong-nv, line 122: There is no custom `T2VApplicationSession` subclasses. might just remove `session_type` and just use `T2VApplicationSession`

ArielG-NV: We are providing infrastructure for others to impl T2V support.

jarcherNV:
More comments on stuff, these are from claude so may or may not be valid:

- backpressure_s was removed from OutputDecision but the realtime driver still reads it, so every existing realtime session now errors on its first step.
- The test covering that line was deleted, which is why nothing catches it.
- Backpressure is applied by sleeping on the model thread inside the WebRTC sink, so the clock never sees lag.
- Application steps still produce no metrics, and bridge drops are recorded nowhere.
- The wait for a browser peer has no timeout, so a launched model can hang indefinitely.
- WebRTC is output-only, so no interactive model can use this transport yet.
- The new WebRTC manager still duplicates preload, liveness, and admission from the existing one.
- Each integration's app still hardcodes total_blocks and the pixel dimensions instead of reading its model config.
- Thread affinity is correct only because the run is synchronous, not because anything enforces it.

apps/t2v/t2v.py
```diff
@@ -0,0 +1,327 @@
...
+
+T2VApplication.session_type = T2VApplicationSession
+
+
+__all__ = [
+    "T2VApplication",
+    "T2VApplicationDefaults",
+    "T2VApplicationSession",
+]
```
gtong-nv, line 327: should the `T2VApplication` or the derived integration apps provide the web css/js adapter? Looks there is no "generate" or "prompt" buttons/text field in the UI. Not sure how to test the webRTC path

ArielG-NV: it auto runs, this means apps that do not need input from webrtc to work, will work.

ArielG-NV: To provide a prompt/generate we have no api design for this yet. Would need to be part of `IFlashdreams*` somehow?

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/455/commits/612a51ee403603ff4509fe2d23b288d1a82fffec

jarcherNV:
Comments from me interrogating claude about this (it thinks some issues are blocking, but they will just have to be fixed later I guess). Claude could be wrong about these, but we should investigate them later.

Would still block on

- Run mode is chosen by isinstance on the I/O factory, so local window runs on the batch driver while output-only T2V gets the realtime one.
- The realtime clock is a stub whose wait_until_window_end and catch_up do nothing, so realtime is a batch loop plus a sleep.
- finalize()'s return value is discarded, so StepResult.metrics is always empty and per-step stats vanish.

Non-blocking

- The app path skips the run-modes layer, so there is no capability validation, warmup plan, admission, or benchmark sink.
- run_application throws away the metrics snapshot it just collected.
- A sink close failure is reported as a completed run with zero artifacts and exit code 0.
- Cosmos2T2VRunnerConfig gained a meaningless --total-blocks flag purely to satisfy the app's defaults reader.
- One binary now has two flag grammars, depending on whether the slug is a runner or an app.
- Smaller items: session_type and the createSession alias survive, and apps/t2v makes slangpy and the serving stack hard dependencies of every T2V integration.

jarcherNV (approved):
LGTM

Great work!!!!

I think the remaining issues should not block this. We can fix those up later.

Issue #443, opened 2026-08-11
feat: MiniMax-H3 support


yamada-sexta:
MiniMax AI has released **MiniMax H3**, an open-source 33B omni-modal generative system capable of joint video and synchronized 32 kHz stereo audio generation. Adding support for MiniMax H3 to `flashdreams` would allow users to leverage native text-to-video-audio (T2VA), image-to-video-audio (FL2VA), and reference-based video editing/generation (Ref2VA) within the framework.

It is one of the best open weight omni-models currently, and I believe it  would be really nice if flashdreams can support it. I am willing to work on the port.

The possible issue might be related to audio generation, because to my knownledge flashdreams doesn't have an audio pipeline yet.

PR #457, opened 2026-08-13
Add MiniMax H3 integration

yamada-sexta:
As mentioned here: https://github.com/NVIDIA/flashdreams/issues/443

This commit adds MiniMax H3 integration to flashdreams.


yamada-sexta:
Tested with both the  regular [MiniMax H3](https://github.com/MiniMax-AI/MiniMax-H3) and the the [Looping-Sketch LoRA](https://huggingface.co/Inner-Reflections/MiniMax-H3-Looping-Sketch-Anime) on A100.

No audio support yet.

integrations/minimax_h3/minimax_h3/model.py
```diff
@@ -0,0 +1,200 @@
...
+            audio_indices=state.audio_indices,
+            text_indices=state.text_indices,
+        )
+        for index, (video_timestep, audio_timestep) in enumerate(
+            zip(video_timesteps, audio_timesteps, strict=True)
+        ):
+            print(
+                f"MiniMax H3 denoise step {index + 1}/{len(video_timesteps)}",
+                flush=True,
+            )
```
ArielG-NV: Prefer to use `loguru` for logging.

integrations/minimax_h3/minimax_h3/model.py
```diff
@@ -0,0 +1,200 @@
...
+        return (
+            rows.reshape(
+                -1,
+                channels,
+                state.num_latent_frames,
+                state.latent_height,
+                state.latent_width,
+            )
+            .contiguous()
+            .cpu()
```
ArielG-NV: Why are we copying the data here to the CPU?

Looks like we should keep it on what-ever device currently has the data since tracking the var through the code, it looks like we have a branch in use:
1. save to disk via `_save_latents`, calling `latents.detach().cpu().contiguous()` anyways.
2. ensure data is (and copy if not) on CUDA device via `self._decode_video(cache, latents)`. Code ensuring the latents are on the GPU: `state.set("latents", latents.to("cuda"))`

integrations/minimax_h3/minimax_h3/pipeline.py
```diff
@@ -0,0 +1,778 @@
...
+            cache.resumed_stage = "decode"
+            latents = _load_latents(cache, self.config.model_id)
+        else:
+            denoise_started = time.monotonic()
+            if cache.low_ram:
+                latents = self._generate_low_ram(cache)
+            else:
+                latents = self._generate_standard(cache)
+            cache.denoise_seconds = time.monotonic() - denoise_started
+            _save_latents(cache, self.config.model_id, latents)
```
ArielG-NV: we could make this async since otherwise we are blocking `generate` just to generate a cache file we don't immediately need every time we have a new cache?

integrations/minimax_h3/minimax_h3/pipeline.py
```diff
@@ -0,0 +1,778 @@
...
+    def _apply_lora(self, transformer: Any, cache: MiniMaxH3PipelineCache) -> None:
+        if cache.lora is None:
+            return
+        converted = load_lora(
+            transformer,
+            cache.lora,
+            cache.lora_scale,
+            cache.lora_weight_name,
+        )
+        print(f"Loaded LoRA {converted} at scale {cache.lora_scale:g}", flush=True)
```
ArielG-NV: Prefer to use `loguru` for logging.

ArielG-NV (commented):
questions/suggestions

integrations/minimax_h3/minimax_h3/model.py
yamada-sexta: I have updated this and the other ones

integrations/minimax_h3/minimax_h3/model.py
yamada-sexta: Sorry, that was initially a part of the attempt to reduce RAM usage. But you are right, that was unnecessary. I have corrected it.

yamada-sexta:
I tried to use the built in RoPE, but H3 precomputes cos and sin once, shaped [S, 96], then reuses them across its transformer blocks.

H3 rotates only the first 96 channels of each 128-channel attention head; the final 32 channels pass through unchanged.

Therefore, I cannot use the current RoPE effectively. I wonder if you guys are open to changes in the RoPE kernel?

ArielG-NV:
Making changes like adding to the core of flashdreams so that future development of world models is easier for future devleopers is very much encouraged.

This would include changes like adding functionality (knobs for performance) or making alternative implementations of RoPE.

yamada-sexta:
I tested my integration and it seems that the RoPE changes were  not really needed.

PR #465, opened 2026-08-17
Route batch applications through run_demo_session

jarcherNV:
Add the application bridge and IOFactory run mode for null, MP4, and local-window outputs.

Preserve canonical input windows and I/O lifecycle ordering, add truthful output specifications, and migrate application coverage to the public run_application entry point.


ArielG-NV:
I think the tests added in this PR are quite fragile and should just be omitted until we are done the full porting of the new system (they will start failing once we remove old and related code like `OutputSpec`), otherwise the PR is likely fine as is.

ArielG-NV (approved):
we will just remove the tests later I suppose...

PR #485, opened 2026-08-18
Clarify agent coding and architecture guidance

wlewNV:
Define simple engineering principles and canonical execution paths so agents extend one coherent architecture instead of copying legacy bypasses.


AGENTS.md
```diff
@@ -1,85 +1,131 @@
...
+
+## Documentation and Comments
+
+- Every public method needs a clear docstring describing behavior, arguments,
+  return value, and important failure modes.
+- Private methods need concise docstrings when their purpose is not obvious.
+- Prefer plain language over architecture jargon.
+- Explain invariants and non-obvious constraints, not line-by-line mechanics.
+- Keep docs, examples, and launch commands synchronized with real entry points.
+- Use the repository SPDX header on new source files.
```
ArielG-NV: we only need the first 2 lines.

ArielG-NV (changes requested):

AGENTS.md
wlewNV: Addressed in latest revision

PR #488, opened 2026-08-19
v2 rewrite

ArielG-NV:
protocols + tests + basic core


flashdreams/api_v2/input_handler.py
jarcherNV, line 3: We should make the license stuff consistent now, only the first 2 lines are needed for the license info, everything else is covered by the licenses file in the workspace and can be handled there. No need to include the extra stuff.

Edit: Apparently CONTRIBUTING.md mandates the full boilerplate. I think our internal guidance is to not do that. I will leave it to others to decide which is best, but I see a mix in this PR so we should pick which we prefer and stick to it.

flashdreams/api_v2/input_handler.py
```diff
@@ -13,8 +13,29 @@
...
-from .app import create_app
+from abc import abstractmethod
+from typing import Protocol, runtime_checkable
 
-__all__ = ["create_app"]
+from flashdreams.core_v2.user_input_events import UserInputEvents
+
+
+@runtime_checkable
+class InputHandler(Protocol):
```
jarcherNV: We should probably be including a lot of documentation from now on, probably more detailed than what is present here.

flashdreams/api_v2/input_handler.py
```diff
@@ -13,8 +13,29 @@
...
+    """Provide time-windowed canonical input state to the host."""
+
+    @abstractmethod
+    def open(self) -> None:
+        """Enable reading."""
```
jarcherNV: As mentioned above, since we are taking our time now, we should give more informative comments about this stuff, including examples of its intended usage like where/how it is expected to be used, same as for the general class stuff above. I guess this comment applies generally to all the classes etc here.

flashdreams/api_v2/README.md
```diff
@@ -0,0 +1 @@
+File of protocols for the FlashDreams API.
```
jarcherNV: This file needs updating with documentation.

flashdreams/core_v2/session_info.py
```diff
@@ -0,0 +1,57 @@
...
+import math
+from collections.abc import Mapping
+from dataclasses import dataclass, field
+from types import MappingProxyType
+
+from flashdreams.core_v2.video_tensor import VideoTensorLayout
+
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class SessionInfo:
```
jarcherNV: I guess this is largely a copy of what we have in io.py. Seems reasonable to me. I'd suggest lowering the default values though, at least for fps, I don't think we have any models we would expect to run at 60 fps at 720p.

integrations_v2/null_model/null_model/decoder.py
```diff
@@ -0,0 +1,54 @@
...
+
+from torch import Tensor
+
+from flashdreams.infra.decoder import (
+    DecoderConfig,
+    StreamingDecoder,
+    StreamingDecoderCache,
+)
+
+
```
jarcherNV, line 17: I am wondering if these null decoders and configs, encoders etc should be defined here or not. Maybe it makes sense for the null integration, but if we want this to be the template followed by other integrations then it may not make sense, since we would not want the other integrations to have to define their own stuff (ideally, flashdreams infra would support it).

Perhaps instead of this approach, we define a generic identity decoder/encoder etc elsewhere, which can then just be used in situations like this one. Then people who need to debug their integrations could just switch to null or even better identity versions if they need to, run simple verification tests etc.

I am open to being told I am wrong here, I am not sure how useful an "identity" version of these things would be in practice. I have heard that an identity decoder can be useful for isolating the previous steps or inspecting raw latents. It may also allow for better testing in future since it could output something rather than nothing.

Same for the below stuff too.

integrations_v2/null_model/null_model/config.py
```diff
@@ -0,0 +1,38 @@
...
+
+from flashdreams.infra.diffusion.model import DiffusionModelConfig
+from flashdreams.infra.diffusion.scheduler import FlowMatchSchedulerConfig
+from flashdreams.infra.pipeline import StreamInferencePipelineConfig
+
+from .decoder import NullDecoderConfig
+from .encoder import NullInputEncoderConfig
+from .transformer import NullTransformerConfig
+
+NULL_MODEL_CONFIG = StreamInferencePipelineConfig(
```
jarcherNV: Since this integration is essentially a "guide" on how to write integrations and how they work, we should include comments on what this stuff is doing, why, etc.

integrations_v2/null_model/null_model/__init__.py
jarcherNV, line 2: Not a comment on this file, but instead a more general one. How do we want to organise the testing of these integrations? I think I suggested in the previous call that we do a separate test file for each integration which can test all of an integrations capabilities to fully verify it. Do we want to proceed with that idea?

My preference would be if it lived with the integration itself rather than a giant test folder, but it doesn't have to live in the null_model folder itself, it could be in something like integrations_v2/null_model_tests or something.

*Pasting this back as I accidentally deleted it earlier.

flashdreams/test_v2/test_io_factory.py
```diff
@@ -0,0 +1,133 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""CPU test for the v2 I/O factory protocol."""
```
jarcherNV: This test file seems fine, but I think my preference would be not to have all of the api_v2 tests in a single test_v2 folder. Perhaps they can be broken up into sub-folders? Since we should expect the test count to grow over time, we may as well account for it early.

flashdreams/core_v2/time_window.py
```diff
@@ -0,0 +1,29 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Time-window implementation."""
+
+import math
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class TimeWindow:
```
jarcherNV: I guess this is ultimately a copy of what is in infra/time.py. That seems fine, but we will need to remember what must be cleaned up from there, if we don't then we will end up with 2 copies of stuff like this.

For all these classes that are either copies or heavily derived from other stuff, we should make a note of where it came from for now, either in comments or an inventory file somewhere. We know we want to delete the entire integrations folder, but it is less clear what needs to be kept and re-organised from places like infra, since code further down directly uses stuff from it rather than copying.

flashdreams/core_v2/time_window.py
```diff
@@ -0,0 +1,29 @@
...
+    """Half-open time window in seconds since session start."""
```
jarcherNV: The comments are also the same as those in infra/time.py, like with the other classes, we should make them a bit more useful.

flashdreams/flashdreams/runtime_v2/user_input_event.py
```diff
@@ -0,0 +1,104 @@
...
+        return self.value
+
+    def __hash__(self) -> int:
+        return hash(self.value)
+
+    def __eq__(self, other: object) -> bool:
+        return self.value == other.value
+
+@dataclass(frozen=True, slots=True)
+class UserInputEvent:
```
jarcherNV, line 103: The purpose of this file won't be immediately obvious to anyone looking at it at a glance and the comment doesn't help too much. Perhaps a more general comment at the start of this file explaining how all the UserInputEvent stuff is intended to work.

flashdreams/core_v2/user_input_event.py
```diff
@@ -0,0 +1,104 @@
...
+class UserInputEventType(Enum):
+    """User input event type."""
+
+    KEYBOARD = "keyboard"
+    MOUSE = "mouse"
+    TOUCH = "touch"
+    GAMEPAD = "gamepad"
+    WHEEL = "wheel"
+    MOTION = "motion"
+    UNKNOWN = "unknown"
```
jarcherNV: integrations may need their own new event types, and we already have one very obvious use case that is not covered here.

I do not think UNKNOWN should be the mechanism by which integrations do this. Especially since an integration may need to handle more than 1 type. I don't think an integration can cleanly extend this enum since I think python will treat this as a closed set that can't be extended.

Is the intent that integrations define their own UserInputEventType, if so, they would also need to include any defaults already provided here, which doesn't feel like a good solution.

runtime/inputs.py ultimately solved this by just allowing any kind of event, as a string. Perhaps that is not a better solution. What do others think about what should be done here?

flashdreams/api_v2/README.md
jarcherNV: I am not sure how detailed we are planning on making these readme files. If we want them to have lots of detail, perhaps a bunch of my other suggested documentation changes could be made here, and more lightweight stuff goes in the code comments.

flashdreams/test_v2/test_io_factory.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    input_handler.update_input(
+        UserInputEvents(TimeWindow(0, 1), [unknown_input])
+    )
+
+    input_handler.open()
+    pipeline = NULL_MODEL_CONFIG.setup().to("cpu")
+    cache = pipeline.initialize_cache()
+    current_inputs = input_handler.current_inputs()
+    input_handler.close()
```
jarcherNV: I guess input is retrieved, but we never prove it affects the output. I guess that is because right now for the null model, the input cannot affect the output. We will probably want to update this to add that capability and test for it.

I think these can raise exceptions too, right? So it might be useful to at least handle them, and later add tests that deliberately trigger and check for them.

flashdreams/flashdreams/runtime_v2/session_desc.py
```diff
@@ -0,0 +1,57 @@
...
+    """frame rate to display ui-thread frames at."""
+
+    frames_per_second_for_step: int = 30
+    """frame rate to display model-generation thread frames at."""
+
+    video_width: int = 720
+    """Output video width in pixels."""
+
+    video_height: int = 1280
+    """Output video height in pixels."""
```
jarcherNV, line 32: I think this is a portrait shape rather than landscape. All our models use landscape, so really it should be width = 1280 and height = 720.

flashdreams/api_v2/input_handler.py
```diff
@@ -13,8 +13,29 @@
...
+        ...
+
+    @abstractmethod
+    def current_inputs(self) -> UserInputEvents:
+        """Return the latest user input events."""
```
jarcherNV: Is this intended to drain all events since the last call or just give a slice? I assume it is intended to drain, but the comment does not say.

flashdreams/core_v2/session_info.py
jarcherNV: Now that I think about it, does SessionInfo as a name make sense for this? It is not really coupled with a webrtc session, which may be confusing for others looking at the code. Maybe that is ok though assuming the intended usage is something like:

Integration initializes a model run -> produces SessionInfo -> WebRTCSession uses SessionInfo to configure its stuff.

That kind of thing should be described in the comments.

flashdreams/api_v2/io_factory.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Protocol, runtime_checkable
+
+from flashdreams.core_v2.session_info import SessionInfo
+
+from .input_handler import InputHandler
+from .output_sink import OutputSink
+
+
+@runtime_checkable
+class IOFactory(Protocol):
```
jarcherNV: Does IOFactory need to exist at all? The run only needs the actual objects like run(pipeline, input_handler, output_sink), it does not need to care how they were constructed.

Perhaps we can simplify this to just be something like:

```
input_handler = NullInputHandler()
output_sink = Mp4OutputSink(session_info)
run(pipeline, input_handler, output_sink)
```

Wheras with the factory approach, you would need to do something like:

```
factory = LocalWindowIOFactory(session_info)
input_handler = factory.create_input_handler()
output_sink = factory.create_output_sink()
run(pipeline, input_handler, output_sink)
```

Note that in v1 the factories were separate, like Mp4IOFactory, LocalWindowIOFactory, etc, so the factory kind of has to match the sink, which is a factory per sink when you could just make the sink directly, and the factories carry very little actual logic so they don't really add anything. I suggest we just make the sink directly with no factory abstraction.

You could even simplify it further to something like:

```
create_io(session_info) -> (input_handler, output_sink)
run(pipeline, input_handler, output_sink)
```

Having the above function would also allow you to create them together, in case you need to bridge the input and output, such as the local window in v1.

*Edit: comment was heavily edited by me, and I realised something after writing it all:

In addition to getting rid of the factory, I'm not sure InputHandler should be user-facing at all. Nobody ever actually picks an input handler (why do you need to pick both LocalWindowInputHandler and LocalWindowOutputSink since you don't get a choice here). What they pick is a transport, browser, local window, etc. That's why the input handler and output sink feel coupled even though neither protocol references the other.

Which makes me wonder whether the user-facing choice should be the transport rather than the two edges:

```
transport = LocalWindow(title="...")   # or WebRTC(port=8080), or Mp4(path)
run(pipeline, transport)
```

Two things probably worth preserving if we went that way, v1's CompositeOutputSink lets a single run write an mp4, show a window and record benchmark stats all at once, and being able to inject a scripted input handler for deterministic tests.

flashdreams/test_v2/test_io_factory.py
```diff
@@ -0,0 +1,133 @@
...
+        metrics={},
+    ))
+    output_sink.close()
+
+
+    assert current_inputs.get_events() == [unknown_input]
+    event_data = current_inputs.get_events()[0].get_event_data()
+    assert isinstance(event_data, UserInputEventDataUnknown)
+    assert event_data.data == 1
+    assert output.shape == (1, 3, 1, 1, 1)
```
gtong-nv: Can we make the output depend to the input? Right now it seems this will always be true.

flashdreams/flashdreams/runtime_v2/session_desc.py
```diff
@@ -0,0 +1,57 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
gtong-nv, line 1: if we put these "runtime" stuff in `flashdreams/core`, should we consider rename `flashdreams/flashdreams/core` to something else, or will move those inferencing building blocks to here after the rewrite?

flashdreams/test_v2/test_io_factory.py
jarcherNV: Yes I think right now this will always be true.

flashdreams/api_v2/io_factory.py
jarcherNV: After thinking about this a little more regarding my above comment, I suggest we just dump the factories. I don't think they add anything useful.

@gtong-nv @wlewNV @csyonghe what do you all think? (see my above comment for context).

*Edit to copy over the important part:

In addition to getting rid of the factory, I'm not sure InputHandler should be user-facing at all. Nobody ever actually picks an input handler (why do you need to pick both LocalWindowInputHandler and LocalWindowOutputSink since you don't get a choice here). What they pick is a transport, browser, local window, etc. That's why the input handler and output sink feel coupled even though neither protocol references the other.

Which makes me wonder whether the user-facing choice should be the transport rather than the two edges:

```
transport = LocalWindow(title="...")   # or WebRTC(port=8080), or Mp4(path)
run(pipeline, transport)
```

Two things probably worth preserving if we went that way, v1's CompositeOutputSink lets a single run write an mp4, show a window and record benchmark stats all at once, and being able to inject a scripted input handler for deterministic tests.

flashdreams/core_v2/time_window.py
```diff
@@ -0,0 +1,29 @@
...
+
+    start_s: float = 0
+    end_s: float = 0
```
gtong-nv: Can you explain how the `TimeWindow` will be used?
It's used in `UserInputEvents`, which I assume would be a list of UserInputEvent within a specified `TimeWindow`, held by `InputHandler`.

In the test

```
    input_handler.update_input(
        UserInputEvents(TimeWindow(0, 1), [unknown_input])
```
means the `InputHandler` receives the 1 `unknown_input` from 0 to 1 second, and the `unknown_input` has timestamp at 0 second, so the future usage of this input is valid, otherwise it should be filtered out (e.g if the `TimeWindow` is (1,2)) ?
should we add the filter logic in `update_input`?

flashdreams/test_v2/test_io_factory.py
```diff
@@ -0,0 +1,133 @@
...
+    unknown_input = UserInputEvent(
+        timestamp=0,
+        event_type=UserInputEventType.UNKNOWN,
+        event_data=UserInputEventDataUnknown(data=1),
+    )
+    input_handler.update_input(
+        UserInputEvents(TimeWindow(0, 1), [unknown_input])
+    )
+
+    input_handler.open()
```
gtong-nv: what will the `input_handler` do in the `open()` func? what is it called after `update_input`?

flashdreams/test_v2/test_io_factory.py
```diff
@@ -0,0 +1,133 @@
...
+    output_sink = factory.create_output_sink()
+    assert isinstance(output_sink, OutputSink)
+
+    unknown_input = UserInputEvent(
+        timestamp=0,
+        event_type=UserInputEventType.UNKNOWN,
+        event_data=UserInputEventDataUnknown(data=1),
+    )
+    input_handler.update_input(
+        UserInputEvents(TimeWindow(0, 1), [unknown_input])
```
gtong-nv: I guess in real life, this will happen in the UI step?
Probably not in the scope of this PR, but would be good to make it run for a few frame at the rate `frames_per_second_for_ui=1` defined in the session

flashdreams/api_v2/io_factory.py
csyonghe: I would still define separate protocols for InputHandler and OutputSink, and then have `IPresentationSurface : InputHandler, OutputSink`. And then we have three types of presentation surfaces. You can pass the single presentation surface object to either a InputHandler or a OutputSInk parameter.

flashdreams/test_v2/test_io_factory.py
csyonghe: I don't quite understand why we need `InputHandler.open`.

flashdreams/test_v2/test_io_factory.py
csyonghe: This method is really weird, why is `update_input` a member method of `input_handler`? This is supposed to be a place where you obtain/query user input, not a place where you write/put input events into.

flashdreams/api_v2/io_factory.py
jarcherNV: Yes that was my thinking, we can have the separate InputHandler and OutputSink protocols, and then instead of factories, we just have a LocalWindow or WebRTC etc.

flashdreams/api_v2/input_handler.py
csyonghe: I don't understand why you mean by "time-windowed" , "canonical" etc. No undefined terminology in comments.

flashdreams/api_v2/input_handler.py
csyonghe: It needs to document the contract. Who is supposed to call what, at when. What should an implementation of the protocol do during each method etc. Here I don't quite get why would we need `open`.

flashdreams/api_v2/input_handler.py
csyonghe: if this is a method, then there needs to be a verb in the method name.

flashdreams/core_v2/session_info.py
ArielG-NV: models run at 30; 60 is ui thread. UI should run at a "smooth" frame-rate.

flashdreams/api_v2/io_factory.py
csyonghe: I agree this feels unnecessary.

flashdreams/core_v2/session_info.py
```diff
@@ -0,0 +1,45 @@
...
```
csyonghe: should this be called `SessionDesc` instead, since it is a descriptor to create a session, not something you obtain from an object after creating a session.

integrations_v2/null_model/null_model/decoder.py
ArielG-NV: Added the basic guide

flashdreams/core_v2/time_window.py
csyonghe: second feels like a weird unit. Should be `uint64 time_start`.

flashdreams/core_v2/time_window.py
csyonghe: Not sure why events have a TimeWindow. An event happens at a point of time, not over a range of time, so to me we only need a single time_stamp.

flashdreams/core_v2/user_input_event.py
```diff
@@ -0,0 +1,92 @@
...
+from dataclasses import dataclass
+from enum import Enum
+from typing import Protocol
+
+
+class UserInputEventData(Protocol):
+    """User input event data."""
+    pass
+
+class UserInputEventDataKeyboard(UserInputEventData):
```
csyonghe: This is wrong naming. Should be `KeyboardUserInputEventData`.

flashdreams/core_v2/user_input_event.py
```diff
@@ -0,0 +1,92 @@
...
+class UserInputEventDataUnknown(UserInputEventData):
+    """User input event data for unknown."""
+
+    data: object
+    """Opaque application-defined input data."""
+
+class UserInputEventType(Enum):
+    """User input event type."""
+
+    KEYBOARD = "keyboard"
```
csyonghe: Is ALLCAPS the pythonic naming convention for enums?

ArielG-NV: I think it is: https://docs.python.org/3/howto/enum.html

image

flashdreams/core_v2/user_input_event.py
ArielG-NV: I will make the changes.

flashdreams/core_v2/time_window.py
jarcherNV: I think this was needed in v1 to handle user input events over a range of time, like input can be held, not just pressed. Hold W for two seconds and you get one key_down followed by several batches with no events in them at all, and v1's keyboard-to-camera conversion uses the window to emit a segment covering those empty batches so the camera keeps moving.

None of that machinery is in v2 yet so this stuff probably feels a bit out of place right now.

ArielG-NV: Still unneeded, `key-A-down at 0` `key-A-up at 1` is a more expressive way to describe `A is pressed from 0 to 1`

jarcherNV: Either way, since it is not needed for any of this stuff, we can just get rid of this file from v2 and worry about that stuff later.

flashdreams/api_v2/input_handler.py
ArielG-NV: ~I don't think this system makes sense as a whole... why are we not just registering an event handler...~

integrations_v2/null_model/null_model/decoder.py
jarcherNV: I assume that reply was intended for my other comment. But I wasn't suggesting we write a full guide, rather comment it well so it can be easy to follow.

flashdreams/core_v2/time_window.py
ArielG-NV: >  but we will need to remember what must be cleaned up from there

We will need to delete the entire old system and related code, yes.

flashdreams/api_v2/input_handler.py
jarcherNV: Now that I think about it, the InputHandler was only ever used in v1 for the SlangPyWindow and NullInput. Do we know that this system can work with webrtc? @gtong-nv do you know?

ArielG-NV: The idea would be: Web publishes inputs to buffer

This buffer is where we pull "UserInputEvents` from.

Assuming this is how things work, we don't need open/close for input handler since... well... the "getter" would return an empty list of inputs.

integrations_v2/null_model/null_model/decoder.py
ArielG-NV: > I am open to being told I am wrong here, I am not sure how useful an "identity" version of these things would be in practice. I have heard that an identity decoder can be useful for isolating the previous steps or inspecting raw latents. It may also allow for better testing in future since it could output something rather than nothing.

What would it even mean to have an identity encoder? sounds like just a model integration that passes data through in a very simplistic fashion.

integrations_v2/null_model/null_model/__init__.py
ArielG-NV: I think putting the test in the `integrations_v2/<model>/tests/**` folder makes sense since we still plan for each integrations package to sync with their unique dependencies before running any demo on them, among other things

flashdreams/flashdreams/runtime_v2/session_desc.py
ArielG-NV: This comment is not actionable in its current state.

I could rename it, but to what? `flashdreams/runtime_v2`?

integrations_v2/null_model/null_model/__init__.py
ArielG-NV: Added basic smoke test

flashdreams/core_v2/time_window.py
ArielG-NV: Not sure if this is actionable right now.

flashdreams/api_v2/README.md
ArielG-NV: Not sure if this is actionable right now. This seems like a "future plan"?

I personally think having a generic "how things work" doc is good, but unsure if our code encapsulates this right now.

flashdreams/test_v2/test_io_factory.py
ArielG-NV: > I guess input is retrieved, but we never prove it affects the output. I guess that is because right now for the null model, the input cannot affect the output. We will probably want to update this to add that capability and test for it.

Fixed.

> I think these can raise exceptions too, right? So it might be useful to at least handle them, and later add tests that deliberately trigger and check for them.

implementation decision of individual target, irrelevant for the API.

flashdreams/api_v2/input_handler.py
ArielG-NV: > Is this intended to drain all events since the last call or just give a slice? I assume it is intended to drain, but the comment does not say.

This is an API, we cannot drain anything; we can *suggest* that a user may drain so users do not assume that draining will/will-not occur but anything more requires a private `_do_thing` method that wraps the API call on an internal-system level.

ArielG-NV: Added documentation specifying the possibility.

flashdreams/test_v2/test_io_factory.py
ArielG-NV: implemented

flashdreams/core_v2/session_info.py
ArielG-NV: changed to sessionDesc

ArielG-NV:
Note: I have been pre-emptively resolving code-review threads due to large-count once comment has been addressed

flashdreams/test_v2/test_io_factory.py
ArielG-NV: no more `open`. `update_input` is removed.

csyonghe (commented):
where is the prototype for application/session?

flashdreams/test_v2/test_io_factory.py
ArielG-NV: > I guess in real life, this will happen in the UI step?
> Probably not in the scope of this PR, but would be good to make it run for a few frame at the rate frames_per_second_for_ui=1 defined in the session

Added

> This method is really weird, why is update_input a member method of input_handler? This is supposed to be a place where you obtain/query user input, not a place where you write/put input events into.

Changed so iofactory handles it (I know we plan to remove iofactory, but for iterative reasons it is still here while I address review)

integrations_v2/null_model/null_model/tests/smoke.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+
+from null_model import NULL_MODEL_CONFIG
```
csyonghe, line 6: this is fine, but I expect to see a test module that is equivalent to an "app"/"demo", with session/application/create_application functions, and that is dynamically loaded(imported) and run by a test.

ArielG-NV:
> where is the prototype for application/session?

We already have a lot of opinions on this small sub-section of the code-base, where review is already quite abstract sometimes (not exact on actionable changes that this PR can address in isolation).

I could add the other components, but this just means that everyone will be blocked on this review for longer.

csyonghe:
OK, so that need to be part 2 to land after this PR.

flashdreams/flashdreams/runtime_v2/user_input_event.py
ArielG-NV: People can refer to `flashdreams\test_v2\test_io_factory.py` for now. It is a full implementation using the tooling,

flashdreams/core_v2/user_input_event.py
ArielG-NV: changed the system to be more correctly typed such that this is no longer an issue

flashdreams/flashdreams/runtime_v2/session_desc.py
csyonghe: I would be fine with `flashdreams/runtime_v2` for now.

flashdreams/core_v2/user_input_event.py
csyonghe: For now, I am OK with this being an enum, and an application will be limited to the user input types we support.

flashdreams/api_v2/input_handler.py
csyonghe: Then that is an internal method to WebRTCPresentationSurface, not IInputHandler.

ArielG-NV:
> OK, so that need to be part 2 to land after this PR.

Someone else can try to land it right-now as a fork from what I already have (copy-paste the old code and start working on clean-up).

I will make one more "big" change, but there is no need to wait for me to deal with some of the less than exact/specific/clear code-review since that may take a bit more

flashdreams/flashdreams/api_v2/input_source.py
fangjunzhou-nv, line 1: imo this file/class should be named InputSource instead of InputHandler. It's similar to OutputSink where we're not actually handling the input but pulling user event.

The input handler should be an actual "handler" that maps the pulled input to canonical input/inference input/etc.

flashdreams/flashdreams/api_v2/client_window.py
fangjunzhou-nv, line 1: As commented in flashdreams/api_v2/input_handler.py, a presentation surface should have both InputSource and OutputSink.

Also, presentation surface are normally considered output only. I believe what we're defining here is an abstract Application?

flashdreams/api_v2/presentation_surface.py
```diff
@@ -0,0 +1,18 @@
...
+
+from typing import Protocol, runtime_checkable
+
+from .input_handler import InputHandler
+from .output_sink import OutputSink
+from flashdreams.runtime_v2.session_desc import SessionDesc
+
+
+@runtime_checkable
+class IPresentationSurface(InputHandler, OutputSink, Protocol):
```
fangjunzhou-nv: Should we prefer composition over inheritance? We should have and Application that holds an InputSource and an OutputSink

fangjunzhou-nv (commented):
Comments on api v2

flashdreams/flashdreams/api_v2/input_source.py
ArielG-NV: I agree, I think the purpose of the code changed

flashdreams/flashdreams/api_v2/client_window.py
csyonghe: It's the IO perspective of an application specific for user interaction. I can't think of a better name so open to ideas here.

flashdreams/api_v2/presentation_surface.py
csyonghe: Inheritance here is fine. Composition is cleaner but it also brings a lot more boilerplate.

flashdreams/flashdreams/api_v2/input_source.py
csyonghe: Agreed.

flashdreams/flashdreams/api_v2/client_window.py
ArielG-NV: > As commented in flashdreams/api_v2/input_handler.py, a presentation surface should have both InputSource and OutputSink.


this was added

ArielG-NV: > Also, presentation surface are normally considered output only. I believe what we're defining here is an abstract Application?

I've been calling it a PresentationBackend out of habit of "its the backend that owns the surface + IO"

flashdreams/api_v2/presentation_surface.py
ArielG-NV: Most of this code should be ABC rather than protocols, I agree here.

flashdreams/flashdreams/api_v2/client_window.py
ArielG-NV: Seems that online others call the idea `PresentationPlatform`

ArielG-NV: I think for simplicity I will call it PresentationBackend since at the every least presentation surface is wrong, we will be implementing on to of it webtc/...

integrations_v2/null_model/null_model/decoder.py
jarcherNV: An identity would just be something that does not change the inputs given to it, ie, it just passes along the stuff that it is given. If it were used here, we would at least be able to test *something* from this model rather than nothing.

integrations_v2/null_model/null_model/__init__.py
jarcherNV: Before resolving this, I think @gtong-nv @wlewNV @csyonghe should give their thoughts.

integrations_v2/null_model/null_model/decoder.py
jarcherNV: As above, before resolving this, I think @gtong-nv @wlewNV @csyonghe should also give their thoughts on this.

flashdreams/flashdreams/api_v2/client_window.py
ArielG-NV: IClientWindow is the new plan (discussed offline)

flashdreams/api_v2/input_source.py
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Canonical application input handling protocol."""
```
jarcherNV: Some comments still refer to canonical, which is undefined terminology here.

ArielG-NV: I will remove this change

integrations_v2/null_model/null_model/tests/smoke.py
ArielG-NV: We have yet to make the interface for how to run an app, we cannot do this without introducing likely "cruft" as we change things closer to the 'app'

flashdreams/runtime_v2/video_tensor.py
```diff
@@ -0,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Video tensor layout implementation."""
+
+from enum import Enum
+
+
+class VideoTensorLayout(Enum):
```
fangjunzhou-nv: Need a bit more document on the layout.

flashdreams/api_v2/application.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class IApplication(ABC):
+    """Session factory for one application, independent of presentation backend.
+
+    The application is the process-lifetime half of the contract: it parses its
+    own arguments, owns inter-session data (a checkpoint, model weights, a
+    compiled pipeline), and hands out sessions. It outlives every session it
+    creates, so an implementation is free to cache shared state across them and
+    release it in :meth:`close`.
```
ArielG-NV: We should not be using `:meth:` and other annotations like this afaict, it is bad for code-readability (it only helps in regards to linters)

flashdreams/api_v2/application.py
```diff
@@ -0,0 +1,57 @@
...
+
+    This and :class:`ISession` are the whole of what an application module
```
ArielG-NV: `This`?
or `IApplication`?

ArielG-NV: This is a new term "application module"
```suggestion
    This and :class:`ISession` are the whole of what an application
```

ArielG-NV:
/ok to test [b8a2a50](https://github.com/NVIDIA/flashdreams/pull/488/commits/b8a2a5058c056f3e2cd6def01ccbde63ef8e4da8)

jarcherNV (approved):

gtong-nv (approved):

PR #490, opened 2026-08-19
Add v2 IApplication/ISession, a two-thread run_session, and a red screen app

jarcherNV:
Builds on the v2 API from #488 with the halves an application actually implements.

- `IApplication` is process-lifetime and creates an `ISession` per run from a `SessionDesc`, which the session reports back as resolved. Neither frame geometry nor run length is an application concern.
- `run_session` drives one session against one client window on two threads: generation on the caller's thread, and an I/O thread that solely owns the window, reading input, calling `step_ui` and writing finished results. A slow step never holds up input or output.
- Runs are bounded by `steps` or ended by a `CloseUserInputEventData`; a `ResetUserInputEventData` restarts the session and the step index. A bounded queue applies backpressure, either blocking so every result is presented or dropping the oldest to keep latency down.
- `integrations_v2/red_screen` is an end-to-end application exercising all of it: red while a key is held, black otherwise.


gtong-nv (approved):
LGTM

flashdreams/flashdreams/api_v2/input_source.py
```diff
@@ -13,14 +13,21 @@
...
     """
 
     @abstractmethod
     def get_user_input_events(self) -> UserInputEvents:
-        """Return the user input events collected so far.
+        """Return the events that arrived since the previous call.
+
+        Each event is returned exactly once, so an implementation buffers what
+        arrives and clears that buffer here. Returning an event twice makes the
+        caller apply it twice.
```
ArielG-NV, line 28: Clearing is still user defined, we cannot promise this via the contract of only `def get_user_input_events(self) -> UserInputEvents:`

flashdreams/flashdreams/api_v2/output_sink.py
```diff
@@ -6,23 +6,34 @@
...
-    def open(self) -> None:
-        """Prepare to receive results."""
+    def open(self, session_desc: SessionDesc) -> None:
+        """Enable writing for a session that emits ``session_desc``-shaped results.
+
+        Args:
+            session_desc: Output description declared by the session. The sink
+                configures itself from it and may reject a description it
+                cannot present.
+        """
```
ArielG-NV, line 29: If session_desc fails, we should return: false, send an exception, or return the new updated desc which we changed the desc to... sending zero info to developer is likely a bad idea

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -3,4 +3,108 @@ SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All
...
+  held after, because it is the earlier edge that says so. A session that must
+  not inherit that input ignores the older events itself.
+
+Threading
+---------
+
+`run_session` uses two threads, and every window runs that way. Generation is on
+the calling thread; the window gets a thread of its own, ticking at
+`frames_per_second_for_ui` to read input, call `ISession.step_ui`, and write
+finished results. A slow step does not hold up input or output, which is why
```
ArielG-NV, line 60:
```suggestion
finished results. A model generation thread (`step`) does not hold up input or output, which is why
```

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -0,0 +1,279 @@
...
+    max_pending: int = 2,
+    when_full: WhenFull = WhenFull.BLOCK,
+) -> None:
+    """Drive one session against one client window.
+
+    Runs on two threads. The calling thread initializes the session and calls
+    ``step`` for each index, with the input collected since the previous step. A
+    second thread owns the window: it opens it, ticks at
+    ``frames_per_second_for_ui`` to read input, call ``step_ui`` and write
+    whatever generation has finished, then closes it. A slow step therefore does
```
ArielG-NV, line 57:
```suggestion
    whatever generation has finished, then closes it. A model generation thread therefore does
```

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -0,0 +1,279 @@
...
+
+    The window ends the run by reporting a :class:`CloseUserInputEventData`, and
+    restarts it by reporting a :class:`ResetUserInputEventData`, which resets the
+    session and takes the step index back to zero. The window stays open. Nothing
+    from the abandoned generation is presented: each result carries the generation
+    it was produced for, so results already waiting and a step that was still
+    running when the reset arrived are both dropped rather than written.
+
+    Input is not split at a reset: the batch carrying it reaches the first step
+    afterwards whole, earlier events included. Events are edges, so a key held
```
ArielG-NV, line 70: This does not make sense as a sentence. "the batch carrying it reaches the first step
    afterwards whole, earlier events included"

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -0,0 +1,279 @@
...
+            # either belongs to the generation being abandoned or to the new one,
+            # never to neither.
+            if _contains(events, ResetUserInputEventData):
+                generation += 1
+        # Stop from here rather than waiting for the step loop to notice, so a
+        # slow step does not delay a client that has gone away.
+        if _contains(events, CloseUserInputEventData):
+            stop.set()
+        session.step_ui(events)
+        present_pending_results()
```
ArielG-NV, line 165: ~Another Race condition.~

Edit: not a race condition, I realized `events` is a copy.

flashdreams/flashdreams/api_v2/user_input_event_data.py
ArielG-NV, line 29: Looking at how we share across processes the event data, I realize that we need to implement `__hash__` via a hash of `get_type_name`, otherwise we can introduce very big conflicts.

Note: we also don't verify if 2 conflicting `get_type_name` exist.

integrations_v2/red_screen/red_screen/app.py
```diff
@@ -0,0 +1,154 @@
...
+    def __init__(self) -> None:
+        self._config: RedScreenConfig | None = None
+
+    def init(self, commandline_args: list[str]) -> None:
+        """Parse the activation key.
+
+        Neither frame geometry nor rollout length is an application argument: the
+        runtime supplies geometry per session through :class:`SessionDesc`, and
+        decides how long to run when it drives the session.
+
```
ArielG-NV, line 126: What is 'frame geometry', we never define it anywhere, we just start using the phrase here and in documentation (without defining where it comes from or what it is). We should define the term in documentation or not use it.

integrations_v2/red_screen/README.md
```diff
@@ -0,0 +1,80 @@
...
+`--inexact` matters: it stops `uv` from uninstalling the other workspace members
+it was not asked about, which the framework tests import.
+
+## Arguments
+
+| Argument | Default | Meaning |
+|---|---|---|
+| `--key` | `r` | Key whose held state selects red over black. |
+
+Frame geometry comes from the `SessionDesc`, and the step count from the caller
```
ArielG-NV, line 76: 'Frame geometry' is never defined; what part of `SessionDesc` specifies frame geometry?

I suppose this means the width/height, but we should explain this, geometry has more than one meaning and being simple is better than being abbreviated in obtuse language.

ArielG-NV (commented):
comments/questions/suggestions on design

PR #491, opened 2026-08-20
webRTC ClientWindow

gtong-nv:
## Summary
- Add v2 `ApplicationRunner` and shared client-window factory.
- Add a WebRTC client window with buffered browser input and video delivery.
- Add a browser-accessible red-screen demo with keyboard and button controls.
- Package HTML and JavaScript as separate web resources.
- Add CPU tests for lifecycle, window selection, input, and video delivery.
## Workflow
```text
Browser
  │ keyboard/button events
  ▼
WebRTCServer
  │ callback
  ▼
WebRTCClientWindow input queue
  │ UserInputEvents
  ▼
ApplicationRunner → run_session → RedScreenSession.step()
                                  │
                                  ▼
                              StepResult
                                  │
                                  ▼
Browser video ← WebRTCServer ← WebRTCClientWindow.write()

PR #492, opened 2026-08-20
Add MP4 file output and a colour fade app

jarcherNV:
One loop runs every session. A run writing a file is driven by `run_session` against `Mp4ClientWindow`, a window that reports no input and encodes every result through `Mp4OutputSink`. What differs between a run somebody is watching and a run that writes a file is the window it is driven against, not the loop and not the session.

A file run is there to produce an artifact that can be compared, and it gets that from being handed no input and from keeping every result, with `WhenFull.BLOCK`. Since such a window never reports a close, `steps` is what ends the run.

Settles how a window reads pixel values, following the WebRTC window: floats hold `[-1, 1]` and integers hold raw 0-255. `red_screen` emitted `[0, 1]` and now emits `[-1, 1]` to match.

`color_fade` is the smallest end-to-end application on that path: no model, frames whose colour depends only on when they play, and a real MP4 written on CPU.

Based on #491, so the review is the commits on top of `dev/gtong/webrtc-api-2`.


jarcherNV:
I've been on the fence about it but I've decided to get rid of the Mp4ClientWindow, and just use an Mp4OutputSink instead. If people prefer the other way let me know and I will switch it back.

flashdreams/flashdreams/runtime_v2/batch_runner.py
```diff
@@ -0,0 +1,91 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Batch loop generating a fixed number of steps and writing each to a sink."""
```
gtong-nv: Is this file really needed? 
I think the `session.step(step_index..)` should be called only in `session_runner.py` ?

jarcherNV: Oh I thought we decided to have a different main loop for each type. Are we no longer doing that?

ArielG-NV:
> 
I've been on the fence about it but I've decided to get rid of the Mp4ClientWindow, and just use an Mp4OutputSink instead. If people prefer the other way let me know and I will switch it back.

I am under the opinion we want a client window to keep future work all funneling through the same objects.

This especially applies for synthetic inputs and routing them such that it looks like they came from a Client Window equally across all targets

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -7,13 +7,23 @@ Protocols for the FlashDreams API.
 
 - `application.py` / `session.py`: `IApplication` creates an `ISession` from a
   `SessionDesc`, and the session reports what it resolved to.
-- `input_source.py` / `output_sink.py` / `client_window.py`: `IClientWindow` is
-  one client's input and output together. It is given the session's `SessionDesc`
-  in `OutputSink.open`.
+- `input_source.py` / `output_sink.py` / `client_window.py`: an `OutputSink` is
+  where results go, and is given the session's `SessionDesc` in
+  `OutputSink.open`. `IClientWindow` is one client's input and output together,
+  for output that has a client to take input from.
```
ArielG-NV:
```suggestion
  `OutputSink.open`. `IClientWindow` is a producer of input's (populates input source) and consumer of outputs  (output sink target) for a single client.
```

simplicitly

jarcherNV:
> > 
> 
> I've been on the fence about it but I've decided to get rid of the Mp4ClientWindow, and just use an Mp4OutputSink instead. If people prefer the other way let me know and I will switch it back.
> 
> I am under the opinion we want a client window to keep future work all funneling through the same objects.
> 
> This especially applies for synthetic inputs and routing them such that it looks like they came from a Client Window equally across all targets

Yeah I did consider that, I was on the fence about that too. I'll see if it makes sense to have them both.

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -7,13 +7,23 @@ Protocols for the FlashDreams API.
...
 - `user_input_event_data.py`: base type for event payloads.
 
-`flashdreams.runtime_v2.session_runner.run_session` drives a session against a
-window, for a fixed number of steps or until the window reports a close.
+Each kind of output has its own loop, and what they share is the session they
+step rather than the loop that steps it:
```
ArielG-NV:
```suggestion
We provide different methods to running a user implemented session interactively/real-time or non-interactively/batch. These run modes are independent of how a session is implemented code, they are just styles of running a session:
```

sessions produce the same output regardless of if "batch-inference" or "real-time inference" mode, they are not different kinds. 
saying "[each] has their own loop... rather than the loop that steps it" is ambiguous

I propose just saying what we want to get across directly, we have 2 different ways to run a session

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -7,13 +7,23 @@ Protocols for the FlashDreams API.
...
+
+- `flashdreams.runtime_v2.session_runner.run_session` drives a session against a
+  window on two threads, for a fixed number of steps or until the window reports
+  a close. This is the interactive path.
```
ArielG-NV:
```suggestion
- `flashdreams.runtime_v2.session_runner.run_session` drives a session against a
  window driving multiple threads, for a fixed number of steps or until the window reports
  a close. This is the interactive path.
```

We may change the threads running to back a user-app- lets not specify a number so the docs don't rot as fast.

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -7,13 +7,23 @@ Protocols for the FlashDreams API.
...
+- `flashdreams.runtime_v2.batch_runner.run_batch` generates a fixed number of
+  steps on one thread and writes each to an `OutputSink`. This is the path for
+  output that is a file, which has no input side, and `Mp4OutputSink` is that
+  sink. Both runners take a session, so the application is the caller's either
+  way.
```
ArielG-NV: "Both runners take a session, so the application is the caller's either way." we already said this in the preamble before. We can just keep it simple here.

Additionally, if we allow synthetic inputs this would violate the "no input" part of the documentation, we can just not touch on that here since we are quite certain we want synthetic inputs for testing infra-structure.

We may not have one thread when we have a ui-thread-- UI would still be async and we would want the UI to draw on the `batch` runner. Separately, what if we have a physics thread?

When we say "fixed number of steps" we likely are referring to model generation thread

```suggestion
- `flashdreams.runtime_v2.batch_runner.run_batch` generates a fixed number of
  steps for the main model-generation thread and writes to an `OutputSink` (generally `Mp4OutputSink`) which links to a backing file.
```

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -50,19 +60,24 @@ Agreed design decisions. Change them by discussion.
   the new generation whole, so a key held down when the client restarts is still
   held after, because it is the earlier edge that says so. A session that must
   not inherit that input ignores the older events itself.
+- An `OutputSink` reads `StepResult.output` by its dtype: a floating point tensor
+  holds `[-1, 1]`, which is what FlashDreams models emit, and an integer tensor
+  holds raw `0`-`255` values. `SessionDesc` does not carry a range, so this is the
```
ArielG-NV:
```suggestion
- An `OutputSink` reads `StepResult.output` via one of the following: a floating point tensor from `[-1, 1]`, and an integer tensor from [0,255]. `SessionDesc` does not carry a range, so this is the
```

grammer.

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -50,19 +60,24 @@ Agreed design decisions. Change them by discussion.
...
+  convention every sink follows rather than something a session declares.
```
ArielG-NV:
```suggestion
  convention every sink follows rather than something else a session may use.
```

Sessions have no way to declare types they will they output with, they can only "juse use" types.

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -7,13 +7,23 @@ Protocols for the FlashDreams API.
...
-window, for a fixed number of steps or until the window reports a close.
+Each kind of output has its own loop, and what they share is the session they
+step rather than the loop that steps it:
+
+- `flashdreams.runtime_v2.session_runner.run_session` drives a session against a
+  window on two threads, for a fixed number of steps or until the window reports
+  a close. This is the interactive path.
+- `flashdreams.runtime_v2.batch_runner.run_batch` generates a fixed number of
+  steps on one thread and writes each to an `OutputSink`. This is the path for
+  output that is a file, which has no input side, and `Mp4OutputSink` is that
```
ArielG-NV: Sounds like if we integrate UI threading, it may be incompatible with batch as it currently works (1 thread, user code may assume one for ui, one for non-ui, hit deadlock). Alternatively, UI will not render for batch mode... this is an issue as well if users do something like embed the StepResult in their UI window.

We should think about separating UI rendering thread from the window driving thread, I think this is a mistake in hind-sight.

[this would likely be different work] We should allow users to launch what-ever thread they want at ISession definition time (optionally, capture inputs for ImGUI or something via some flag), and we should just make a compositor that works separate from the UI thread.

Maybe in `ISession` creation we have a `Threads` struct that allows adding `IThread` that run along side `step`. `IThread` has a few ways to run:
* `UIThread` runs at `frames_per_second_for_ui`. Outputs `StepResult` since UI may be generated by a model...
* `ModelThread` that allows separate model running in a new thread, outputs `StepResult`.
* When compositing everything we look at the "most recent ready to present frame" and composite them all together, presenting at rate of `frames_per_second_for_ui`
* A StepResult can skip compositing ("disabled", example would be for a complex UI effects or particle system I suppose) via a StepResult flag (compositor sees `StepResult::disabled` as most recent and skips compositing).

this way batch processing will not have "different" results than runtime for testing/batch-processing purposes

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -50,19 +60,24 @@ Agreed design decisions. Change them by discussion.
...
+because a file has no client to take input from or to keep up with. The rest of
+this section is about `run_session`.
 
-A window with no input to report, such as one writing an MP4, returns no events
-and leaves `step_ui` at its default; the threading is unchanged.
+Generation is on the calling thread; the window gets a thread of its own,
+ticking at `frames_per_second_for_ui` to read input, call `ISession.step_ui`,
+and write finished results. A slow step does not hold up input or output, which
+is why `SessionDesc` carries the two frame rates separately. Only the UI rate is
+read so far — generation currently runs as fast as it can.
```
ArielG-NV: What is a slow step?

ArielG-NV: what is the second frame-rate, we only bring up `frames_per_second_for_ui`?

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -71,9 +86,10 @@ locking.
 Writing happens on that thread too, so a window slower than generation leaves
 results waiting. `run_session` bounds how many wait, with `max_pending`, and
 `when_full` decides the rest: `WhenFull.BLOCK` holds generation back so every
-result is presented, which is what a file output wants, and `WhenFull.DROP_OLDEST`
-skips frames to keep latency down, which is what a realtime one wants. The caller
-picks, since it is the caller that created the window.
+result is presented, which is what a window that must show all of them wants,
+and `WhenFull.DROP_OLDEST` skips frames to keep latency down, which is what a
+realtime one wants. The caller picks, since it is the caller that created the
+window. A batch run needs neither, since it writes each result as it comes.
```
ArielG-NV:
```suggestion
result is presented, which is what a window that must show all of them wants,
and `WhenFull.DROP_OLDEST` skips frames to keep latency down, which is what a
realtime program wants. The caller picks, since it is the caller that created the
window. A batch run needs neither, since it writes each result as it comes.
```

jarcherNV:
> > 
> 
> I've been on the fence about it but I've decided to get rid of the Mp4ClientWindow, and just use an Mp4OutputSink instead. If people prefer the other way let me know and I will switch it back.
> 
> I am under the opinion we want a client window to keep future work all funneling through the same objects.
> 
> This especially applies for synthetic inputs and routing them such that it looks like they came from a Client Window equally across all targets

I've thought about this a bit more and while I agree on the goal, I'm not sure a scripted input source should really be paired with an mp4 file via a window. The script can just be an InputSource rather than a window: a session never sees what produced its events, and a ScriptedInputSource could drive WebRTC and a native window too, where an
Mp4ClientWindow only helps files.

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -97,12 +113,13 @@ Not built yet
 - Input that keeps up with generation. Input is polled at the UI rate, so a run of
   fast steps can finish several of them between polls and hand them all the same
   batch. Pacing generation is what would fix it.
-- `ApplicationRunner`: takes an `IApplication` and an `IClientWindow` and drives
-  the main loop. `run_session` is what exists today; it drives a session the
-  caller already created.
-- `flashdreams-run`: a CLI that creates the requested kind of client window,
-  loads an application module, and hands both to `ApplicationRunner`. Until it
-  exists, the caller wires that up and an integration ships no entry point.
+- The loops an interactive window needs. A WebRTC or native window owns an event
```
ArielG-NV: What does "The loops an interactive window needs" mean?

ArielG-NV (commented):
Initial review

flashdreams/flashdreams/api_v2/README.md
jarcherNV: Updated.

flashdreams/flashdreams/api_v2/README.md
jarcherNV: Updated.

flashdreams/flashdreams/api_v2/README.md
jarcherNV: Updated.

flashdreams/flashdreams/api_v2/README.md
jarcherNV: Updated.

flashdreams/flashdreams/api_v2/README.md
jarcherNV: Updated to be less ambiguous.

jarcherNV: Updated to specifically refer to this.

flashdreams/flashdreams/api_v2/README.md
jarcherNV: Updated.

flashdreams/flashdreams/api_v2/README.md
jarcherNV: Updated for better wording.

flashdreams/flashdreams/api_v2/README.md
jarcherNV: Updated

flashdreams/flashdreams/api_v2/README.md
jarcherNV: I think we may be targeting different things here, so I have updated the readme to better reflect what the batch path is for.

The reason to have a batch path is comparability, same prompt and config in, same frames out, so a run can
be scored against others running the same model and against our own runs. It has a fixed step count, every frame written, and no dependence on timing, which is what run_batch does here. That lets us compare PAI-Bench scores with other model integrations like vllm-omni and certain internal orgs and ideally get comparable results, so people can't claim that our stuff is somehow worse.

If that's the goal, batch is the reference rendering and the interactive path is a lossy approximation of it since it can drop frames and polls inputs on threads, so it can't be reproduced.

Compositing into batch output would also mean that the file is no longer the model's output, and a reliable benchmark score can't be computed from it. Maybe that's not a big deal if we only compare scores from integrations that have no UI? I'll need to think about this one.

I am wondering now if we want two types of mp4 output. The ability to do regular batch inference like the other big model repos, and the ability to output mp4 from an interactive run (ie a native window outputs to mp4, or possibly both the window and mp4). Then you can get output of the interactive run as well.

jarcherNV: Thinking about this more, I think we can have both, and it doesn't need a second runner.

IClientWindow is already an InputSource + OutputSink, so an adapter pairing any source with any sink is a window by definition, and Mp4OutputSink is the output half of one. Recording an interactive run then becomes something like `run_session(session, ClientWindow(webrtc_input, Mp4OutputSink(path)))`

I think that would get you what you are after?

flashdreams/flashdreams/runtime_v2/batch_runner.py
jarcherNV: As I mentioned in a lower comment, I think we will want this for batch inference for comparability with other model serving repos like vllm-omni, so we can get similar PAI-Bench scores etc, where an interactive version can drop frames and potentially give lower scores.

gtong-nv: hmm, I think there is some misunderstanding.
We thought we still want to maintain the main loop on the top level, like the the session-runner..
But some backends/presentation mode (e.g webRTC) does it's own event loop (e.g listen to the input from browser) and we can't easily integrate that to the loop in session-runner, so we let the event-loop running in the background and give it a callback handler to receive the input. I believe native-window will behave similarly.

But for batch runner/mp4 output, I don't see a reason to have a separate loop... There would be no input during a rollout/session, and it can keep writing each frame to the mp4_output_sink by ignoring the backpressure mechanism.

jarcherNV: Oh ok, I thought the decision was to have a separate main loop for each type. I'll see if the existing main loop can be used for batch still.

jarcherNV: I have put this inference path on the same one as webrtc. I guess we will find out if that has any performance or quality implications when I run the benchmarks for it.

jarcherNV:
Apparently I was still on the fence enough to go back to having an Mp4ClientWindow.

flashdreams/flashdreams/runtime_v2/application_runner.py
```diff
@@ -27,6 +32,8 @@ def run(
         self,
         session_desc: SessionDesc,
         commandline_args: Sequence[str] = (),
+        *,
+        steps: int | None = None,
```
gtong-nv: Can we remove this from `ApplicationRunner`? `steps` should belong to a session.
For example, a t2v app will always be running, and can start one session after another to generate multiple rollouts, each rollout will have its one `steps` or `length`.

jarcherNV: Yeah that might make more sense, I'll see how easy it is to do.

jarcherNV: I have implemented this.

gtong-nv (approved):
LGTM

PR #493, opened 2026-08-20
Add t2v API and integrations and benchmark support

jarcherNV:
The first model behind the v2 protocol and the first user of the batch runner and Mp4OutputSink. Output is MP4 only, so there is no input side yet. The check it runs is shared, in flashdreams.testing_v2.t2v_conformance, since every t2v model wants the same coverage. CPU tests use a stand-in model; the real one carries ci_gpu and skips unless T2V_SELF_FORCING_REAL_MODEL_RUN is set, because pytest-manual-marker fails manual tests at setup and they never run.


flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,183 @@
...
+        help=(
+            "JSON file to record what each step measured in, for a benchmark "
+            "to read. Nothing is measured unless this is asked for."
+        ),
+    )
+    return parser
+
+
+def _as_t2v(application: IApplication) -> T2VApplication:
+    """Return the application as one that can describe its own session.
```
gtong-nv: Is this function trying to specialize `IApplication` to `T2VApplication`?
This does not sound right or necessary in this generic runtime layer

flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,183 @@
...
+        ValueError: ``steps`` is not positive.
+    """
+    if steps is not None and steps <= 0:
+        raise ValueError(f"--steps must be > 0, got {steps}.")
+
+    try:
+        # Inside, as the loop closes a session that failed to init: an
+        # application that got half way through holds whatever it acquired.
+        application.init(list(application_args))
+        describes_itself = _as_t2v(application)
```
gtong-nv: This feels unnecessary - the `flashdreams-run` does not need to know if it's what application it is, as long as it's the `IApplication` protocol.

flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,183 @@
...
+        run_session(
+            application.create_session(describes_itself.session_desc()),
+            Mp4ClientWindow(output_path, stats_path=stats_path),
+            steps=steps if steps is not None else describes_itself.total_blocks,
```
gtong-nv: `Mp4ClientWindow` or another `ClientWindow`, should be determined by args

flashdreams/flashdreams/t2v_v2/application.py
```diff
@@ -0,0 +1,327 @@
...
+        Where the seed lives is the model's business, and every model built on
+        this framework keeps it on the diffusion model, so this is written once
+        here and overridden by an integration that keeps it elsewhere.
+        """
+        return derive_config(pipeline_config, diffusion_model={"seed": seed})
+
+    ## Internals
+
+    def _resolved(self, called: str) -> T2VSessionConfig:
+        """Return what the command line resolved to.
```
gtong-nv: I don't understand the point of this function.
It seems it acts like a error message logging system. But feels unnecessary for now as those codes will only be touched by ourself and we should make the steps straightforward so it's hard to miss anything.

configs/v2_model_benchmarks.json
```diff
@@ -0,0 +1,414 @@
+{
```
jarcherNV, line 1: This file is just the benchmark config stuff for running the new integrations.

flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,188 @@
...
+    application = create_application(parsed.slug)
+    session_desc = _session_desc(application, parsed)
+    window = _client_window(parsed)
+    if isinstance(window, WebRTCClientWindow):
+        print(f"Open {window.server.url} in a browser.", flush=True)
+    # Nothing here says how long the run is. A session generates what it was
+    # configured to and reports itself finished, and a window with a client on
+    # the other end ends the run when that client goes away.
+    ApplicationRunner(application, window).run(session_desc, application_args)
+    if parsed.mode == _MP4_MODE:
```
gtong-nv: ideally, these ClientWinow specific things should be handled by each ClientWindow and then we don't need to branch here.

flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,188 @@
...
+    }
+    described = application.session_desc()
+    if described is None:
+        return SessionDesc(**asked_for)
+    return replace(described, **asked_for)
+
+
+def _client_window(parsed_args: argparse.Namespace) -> IClientWindow:
+    """Create the window the arguments ask for."""
+    if parsed_args.mode == _MP4_MODE:
```
gtong-nv: I still believe this function deserves its own file.. so when we adding another mode, we don't need to touch `cli.py`.
But I'm not sure if that's the `factory` idea..

flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,188 @@
...
+from dataclasses import replace
+from pathlib import Path
+from typing import Any
+
+from flashdreams.api_v2.application import IApplication
+from flashdreams.api_v2.client_window import IClientWindow
+from flashdreams.runtime_v2.application_runner import ApplicationRunner
+from flashdreams.runtime_v2.applications import (
+    create_application,
+    registered_application_slugs,
```
gtong-nv, line 25: not sure if it's just me, but do people feel these module names too similar and confusing ?
* `api_v2.application`, 
* `runtime_v2.application_runner`, 
* `runtime_v2.applications `
* `t2v_v2.application`

Some of them are contract, and some of them seems just provide util functions (like the `create_application`)?
Not sure if we can combine some of them, or relocate them..

flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,188 @@
...
+    )
+    parser.add_argument("slug", help="Application to run.")
+    parser.add_argument(
+        "--mode",
+        choices=(_MP4_MODE, _WEBRTC_MODE),
+        default=_MP4_MODE,
+        help="Where the run goes. Default: %(default)s.",
+    )
+    parser.add_argument(
+        "--output-path", type=Path, help="MP4 file to write. Required for mp4."
```
gtong-nv: some `args` are mode specific, maybe we can parse common ones, delegate the the mode specific ones to be parsed and resolved by ClientWindow itself?

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: I have made this application agnostic now.

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: Fixed.

flashdreams/flashdreams/t2v_v2/application.py
jarcherNV: This function is gone in the current version.

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: I added a --mode flag, feel free to change that to just use the output in your PR if you prefer.

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: Yeah these do feel similar. I'll see what I can do about that.

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: I'll think about if there is a better way.

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: Done. The mode says where the run went now, starting() before it and finished() after, so the command just prints what it is handed.

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: Moved into client_window_factory.py, which is what that factory was for. Adding a mode is adding it there and cli.py doesn't change.

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: Each mode declares and validates its own arguments now, so --output-path and --port belong to the modes that read them rather than to the command.

flashdreams/flashdreams/runtime_v2/cli.py
jarcherNV: Renamed runtime_v2.applications to application_registry, since it was the odd one out. I'm still deciding where the t2v files should live, maybe they can stay where they are for now and we move them in the next PR.

flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # Before the window, so a slug this cannot run costs nothing to find out.
+    application = create_application(parsed.slug)
+    session_desc = _session_desc(application, parsed)
+    window = _client_window(parsed)
+    if isinstance(window, WebRTCClientWindow):
+        print(f"Open {window.server.url} in a browser.", flush=True)
+    # Nothing here says how long the run is: a session reports itself finished,
+    # and a window ends the run when its client goes away.
+    ApplicationRunner(application, window).run(session_desc, application_args)
```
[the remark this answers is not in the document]
jarcherNV: This one is in serving/webrtc_server.py rather than in this PR, and the same exposure existed before it, the CLI constructed the window directly then too.

It is also narrower than it looks. The startup error path is already clean: when _start_server raises, _run_server records the error, closes the loop, and the thread exits, with the socket closed on the way out. Only a genuine timeout leaks, and reaching it means the thread is stuck inside runner.setup() or the bind for ten seconds. The thread is a daemon and flashdreams-run-v2 exits on that exception, so nothing outlives the failed run; it would only matter to a long-lived host that builds windows repeatedly, which nothing does yet.

Worth fixing when there is such a host, but that is not within the scope of this PR.

flashdreams/flashdreams/runtime_v2/cli.py
```diff
@@ -0,0 +1,156 @@
...
+    mode = client_window_mode(parsed.mode)
+    try:
+        mode.check_arguments(parsed)
+    except ValueError as error:
+        parser.error(str(error))
+
+    # Before the window, so a slug this cannot run costs nothing to find out.
+    application = create_application(parsed.slug)
+    session_desc = _session_desc(application, parsed)
+    window = mode.create(parsed)
```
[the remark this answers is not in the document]
jarcherNV: Same comment as before:

This one is in serving/webrtc_server.py rather than in this PR, and the same exposure existed before it, the CLI constructed the window directly then too.

It is also narrower than it looks. The startup error path is already clean: when _start_server raises, _run_server records the error, closes the loop, and the thread exits, with the socket closed on the way out. Only a genuine timeout leaks, and reaching it means the thread is stuck inside runner.setup() or the bind for ten seconds. The thread is a daemon and flashdreams-run-v2 exits on that exception, so nothing outlives the failed run; it would only matter to a long-lived host that builds windows repeatedly, which nothing does yet.

Worth fixing when there is such a host, but that is not within the scope of this PR.

gtong-nv (approved):
LGTM!

PR #497, opened 2026-08-21
[API-Refactor] Add full threading API + UI + Example ImGUI Demo

### LINKED ISSUE #468: [API] Add Proper UI (input & output) Pass
[API] Add Proper UI (input & output) Pass so that users may build a UI similar to how a commercial game-engine functions.

ArielG-NV:
Resolves: https://github.com/NVIDIA/flashdreams/issues/468

This PR evolves the v2 runtime from a fixed two-thread design into a general session->worker model.

For design, refer to flashdreams/flashdreams/api_v2/README.md
For demos, refer to integrations_v2/imgui_demo/imgui_demo/

Other things added:
- More event support: Keyboard Pressed & Released + mouse support + focused window
- Removed `Reset` and `Activate` buttons on the webrtc window; use ImGUI UI instead.


flashdreams/flashdreams/runtime_v2/internal_session.py
```diff
@@ -0,0 +1,233 @@
...
+        threads = self._freeze_thread_registry()
+        for thread_id in sorted(threads):
+            latest = threads[thread_id]._snapshot_latest()
+            if (
+                latest is None
+                or latest.generation != generation
+                or latest.result.disabled
+            ):
+                continue
+            composed = _composite_frame(composed, _latest_frame(latest.result))
```
jarcherNV: Does this mean the workers will only expose the one frame? I am wondering about frames that arrive faster than presentation, do they just get dropped? I imagine that will work for native window or webrtc but it will drop frames for the benchmarks, and I guess duplicate results if the model is slower. I am not sure if this is what we would want for benchmark evaluation stuff.

We may want different presentation modes or something. Like a latest-snapshot for interactive and ordered delivery for benchmarks.

flashdreams/flashdreams/runtime_v2/internal_session.py
```diff
@@ -0,0 +1,233 @@
...
+        if composed is None:
+            return None
+        output_layout = self.session_desc.output_layout
+        return StepResult(
+            step_index=presentation_index,
+            output=_frame_to_layout(composed, output_layout),
+            frame_count=1,
+            output_layout=output_layout,
```
jarcherNV: Does this drop the metrics?

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -70,260 +95,137 @@ def run_session(
...
+                    event_buffer.generation,
+                    presentation_index,
+                )
+                if composite is not None:
+                    dropped_frames += presentation_buffer.push(
+                        composite,
+                        window.write,
+                    )
+                    presentation_index += 1
+                presentation_buffer.drain(window.write)
```
jarcherNV: Does the push and drain system here allow for both of the backpressure mechanisms like BLOCK and DROP_OLDEST?

flashdreams/flashdreams/runtime_v2/internal_session.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+
+    @final
+    def _stop_threads(self) -> None:
```
jarcherNV: I'm guessing this would be called when the application terminates. We should probably include a timeout in the _join method then so it can't hang when you want to ctrl+c, a hang there would keep the app from exiting.

flashdreams/flashdreams/runtime_v2/internal_thread.py
```diff
@@ -0,0 +1,232 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Runtime-owned session worker behavior."""
```
jarcherNV: This is a general question about this code. Can a worker queue a message for itself and then wait for it? Can there be a cycle where two workers are waiting for each other? I haven't looked at the code enough to know if this can happen or not.

integrations_v2/imgui_demo/imgui_demo/app.py
```diff
@@ -0,0 +1,218 @@
...
+
+class DemoImGUIThread(ImGUIThread[ImGUIDemoState]):
+    """Draw the demo widgets on an independent UI worker."""
+
+    def draw_ui(
+        self,
+        imgui: Any,
+        step_index: int,
+        events: UserInputEvents,
+    ) -> None:
```
jarcherNV, line 61: It looks like the state is only changed in draw_ui and reset which I guess makes this safe. But I am wondering about other apps and integrations. If they don't follow these boundaries themselves then I imagine they would run into trouble. The API should probably be robust enough that this can't happen.

integrations_v2/imgui_demo/imgui_demo/app.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+
+    @property
+    def session_desc(self) -> SessionDesc:
+        """Return the resolved session description."""
+        return self._session_desc
+
+    def init(self) -> None:
+        """Register the ImGui worker before the runtime freezes registration."""
+        _ui_thread = DemoImGUIThread(
```
jarcherNV: Is it possible for an integration to just completely bypass the threading and directly use a session? T2V currently does that since this PR is not merged yet, and I am wondering what the implications would be.

jarcherNV:
I guess my general comment is that I am wondering if there should be some kind of readme file to give an example of what resources we are using, what thread should own them, who creates or destroys them, who should have access, and how it is enforced. Based on the code it is hard for me to visualise that and how the threads work and how the boundaries are enforced.

ArielG-NV:
Yes, there is a plan to have a readme + example for all of this.

integrations_v2/imgui_demo/imgui_demo/app.py
ArielG-NV: I need to make docs, but, to put it simply

 main generation thread (essentially just `step` + `reset`, where state is `ISession`) is treated same as any other thread. It is just a mandatory and implicit one

ArielG-NV:
Note: the thread boundaries right now are very poor, but that will be resolved

integrations_v2/imgui_demo/imgui_demo/app.py
ArielG-NV: model is clearly described in readme now

integrations_v2/imgui_demo/imgui_demo/app.py
ArielG-NV: added more complex demos.
gist is: only allowed to talk to other threads via `invoke_async` (messages), which modifies a `state` variable.

flashdreams/flashdreams/runtime_v2/session_runner.py
ArielG-NV: Yes, mechanism is implemented in `_PresentationBuffer::push`

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -48,232 +95,137 @@ def run_session(
...
-            # throw away the results it just finished.
+
+            # Before we start trying to read input and composite frames we should wait for the workers to be started.
+            workers_started.wait()
             while not stop.wait(tick_seconds):
-                tick()
-            # Present anything the final step produced after the last tick.
-            present_pending_results()
+                read_input()
+                event_buffer.collect_garbage()
```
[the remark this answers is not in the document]
ArielG-NV: Fixed

flashdreams/flashdreams/runtime_v2/internal_session.py
ArielG-NV: > Does this mean the workers will only expose the one frame? I am wondering about frames that arrive faster than presentation, do they just get dropped?

What I did for now is:
1. If we only have a `model_generation_thread` running, present in a lossless way
2. If we have multiple threads running, present in a lossy way

I think `1` makes sense. `2` is a more complex problem since we allow threads to tick at different rates, so the only way to produce deterministic output is to snapshot composite frames whenever a "baseline" frame is generated (likely `model_generation_thread`).

flashdreams/flashdreams/runtime_v2/internal_session.py
ArielG-NV: It did, this has been fixed.

flashdreams/flashdreams/runtime_v2/internal_thread.py
ArielG-NV: `Can a worker queue a message for itself and then wait for it?`
we snapshot a message_queue before processing and only remove the snapshot worth of messages.
A user may invoke_async on their-self as many times as they want to.

flashdreams/flashdreams/runtime_v2/internal_session.py
ArielG-NV: This has been added

ArielG-NV:
New docs/readme & examples demo'ing various different behaviors with ImGUI & threads has been added. Code is now RFR.

Note 1: Code cleanup will be actively occurring, but the general_design (minus UX/encapsulation things) & features should all be there

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/497/commits/a6519e744c6a65cff38ec80860bbe5e496e418df

flashdreams/flashdreams/runtime_v2/internal_session.py
ArielG-NV: I suppose for `2` we could track with every generated frame the `model_generation_thread_step_index` we generated a frame for, but we would still suffer inconsistencies given that threading is not "frame exact" every time (only very-close)

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/497/commits/4c32ac19de3f7eeb534b782436ef693004b94b65

flashdreams/flashdreams/api_v2/thread.py
```diff
@@ -0,0 +1,144 @@
...
+    @final
+    def step(self, step_index: int, events: UserInputEvents) -> StepResult:
+        """Render one UI frame and wrap it in a step result."""
+        frame = self.step_ui(step_index, events)
+        return StepResult(
+            step_index=step_index,
+            output=frame,
+            frame_count=1,
+            output_layout=self.output_layout,
+        )
```
jarcherNV, line 90: I am wondering why the UIThread has both step and step_ui, when would you use one function or the other?

flashdreams/flashdreams/runtime_v2/thread_manager.py
```diff
@@ -0,0 +1,375 @@
...
+        original result must retain its full frame batch and metrics. Sessions
+        with additional user-visible-threads are presentation streams: they
+        consume each thread's newest result and composite one current frame.
+        """
+        self._set_generation(generation)
+        threads = self._freeze()
+
+        # Preserve every model result when there are no additional
+        # user-visible-threads to composite.
+        if set(threads) == {_MODEL_GENERATION_THREAD_ID}:
```
jarcherNV: I know this is what we suggested in the sync and I think it makes sense, but I am wondering if we should be more explicit about it, like we have some kind of flag or policy that explicitly states if it is EXACT or LATEST_SNAPSHOT or something. I don't know if there is a specific need for it though, maybe something to think about for later.

flashdreams/flashdreams/api_v2/thread.py
ArielG-NV: > I am wondering why the UIThread has both step and step_ui, when would you use one function or the other?

`step` is `final`, therefore the user cannot implement logic for it. To impl UIThread user must supply `step_ui`.

flashdreams/flashdreams/runtime_v2/step_result.py
```diff
@@ -4,12 +4,27 @@
...
+class PresentationMode(Enum):
+    """Control how a step result participates in frame presentation."""
+
+    showPresentation = "showPresentation"
+    """Update the last-presented frame and composite it into the client backbuffer."""
+
+    hidePresentation = "hidePresentation"
+    """Update the last-presented frame without affecting the client backbuffer."""
+
+    disablePresentation = "disablePresentation"
```
jarcherNV: I think so far we've been using upper case enums, we may as well stick with that convention.

flashdreams/flashdreams/runtime_v2/thread_manager.py
ArielG-NV: We could provide a `--lossless-presentation` flag, but this makes very little sense to provide imo given the low overhead of this code, there is never a reason to really disable the flag?

ArielG-NV: although... it would be nice to not have "random floating behavior" in our compositor code.

flashdreams/flashdreams/runtime_v2/step_result.py
ArielG-NV: yes, enums should be upper, this is an error

flashdreams/flashdreams/runtime_v2/imgui_thread.py
```diff
@@ -0,0 +1,472 @@
...
+        "[": "left_bracket",
+        "\\": "backslash",
+        "]": "right_bracket",
+        "`": "grave_accent",
+    }
+    normalized = aliases.get(key.lower(), aliases.get(normalized, normalized))
+    normalized = punctuation.get(key, normalized)
+    if len(normalized) == 1 and normalized.isdigit():
+        normalized = f"_{normalized}"
+    return getattr(imgui.Key, normalized, None)
```
jarcherNV, line 464: I guess this would ignore any keys we haven't mapped here. Maybe that's fine for now, we may want to think about how to handle this later.

ArielG-NV: I think we can iterate on this later, the utility UI `ImGUIThread` is not critical/core code that is hard to add to (we are just using `ImGUI` after all).

flashdreams/flashdreams/t2v_v2/session.py
```diff
@@ -52,12 +66,13 @@ def init(self) -> None:
         Where the text encoder runs, so it is slower than a step.
         """
         self._cache = self._new_cache()
+        self.register_model_generation_thread(T2VModelThread, state=self)
```
gtong-nv: so the `state` is an `ISession`, and T2VModelThread is essentailly a wrapper around T2VSession?

flashdreams/flashdreams/runtime_v2/imgui_thread.py
```diff
@@ -0,0 +1,472 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
```
jarcherNV, line 3: I guess we should reduce the license size here.

flashdreams/flashdreams/runtime_v2/event_buffer.py
```diff
@@ -0,0 +1,100 @@
...
+import threading
+
+from flashdreams.runtime_v2.user_input_event import (
+    ResetUserInputEventData,
+    UserInputEvent,
+)
+from flashdreams.runtime_v2.user_input_events import UserInputEvents
+
+
+class EventBuffer:
```
jarcherNV, line 15: Does the stuff in this class get tested?

flashdreams/flashdreams/t2v_v2/session.py
gtong-nv: codex also spots this issue lol
```
The T2V implementation demonstrates the resulting awkwardness: [T2VModelThread (line 15)](/repo/flashdreams/flashdreams/flashdreams/t2v_v2/session.py:15) simply delegates step, reset, and is_finished back to private methods on T2VSession.
```

I guess we need to move the `step()` etc to `T2VModelThread` in this new threading model.

flashdreams/flashdreams/runtime_v2/thread_manager.py
jarcherNV: As discussed in the sync, I think we can leave this behavior in as-is for now.

ArielG-NV:
Plan is to redesign around a more restricted and minimal threading approach since our system needs less expressiveness

PR #499, opened 2026-08-21
Add Omnidreams driving application on the v2 API

jarcherNV:
The first v2 application that generates from something other than a prompt: it continues from a first frame and follows an HDMap for every frame it generates. The model is the single-view distilled checkpoint the v1 flashdreams-omnidreams package already configures, so this package is the application around it and holds no model code.

Per-step conditioning goes through an HDMapSource seam rather than through ISession, so that a Ludus scene rendered from driver input can be added later without changing anything above it. The one implementation today reads recorded video. The source is also what says how long a run is, since a recording ends when the road does.

The Python module is omnidreams_v2 because v1 owns the omnidreams name and this imports it, the same way t2v_self_forcing imports self_forcing.


jarcherNV:
Closing this PR as it was written before the new threading changes and I think it would be easier just to write this again rather than attempt to rebase, port, and clean it up after those changes were merged.

PR #515, opened 2026-08-25
Document the v2 API, runtime and integrations

jarcherNV:
ARCHITECTURE.md holds the design that spans packages: the layers, the control flow of a run, and the two threads it runs on. The api_v2 and runtime_v2 READMEs are each now only their own package, so the flashdreams-run-v2 command line moves to the runtime.

Adds a general integration guide and expands t2v_v2 into the t2v one, absorbing the boilerplate the nine integration READMEs repeated. Docstrings and comments across both packages match what the code does. A ci_cpu test checks the references in all of it still resolve, since nothing else reads Markdown.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,227 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Architecture
+
+How a FlashDreams application is put together, and what happens when one runs.
+
+This is the mental model the v2 packages assume. Each package documents its own
+surface in detail; the job here is to say how they fit together and why the seams
+are where they are.
+
+- [`flashdreams.api_v2`](flashdreams/flashdreams/api_v2/README.md) - the
+  protocols an application implements.
+- [`flashdreams.runtime_v2`](flashdreams/flashdreams/runtime_v2/README.md) - the
+  machinery that runs one.
+- [`integrations_v2`](integrations_v2/README.md) - how to write an application.
+- [`flashdreams.t2v_v2`](flashdreams/flashdreams/t2v_v2/README.md) - the
+  text-to-video layer built on all of it.
+
+The model side has a mental model of its own, warmup, CUDA-graph capture, the
+autoregressive-step body, ring attention, finalize, described in the
+[inference pipeline overview](docs/source/developer_guides/inference_pipeline_overview.rst).
+That covers what happens inside one generation step; this covers what drives the
+steps.
+
+## The shape of it
+
+Four layers, each of which only knows about the one below.
+
+```mermaid
+flowchart TB
+  subgraph integration [An integration, under integrations_v2 or flashdreams.t2v_v2]
+    App["IApplication, arguments and whatever is expensive to load"]
+    Session["ISession, one run"]
+    Model["IModelLoop, generates"]
+    UI["IUILoop, optional, decides what is shown"]
+  end
+
+  subgraph api [flashdreams.api_v2, the protocols and nothing else]
+    Protocols["IApplication, ISession, ILoop, IClientWindow, InputSource, OutputSink"]
+  end
+
+  subgraph runtime [flashdreams.runtime_v2, everything that runs one]
+    Entry["cli, application_registry, find the application"]
+    Runner["ApplicationRunner, owns the lifecycle"]
+    RunSession["run_session, owns the two threads"]
+    Meet["EventBuffer, PresentationManager, where the threads meet"]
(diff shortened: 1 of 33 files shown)


flashdreams/flashdreams/runtime_v2/README.md
```diff
@@ -0,0 +1,195 @@
...
+  batch of them a source hands over.
+
+## The command line
+
+`flashdreams-run-v2 SLUG` finds an application through the
+`flashdreams.applications_v2` entry point group and runs it with
+`ApplicationRunner`. Arguments after `--` go to the application, so
+`flashdreams-run-v2 SLUG -- --help` describes the application rather than this
+command. The split is stated rather than guessed because an application may
+declare arguments this command also has.
```
jmccaffrey-nv, line 63: this errors with default mp4 sink. for 2 slugs:

>  PYTHONPATH=/home/jmccaffrey/projects/waypoint-integ-v2-api/flashdreams /home/jmccaffrey/projects/flashdreams/.venv/bin/flashdreams-run-v2 t2v-wan21 -- --help
> usage: flashdreams-run-v2 [-h] [--mode {mp4,webrtc}] [--stats-path STATS_PATH] [--output-path OUTPUT_PATH] [--host HOST] [--port PORT]
>                           [--pixel-width PIXEL_WIDTH] [--pixel-height PIXEL_HEIGHT] [--fps FPS] [--layout {tchw,btchw,bcthw,bvtchw}]
>                           [--backpressure-mode {block,drop_oldest}] [--presentation-mode {only_present_new,only_present_newest}]
>                           slug
> flashdreams-run-v2: error: --output-path is required when writing an MP4.

ARCHITECTURE.md
```diff
@@ -0,0 +1,225 @@
...
+## Resets and failures
+
+A **reset** is a client asking to start over without closing the window. It
+arrives as an input event, and the event buffer turns it into a generation
+number. Every loop and the presentation manager compare that number against
+their own: loops reset their state and restart their step index, and frames
+generated before the reset are discarded rather than shown. One counter is the
+whole mechanism — none of those components has to know about the others.
+
+A **failure** ends the run, and whichever failure happened first is the one
```
jmccaffrey-nv: gpt 5.6 sol:
run_session always prefers a queued loop failure over a main-thread failure. A synchronized reproduction raised model close later instead of the earlier window write first.

flashdreams/flashdreams/runtime_v2/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+- `ONLY_PRESENT_NEWEST` runs the UI loop every tick, re-presenting the newest
+  model frame.
+- `ONLY_PRESENT_NEW` runs the UI loop only on a tick where `advance` actually
+  moved to a new frame.
+
+For output that has to be compared frame by frame, use `BLOCK` with
+`ONLY_PRESENT_NEW`: together they keep every frame and present each exactly once,
+in order. Frames that could not be kept are counted in `dropped_for_space` and
+`discarded_at_reset`, and logged when the run ends.
```
jmccaffrey-nv: gpt 5.6 sol:

[P3] Drop counters count chunks, not frames.
The [runtime guide says frames are counted](https://github.com/NVIDIA/flashdreams/blob/198491ad195d5a93a6a146d28a6b4ceac76ee5f2/flashdreams/flashdreams/runtime_v2/README.md#L154-L157). The implementation increments once per evicted chunk. Reproduction: two counters for eight dropped frames.

flashdreams/flashdreams/runtime_v2/README.md
jarcherNV: Thanks. It turns out this was a cli bug that hit every slug rather than two, since mp4 is the default
mode.

The cause was ordering in cli.py: mode.check_arguments ran before the application existed. Instead of documenting the current behavior I just fixed the cli bug instead.

Help now skips the window check and returns before mode.create, so `SLUG -- --help` needs no --output-path. It also stops `--mode webrtc -- --help` binding a port just to print help.

I added a test for this too.

ARCHITECTURE.md
jarcherNV: Yeah that sounds right. session_runner line 207 is `loop_failures or high_level_failures`, so a loop failure always wins on rank, not timing.

Fixed in ARCHITECTURE.md, the runtime README, and run_session's own docstring, which repeated it: a loop's failure, else the main thread's, else the first cleanup failure.

flashdreams/flashdreams/runtime_v2/README.md
jarcherNV: It turns out that the class docstring contradicted its own attribute docstrings four lines down. Both now say steps, with one step of twelve frames counting once. I also fixed run_session's log messages, which said "results" and so overcounted by the channel count.

jmccaffrey-nv (approved):

PR #516, opened 2026-08-25
[API] Native window support

### LINKED ISSUE #510: [API] Native window support
Add support for a Slangpy driven Native Window

ArielG-NV:
Resolves: #510

Add Slangpy native-window support


flashdreams/flashdreams/runtime_v2/native_window_client_window.py
```diff
@@ -0,0 +1,919 @@
...
+        cuda_resident = cuda_device is not None
+        self._ensure_render_resources(cuda_device)
+        if self._cuda_rgb_interop is not None:
+            cuda_frame = self._cuda_rgb_interop.as_cuda_rgb_frame(frame)
+            if cuda_frame is not None:
+                if not self._cuda_rgb_interop.enqueue_rgb_to_shared_rgba(cuda_frame):
+                    return True
+                while not self._submit_ready_cuda_rgb():
+                    if not self._wait_for_presentation_progress():
+                        return False
```
jarcherNV: Can this exit be reached? I think _wait_for_presentation_progress() returns False only when self._closed is set, and since write() runs on the same thread as close() and the event pump, I don't think anything can set it while it's spinning here.

flashdreams/flashdreams/runtime_v2/native_window_client_window.py
```diff
@@ -0,0 +1,919 @@
...
+        self._present_host_upload()
+        return True
+
+    def close(self) -> None:
+        """Release CUDA interoperability and native window resources."""
+        if self._closed:
+            return
+        self._closed = True
+        if self._cuda_rgb_interop is not None:
+            self._cuda_rgb_interop.close()
```
jarcherNV: Does this also need to call self._device.wait_for_idle() like how the slangpy UI stuff works?

Without that call I suspect this could be setting resource to None while the GPU may still be using them.

flashdreams/flashdreams/runtime_v2/native_window_client_window.py
```diff
@@ -0,0 +1,919 @@
...
+    """CUDA event recorded after conversion to presentation format."""
+
+
+class NativeWindowClientWindow(IClientWindow):
+    """Present UI output through a main-thread GLFW window.
+
+    ``run_session`` owns threading. It calls :meth:`open`,
+    :meth:`get_user_input_events`, and :meth:`close` on its I/O thread, while
+    :meth:`write` runs on the runtime's presentation thread.
+    """
```
jarcherNV: I'm not sure this description is correct. Doesn't IO and presentation happen on the same thread? Line 190 also refers to a presentation thread.

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -23,8 +23,7 @@ The application chooses its default `SessionDesc`; `--pixel-width`,
 `--presentation-mode` can override it.
 
 The selected client-window mode handles the run's input and output. MP4 mode
-writes a file and has no input. WebRTC mode streams to a browser. Mode-specific
-arguments are defined in `runtime_v2/client_window_factory.py`.
+writes a file and has no input. WebRTC mode streams to a browser. Native-window mode presents GPU-resident tensors through SlangPy without materializing them on the CPU. Mode-specific arguments are defined in `runtime_v2/client_window_factory.py`.
```
jarcherNV: The readme changes will probably need to be updated after this is rebased.

flashdreams/flashdreams/runtime_v2/native_window_client_window.py
ArielG-NV: yes, but are we supposed to write code that is fragile to a different thread changing its state (when reasonably, another thread may have the ability to `close` our client window)?

ArielG-NV: I suppose we should just assume thread-unsafe

flashdreams/flashdreams/api_v2/README.md
ArielG-NV: done 👍

flashdreams/flashdreams/runtime_v2/native_window_client_window.py
ArielG-NV: 👍

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/516/commits/242ac1cf8bdb1ee9c266823f926ecd99d9d6f090

flashdreams/flashdreams/runtime_v2/native_window_client_window.py
ArielG-NV: note: `_cuda_rgb_interop` only closes copy stream (blocks until copies are done if I recall right)

I suppose a device wait would be needed after though (before window close to ensure presentation/writing is done?)

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/516/commits/7659064e0887cb232ff379d2a824367ecdda09ab

ArielG-NV:
Since we are assuming all client-windowing happens on the same thread as io frame submission I simplified the code around this idea

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/516/commits/47474a03a3e28722ed3f676795d2526b8fc5d134

jarcherNV (approved):

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/516/commits/504511d4cea1b4b1b16a79288a8879e6d559e2eb

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/516/commits/a6b23d7c4f54d08c5f1225bdd2744263f7340cd8

PR #1, opened 2026-04-28
Add GitHub Actions CI configuration for NVIDIA self-hosted runners

jarcherNV:
Configure CI workflows to use NVIDIA's self-hosted GPU and CPU runners:
- Add copy-pr-bot configuration to enable automated PR branch syncing
- Add CI workflow with CPU job for wheel builds and doc builds
- Add CI workflow with GPU job for world model diffusion (2x RTX Pro 6000)
- Add Renovate configuration for automated dependency updates

The CPU job uses linux-amd64-cpu8 runners for build tasks, while the GPU job uses linux-amd64-gpu-rtxpro6000-latest-2 runners with 2x RTX Pro 6000 GPUs for world model diffusion workloads.

PR #464, opened 2026-08-15
Add waypoint-1.5-1B support

lpy:
This patch adds support for waypoint-1.5-1B, a real time interactive video world model, which is designed to generate on consumer hardware.

This patch:

1) extends the existing taehv to support encoder;
2) implements the necessary bits, e.g. DiT, adaptive RMSNorm, etc;
3) supports using JSON file to pass in controls conditions.

PR #14, opened 2026-04-29
Delete files which accidentally reappeared after rebase

jatentaki:
These files were renamed in https://github.com/NVIDIA/flashdreams/pull/12 and rebase accidentally made them reappear. This MR deletes them for good.

PR #519, opened 2026-08-25
Add Waypoint 1.5 1B support with V2 interactive runtime

jmccaffrey-nv:
## Summary

Adds Waypoint 1.5 1B inference to FlashDreams as the `waypoint-1-5-1b` V2
application. This integrates the model work from #464 and ports it to the
current inference, application, session, event, and presentation APIs.

The integration supports deterministic MP4 rollouts and interactive browser
control through the V2 WebRTC runtime.

## What changed

- Add the independently packaged Waypoint model pipeline,
  published-checkpoint loading, four-step scheduler, control encoding,
  transformer, and autoregressive cache.
- Extend the shared Hunyuan Video 1.5 TAEHV recipe for the streaming
  32-channel latent format, including a guard against incomplete meta-device
  checkpoint loads.
- Add a V2 `WaypointApplication` and `WaypointSession` with reusable
  application-level model state and isolated per-session cache, controls, and
  RNG state.
- Support finite JSON control timelines, bundled example data, deterministic
  seeds, profiling, reset, and live keyboard, mouse, and wheel events.
- Present seed and generated frames directly on the native 1024x512 canvas,
  avoiding an intermediate 1280x720 resize and a second seed resample.
- Reuse fixed-attention masks by cache visibility pattern and reuse each Q/K
  FP32 RoPE promotion.
- Detach seed and generated presentation tensors at the model-loop boundary.
- Add a design-review architecture reference with use-case, component,
  class/ownership, and sequence diagrams.
- Add a public model card with the support matrix, measured performance,
  official-reference parity, limitations, safety, and provenance.

Waypoint emits four RGB frames per action at 60 FPS playback timing. The
published checkpoint has no text-conditioning path, so the application is
driven by a seed image and controls.

## Validation

- Full pre-commit gate passed: Ruff fix/check, Ruff format, lockfile/version
  checks, and `ty`.
- Warning-as-error Sphinx build passed for the complete documentation site,
  including the new model card and gallery/homepage registrations.
- Changed-code CPU suites: 52 passed across TAEHV, Waypoint inference, and
  Waypoint V2.
- Broad CPU tier: 1,584 passed, 2 skipped, 337 deselected after excluding only
  the unrelated Omnidreams interactive-drive directory whose optional packages
  were not installed in this worktree.
- CUDA fixed-cache FlexAttention equivalence and mask reuse over local/global
  layers and eight autoregressive frames: 2 passed.
- Published-weight one-action and 40-action inference completed on an NVIDIA
  RTX PRO 6000 Blackwell Workstation Edition.
- The final 40-action MP4 is native 1024x512 with exactly 164 frames.
- Actions 20-40 improved from 88.427 ms median to 68.887 ms median per four
  generated frames. The before/after 40-action MP4s are byte-identical.
- Earlier extended validation ran all 15 available Biome seeds for 40 actions
  and the pinned example for all 118 actions. Full-schedule parity with the
  pinned official implementation reached 0.999869 cosine similarity for the
  clean latent.

Detailed commands, checkpoint revisions and hashes, latency samples, output
hashes, and parity measurements are recorded in
`integrations_v2/waypoint/VALIDATION.md`.

## Design-review references

- `integrations/waypoint/README.md` — scope, decisions, four static/dynamic
  diagrams, tensor contracts, ownership, concurrency, and cache semantics.
- `docs/source/models/waypoint.rst` — user-facing model card with modalities,
  requirements, commands, measured performance, parity, limitations, and
  provenance.
- `integrations_v2/waypoint/ADR-1-control-events.md` — browser-to-model input
  and reset/focus semantics.

## Design notes

- `integrations/waypoint` owns model inference and remains independent of
  runtime orchestration.
- `integrations_v2/waypoint` owns CLI parsing, lifecycle, event adaptation,
  and finite/live presentation behavior.
- File-driven controls take precedence over live input for deterministic
  rollouts.
- One application intentionally serializes shared model and RNG work while
  preserving per-session cache and RNG state.

### DIFF
--- a/integrations_v2/waypoint/waypoint_v2/app.py
+++ b/integrations_v2/waypoint/waypoint_v2/app.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import argparse
+import math
 import secrets
 import threading
 from collections.abc import Callable, Sequence
@@ -32,8 +33,8 @@ from flashdreams.runtime_v2.session_desc import (
 from flashdreams.runtime_v2.video_tensor import VideoTensorLayout
 from waypoint_v2.session import WaypointSession
 
-_OUTPUT_WIDTH = 1280
-_OUTPUT_HEIGHT = 720
+_OUTPUT_WIDTH = 1024
+_OUTPUT_HEIGHT = 512
 _PLAYBACK_FPS = 60
 _EXAMPLE_IMAGE_FILENAME = "crystal_desert_blade.jpg"
 _EXAMPLE_IMAGE_URL = (
@@ -91,7 +92,7 @@ class WaypointApplication(IApplication):
             ValueError: Inputs do not describe a valid file or live rollout.
         """
         parser = argparse.ArgumentParser(
-            prog="waypoint-1.5-1b",
+            prog="flashdreams-run-v2 waypoint-1-5-1b --",
             description="Run Waypoint 1.5 from an image using file or live controls.",
         )
         parser.add_argument("--seed-image", type=Path)
@@ -112,10 +113,7 @@ class WaypointApplication(IApplication):
             raise ValueError("pass --seed-image or --example-data")
         if args.seed is not None and args.seed < 0:
             raise ValueError(f"--seed must be non-negative, got {args.seed}")
-        if (
-            not torch.isfinite(torch.tensor(args.mouse_sensitivity)).item()
-            or args.mouse_sensitivity < 0
-        ):
+        if not math.isfinite(args.mouse_sensitivity) or args.mouse_sensitivity < 0:
             raise ValueError("--mouse-sensitivity must be finite and non-negative")
 
         controls_path = args.controls_file
@@ -148,7 +146,7 @@ class WaypointApplication(IApplication):
         )
 
     def session_desc(self) -> SessionDesc:
-        """Return Waypoint's fixed 720p TCHW presentation contract."""
+        """Return Waypoint's native 1024x512 TCHW presentation contract."""
         return SessionDesc(
             output_layout=VideoTensorLayout.tchw,
             backpressure_mode=BackpressureMode.BLOCK,
@@ -210,7 +208,7 @@ class WaypointApplication(IApplication):
 
 
 def load_seed_display_frames(path: Path) -> Tensor:
-    """Load one RGB/RGBA image as four normalized 720p TCHW seed frames.
+    """Load one RGB/RGBA image as four normalized 1024x512 TCHW seed frames.
 
     Args:
         path: Image used to establish the initial world state.
--- a/integrations_v2/waypoint/waypoint_v2/app.py
+++ b/integrations_v2/waypoint/waypoint_v2/app.py
@@ -1,0 +1,293 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Waypoint 1.5 V2 application."""
+
+from __future__ import annotations
+
+import argparse
+import secrets
+import threading
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
+from pathlib import Path
+from typing import cast
+
+import torch
+from loguru import logger
+from torch import Tensor
+
+from flashdreams.api_v2.application import IApplication
+from flashdreams.core.io.disk import default_flashdreams_cache_dir
+from flashdreams.core.io.download import download_to_cache
+from flashdreams.infra.config import derive_config
+from flashdreams.runtime_v2.session_desc import (
+    BackpressureMode,
+    PresentationMode,
+    SessionDesc,
+)
+from flashdreams.runtime_v2.video_tensor import VideoTensorLayout
+from waypoint import WaypointControl, load_controls_from_file
+from waypoint.config import PIPELINE_WAYPOINT_1_5
+from waypoint.pipeline import WaypointInferencePipeline
+
+from waypoint_v2.session import WaypointSession
+
+_OUTPUT_WIDTH = 1280
+_OUTPUT_HEIGHT = 720
+_PLAYBACK_FPS = 60
+_EXAMPLE_IMAGE_FILENAME = "crystal_desert_blade.jpg"
+_EXAMPLE_IMAGE_URL = (
+    "https://raw.githubusercontent.com/Overworldai/Biome/14343a6/seeds/"
+    + _EXAMPLE_IMAGE_FILENAME
+)
+_EXAMPLE_CONTROLS = Path(__file__).parent / "assets" / "example_controls.json"
+
+PipelineFactory = Callable[[int, torch.device, bool], WaypointInferencePipeline]
+SeedLoader = Callable[[Path], Tensor]
+ExampleResolver = Callable[[], Path]
+
+
+@dataclass(frozen=True, slots=True)
+class _ApplicationConfig:
+    seed_image: Path | None
+    controls: tuple[WaypointControl, ...] | None
+    seed: int
+    device: torch.device
+    profile: bool
+    mouse_sensitivity: float
+
+
+class WaypointApplication(IApplication):
+    """Load one Waypoint model and create isolated image-established sessions."""
+
+    def __init__(
+        self,
+        *,
+        pipeline_factory: PipelineFactory | None = None,
+        seed_loader: SeedLoader | None = None,
+        example_resolver: ExampleResolver | None = None,
+    ) -> None:
+        """Create a lazy Waypoint application.
+
+        Args:
+            pipeline_factory: Test seam replacing real checkpoint construction.
+            seed_loader: Test seam replacing image decode and normalization.
+            example_resolver: Test seam replacing the example image download.
+        """
+        self._pipeline_factory = pipeline_factory or _create_pipeline
+        self._seed_loader = seed_loader or load_seed_display_frames
+        self._example_resolver = example_resolver or _download_example_image
+        self._config: _ApplicationConfig | None = None
+        self._pipeline: WaypointInferencePipeline | None = None
+        self._pipeline_lock = threading.Lock()
+
+    def init(self, commandline_args: Sequence[str]) -> None:
+        """Parse seed, control mode, device, and deterministic rollout settings.
+
+        Args:
+            commandline_args: Application-specific command-line arguments.
+
+        Raises:
+            ValueError: Inputs do not describe a valid file or live rollout.
+        """
+        parser = argparse.ArgumentParser(
+            prog="waypoint-1.5-1b",
+            description="Run Waypoint 1.5 from an image using file or live controls.",
+        )
+        parser.add_argument("--seed-image", type=Path)
+        parser.add_argument("--example-data", action="store_true")
+        parser.add_argument("--controls-file", type=Path)
+        parser.add_argument(
+            "--actions",
+            type=int,
+            help="Use this many actions from a controls file.",
+        )
+        parser.add_argument("--seed", type=int)
+        parser.add_argument("--device", default="cuda")
+        parser.add_argument("--profile", action="store_true")
+        parser.add_argument("--mouse-sensitivity", type=float, default=1.0)
+        args = parser.parse_args(list(commandline_args))
+
+        if args.seed_image is None and not args.example_data:
+            raise ValueError("pass --seed-image or --example-data")
+        if args.seed is not None and args.seed < 0:
+            raise ValueError(f"--seed must be non-negative, got {args.seed}")
+        if (
+            not torch.isfinite(torch.tensor(args.mouse_sensitivity)).item()
+            or args.mouse_sensitivity < 0
+        ):
+            raise ValueError("--mouse-sensitivity must be finite and non-negative")
+
+        controls_path = args.controls_file
+        if controls_path is None and args.example_data:
+            controls_path = _EXAMPLE_CONTROLS
+        controls = (
+            load_controls_from_file(controls_path)
+            if controls_path is not None
+            else None
+        )
+        if args.actions is not None:
+            if controls is None:
+                raise ValueError("--actions requires --controls-file or --example-data")
+            if not 1 <= args.actions <= len(controls):
+                raise ValueError(
+                    f"--actions must be in [1, {len(controls)}], got {args.actions}"
+                )
+            controls = controls[: args.actions]
+
+        seed = secrets.randbits(63) if args.seed is None else args.seed
+        if args.seed is None:
+            logger.info(f"[waypoint-1.5-1b] generated seed {seed}")
+        self._config = _ApplicationConfig(
+            seed_image=args.seed_image,
+            controls=controls,
+            seed=seed,
+            device=torch.device(args.device),
+            profile=args.profile,
+            mouse_sensitivity=args.mouse_sensitivity,
+        )
+
+    def session_desc(self) -> SessionDesc:
+        """Return Waypoint's fixed 720p TCHW presentation contract."""
+        return SessionDesc(
+            output_layout=VideoTensorLayout.tchw,
+            backpressure_mode=BackpressureMode.BLOCK,
+            presentation_mode=PresentationMode.ONLY_PRESENT_NEW,
+            frames_per_second_for_ui=_PLAYBACK_FPS,
+            frames_per_second_for_step=_PLAYBACK_FPS,
+            video_width=_OUTPUT_WIDTH,
+            video_height=_OUTPUT_HEIGHT,
+            metadata={
+                "model": "waypoint-1.5-1b",
+                "frames_per_action": 4,
+                "internal_resolution": "1024x512",
+            },
+        )
+
+    def create_session(self, session_desc: SessionDesc) -> WaypointSession:
+        """Resolve cheap inputs, load the shared model once, and create a session.
+
+        Args:
+            session_desc: Runtime-requested presentation settings.
+
+        Returns:
+            An uninitialized session with independent cache, controls, and RNG.
+
+        Raises:
+            RuntimeError: The application has not been initialized.
+            ValueError: Waypoint cannot honor the requested layout or dimensions.
+        """
+        config = self._require_config()
+        _validate_session_desc(session_desc)
+        seed_image = config.seed_image or self._example_resolver()
+        seed_frames = self._seed_loader(seed_image)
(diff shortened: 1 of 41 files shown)


jmccaffrey-nv:
## Code review — Waypoint 1.5 V2 integration

Reviewed the full diff against `main` (36 files, +5371/-36). Verified locally: repo-pinned `ruff==0.12.7` check / format / isort are all clean, and `pytest integrations_v2/waypoint -m ci_cpu` gives 20 passed in 6s. Docs (README/ADR-1/VALIDATION) are unusually good for an integration of this size.

Two things I specifically checked and want to record as **correct**, because they're the parts most likely to be wrong:

- **The fixed KV ring/eviction math holds.** I traced `_fixed_view` against `WaypointAttentionPolicy.visible_frame_indices` for local (`dilation=1`) and global (`dilation=8`) layers at warm-up, at first wrap, and in steady state (e.g. frame 136: ring holds pinned 16…120 + 128, tail holds 136, slot 1 masked during the pass). Visible counts match the policy exactly at every point, with no duplicate or stale frame. The `visible[ring_slice] &= not write_step` trick is correct — just very hard to read; a comment naming *why* the slot is hidden (it still holds the frame being evicted) would help the next reader.
- **The wheel sign convention in ADR-1 is right.** `runtime_v2/serving/web/app.js:136` already sends `wheel_y = -Math.sign(event.deltaY)`, so positive really is wheel-up. Worth keeping the ADR line that says so.

---

### Correctness / behavior

**1. Presentation is anamorphically stretched, and the seed image is resampled twice.** `app.py:212-246` resizes the source image to 1280×720, then `session.py:238-246` immediately resizes it again to 512×1024 for the model, and `session.py:249-270` stretches every generated 1024×512 frame back to 1280×720. Waypoint's canvas is 2:1; 1280×720 is 16:9. The vertical axis is scaled 1.406× against the horizontal's 1.25×, so the output is ~12.5% vertically stretched, and the seed makes a lossy round trip through the wrong aspect ratio on the way in. Options: declare `session_desc` at the native 1024×512 (see #14 — this also deletes both interpolate paths), or letterbox instead of stretching. Right now the README says "the adapter resizes results for presentation" without noting geometry is not preserved.

**2. `reset()` doubles peak cache memory.** `session.py:118-119`:
```python
with state.pipeline_lock:
    state.cache = state.pipeline.initialize_cache(seed_pixels=state.seed_pixels)
```
The RHS is evaluated while `state.cache` still holds the old one. Summing the fixed KV stores (6 global layers × 129 frames × 512 tokens × 16 kv-heads × 128 × 2 × 2B ≈ 3.2 GiB, plus 18 local layers ≈ 1.3 GiB) that's a ~4.5 GiB transient spike on top of the 6.355 GiB steady-state peak you measured. `state.cache = None` before the call fixes it.

**3. Every session in a process shares one seed.** `app.py:138` draws the seed once in `init()` and `create_session` hands the same value to every session. The docstrings say sessions have "isolated ... RNG state", which is true, but they're isolated copies of the *same* stream — two WebRTC users on the same seed image get byte-identical rollouts. For a served application, deriving per-session (`seed + session_ordinal`) while keeping `--seed` as the base would be more useful, and it's what "isolated" implies.

**4. `pipeline_lock` serializes far more than it needs to.** `session.py:78-85` holds the lock across the whole `generate()` + `finalize()` — ~92 ms measured. The only shared mutable thing is `diffusion_model.rng` (`infra/diffusion/model/base.py:124-138`); the weights are read-only and could run concurrently. Two live sessions today each get ~5.5 actions/s instead of sharing the GPU. Threading a per-session `torch.Generator` through `DiffusionModel.generate` would be the framework-level fix; at minimum, note the limitation.

**5. Hard-coded checkpoint fingerprint.** `checkpoint.py:288` rejects anything that isn't exactly `{"BF16": 393}` / `1_860_823_096` elements. Any republish by Overworld — even a byte-identical model with one added metadata tensor — hard-fails. (This is currently unreachable; see #10.)

**6. `legacy_to_blocks_keys` now rewrites `encoder.*` for every existing TAEHV consumer** (`recipes/taehv/checkpoint.py`). Harmless today because `load_state_dict(strict=False)` drops them for decoder-only instances, but combined with `strict=False` a future encoder-key typo loads silently. A post-load check that no parameter is still on `meta` would close that.

**7. `use_cuda_graph=False, use_compile=False`** in `config.py:213-215` overrides the `Hy15TAEHVDecoderConfig` defaults (both `True`). Intentional? If it's a streaming-cache shape issue it deserves a comment, because it's leaving decoder time on the table in the one component the recipe already optimized.

---

### Performance

**8. The block mask is rebuilt on every attention call, with two device syncs each.** `_fixed_block_mask` (`cache.py:305-360`) runs `torch.equal(...)` (line 320) and `.nonzero()` → `active.numel()` (line 322, 330, 333) per call. Both force a GPU→CPU sync. That's 24 layers × 5 forward passes = **~240 syncs per action**, plus a `written.clone()` of up to 66k bools each time. The mask only changes when a frame commits — once per action per layer — and after 16/128 frames `written` is all-ones and the mask is *constant* for the rest of the rollout. Mirroring `written` as a CPU-side bitmap and memoizing the `BlockMask` per (layer, pattern) should remove essentially all of it. This is my highest-confidence perf item against the 91.7 ms/action number. (Also worth a `TORCH_LOGS=recompiles` pass: `visible_slot` is a fresh closure per call feeding `@torch.compile`'d `_compiled_fixed_attention`.)

**9. `apply_waypoint_ortho_rope` casts to fp32 twice.** `rope.py:51-52`:
```python
first  = tokens.float()[..., 0::2]
second = tokens.float()[..., 1::2]
```
Each `.float()` materializes the whole tensor before slicing. Q is 8 MB and K is 4 MB per layer per pass → ~1.4 GB of redundant fp32 allocation+copy per action. One-line fix: hoist to `promoted = tokens.float()`.

**10. `_ControlFusion` recomputes an identical projection over 512 duplicated tokens.** `network.py:676` expands `[B,1,D]` control to `[B,S,D]`, then each of the 8 fusion blocks runs `F.rms_norm` and `fc1_c` (2048×2048) over all 512 rows — but every row is identical, so `rms_norm` and `fc1_c` produce the same vector 512 times. Keeping control at `[B,1,D]` and letting `fc1_x(tokens) + fc1_c(control)` broadcast is mathematically identical and saves ~34 GFLOP per forward (~170 GFLOP/action, ~3% of model FLOPs) plus 8 × 2 MB of `rms_norm` output per pass.

**11. Two per-forward recomputations of constants.** `network.py:537-545` re-derives `output_weight` from `unpatchify.weight` with a `permute().reshape()` (forces a copy) and re-expands the bias on every call; `network.py:700-719` rebuilds the fixed `row_index` / `column_index` `arange`s every call. Both are rollout-invariant — cache them as buffers.

**12. Redundant permute round trip.** `pipeline.py:105-107` does `seed_latent.permute(0,2,1,3,4).contiguous()` and hands it to `self.decoder(...)`, which is `WaypointTAEHVDecoder.forward` — whose first act is to permute and `.contiguous()` straight back. Two full copies for a no-op.

**13. Same for `TPool` in the streaming encoder.** `recipes/taehv/impl.py`: `torch.cat(group, dim=1).view(batch*stride, C, H, W)` is immediately reshaped back to `[-1, stride*C, H, W]` inside `TPool.forward`. Calling `block.conv(torch.cat(group, dim=1))` directly is the same thing without the round trip (and the `elif len(group) > block.stride: raise` branch below is unreachable, since the group resets at `==`).

---

### Reducing SLoC / leaning on existing components

Production (non-test) code is **3402 lines** across the two packages. About **~200 lines are removable with no behavior change**, and another ~25 fall out with #1:

| # | What | Where | ~Lines |
|---|------|-------|--------|
| 14 | `Hy15TAEHVEncoder`'s `StreamingVideoEncoder` surface (`forward`, `initialize_autoregressive_cache`, both ratio properties, `get_input/output_temporal_size`) is **never called in production** — `pipeline.py:86` reaches past it to `self.seed_encoder.taehv.encode(...)`. Only `flashdreams/tests/test_vae.py` exercises `forward`. Either use the wrapper (it does the `[-1,1]→[0,1]` conversion the pipeline currently open-codes) or drop it and expose `TAEHV(enable_encoder=True)` directly. | `recipes/taehv/__init__.py:296-363` | ~55 |
| 15 | `inspect_safetensors_checkpoint` and `validate_waypoint_1_5_checkpoint` are referenced **nowhere** — not by the pipeline, not by tests. They're also ~80% duplicated with each other (both open safetensors, both loop keys counting dtypes and elements). | `checkpoint.py:211-290` | ~80 |
| 16 | `make_control_context` builds `frame_idx` and `frame_timestamp` tensors that `predict_flow` never reads — it passes `frame_index` as an `int` separately. Only `test_spec.py:104-105` asserts on them. | `controls.py:190-198, 203-204` | ~15 |
| 17 | `assets/example_data/waypoint/example_controls.json` is byte-identical to the packaged `waypoint_v2/assets/example_controls.json`, and nothing reads the repo-root copy. | new file | 15 |
| 18 | Output-layout validation is written twice — `app.py:249-261` (`_validate_session_desc`) and again in `WaypointSession.__init__` (`session.py:168-172`). Pick one boundary. | | ~12 |
| 19 | `expected_waypoint_1_5_checkpoint_keys` is `frozenset(expected_..._shapes(spec))` with a 12-line docstring; inline it. | `checkpoint.py:118-129` | ~12 |
| 20 | `hasattr(self, "ctrl_mlpfusion")` (`network.py:342`) as control flow — declare `self.ctrl_mlpfusion: _ControlFusion \| None` and test `is not None`. Friendlier to `ty` and to Dynamo. | | ~0 |
| 21 | Declaring `session_desc` at the native 1024×512 removes `_presentation_frames`' resize + shape validation and the second seed resample entirely — and fixes #1. | `session.py:238-270` | ~25 |

**Reusing `flashdreams.core.attention`:** `apply_waypoint_ortho_rope` (`rope.py:26-58`) hand-rolls the rotation in fp32 with a `torch.cat`. `flashdreams.core.attention.apply_rope_freqs` is a fused Triton kernel that does it in place. The reason it doesn't drop in cleanly is that Waypoint reads adjacent pairs `(2k, 2k+1)` but writes packed halves — neither of the kernel's two modes. But that output ordering is a *shared permutation* applied identically to Q and K, and nothing downstream depends on head-channel order (V is unroped; `out_proj` runs after the score). So `interleaved=True` produces identical attention scores. If that holds up in a parity test, it deletes the function, the two fp32 copies from #9, and the `torch.cat`, in exchange for the fused kernel. Worth measuring — it's the single largest reuse opportunity in the model package.

**`BlockKVCache`** (`core/attention/kvcache.py`) already implements the dense sliding window that 18 of the 24 layers use. It doesn't cover the dilated pinned-global policy, so it isn't a drop-in — but if the global policy were pushed into it as a stride option, `transformer/cache.py` (360 lines) could shrink substantially and the framework would gain a reusable sparse-history cache. That's a follow-up, not a blocker for this PR.

**On the larger structural question:** `integrations/waypoint/waypoint/transformer/` (1635 lines) sits outside `flashdreams/recipes/`, where `wan` and `cosmos` live with the same `transformer/{__init__,impl/{network,modules}}` shape. The comparison point is `hy_worldplay` (8126 lines under `integrations/`), so the placement has precedent and I don't think it should block. But `norm.py` (81 lines of `adaptive_rms_norm` / `adaptive_gate`) and the RoPE construction are generic AdaLN-Zero/3D-RoPE primitives that other DiTs will want; if a second interactive world model lands, `recipes/` is where they should end up.

---

### Nits

- `app.py:116` validates `--mouse-sensitivity` with `torch.isfinite(torch.tensor(...))` while `control_events.py:343` uses `math.isfinite` for the same value. Use `math` in both.
- `app.py:94` sets `prog="waypoint-1.5-1b"` but the registered slug is `waypoint-1-5-1b`, and `t2v_v2` uses `prog="flashdreams-run-v2 SLUG --"`. The `--help` output currently prints a name you can't type.
- `network.py` hard-codes the noise-embedding width as `512` in `embed_noise` and `steps=256` in `_NoiseEmbedder.__init__` — two magic numbers that must stay in sync and aren't in `WaypointModelSpec`.
- `TAEHV.encode` uses `@torch.no_grad()` while `TAEHV.decode` uses `@torch.inference_mode()`.
- `MemBlock`'s `x * 0 if past is None else past` allocates; `torch.zeros_like(x)` reads better and is the same cost.
- `secrets.randbits(63)` for a reproducibility seed — `random.getrandbits` conveys intent better.

Nothing here is a blocker on the model math, which parity-checks at 0.999869 cosine against the reference. The items I'd most want addressed before merge are **#1** (visible geometry bug), **#2** (reset spike), and **#8** (the per-call block-mask syncs); **#14–#17** are free deletions.

jmccaffrey-nv:
Addressed the suitable items from the Claude review in the commits ending at cbed12ed.

Accepted changes:
- Switched V2 presentation to native 1024x512 and removed the second seed resize plus generated-frame upscale.
- Release the old session KV cache before reset allocates its replacement.
- Corrected the application help program name and replaced tensor-based finite-value validation with math.isfinite.
- Cached fixed-attention BlockMask objects by visibility pattern, documented why the active ring slot is hidden, and added CUDA identity/parity coverage through wraparound.
- Reused each Q/K FP32 RoPE promotion and removed the redundant seed-latent contiguous copy.
- Added a TAEHV post-load guard for parameters or buffers left on meta.
- Removed the unused path-level checkpoint fingerprint helpers while retaining strict key and shape validation.
- Made control fusion explicitly optional, moved the Fourier width into WaypointModelSpec, and documented why seed encoder decoder graph/compile settings are disabled.
- Documented intentional deterministic session seeding, application-level model/RNG serialization, and the one-client WebRTC runtime.

Not changed after evaluation:
- Per-session seed randomization: the configured seed intentionally reproduces a control sequence across reset and session recreation.
- Control projection over one token: an isolated trial changed generated output materially (one-action encoded comparison PSNR 44.14 dB and SSIM 0.9836), so it was rejected.
- TPool reshape/view paths: these are metadata-only views, not tensor copies.
- Direct seed_encoder.taehv.encode: the wrapper is a reusable public encoder contract, while the direct call intentionally avoids a normalize/denormalize round trip for already normalized seed pixels.
- Duplicate app/session layout checks: these protect separate construction boundaries; direct session construction is covered by tests.
- Root example controls and frame metadata: both have active test/API consumers.
- Output-weight/static-index caching and larger BlockKVCache/RoPE-factor refactors: deferred because they require checkpoint/meta-safe buffer design and separate parity evidence.

Validation:
- Full pre-commit passed, including Ruff and ty.
- 52 focused CPU tests passed; broad CPU run passed 1,584 with 2 skipped after excluding only the unrelated Omnidreams interactive-drive optional-dependency directory.
- 2 CUDA fixed-cache parity tests passed.
- Published-weight one-action and 40-action runs passed on the RTX PRO 6000.
- Steady-state median improved from 88.427 ms to 68.887 ms per action, and the before/after 40-action MP4s are byte-identical.

integrations_v2/waypoint/waypoint_v2/app.py
```diff
@@ -0,0 +1,290 @@
...
+        _EXAMPLE_IMAGE_URL,
+        cache_dir=default_flashdreams_cache_dir() / "example_data" / "waypoint",
+        filename=_EXAMPLE_IMAGE_FILENAME,
+        validator=load_seed_display_frames,
+    )
+
+
+def create_app() -> IApplication:
+    """Return a new lazy Waypoint 1.5 V2 application."""
+    return WaypointApplication()
```
gtong-nv, line 287: Not sure if we want a dedicated `WaypointApp`, or maybe we just integrate that to the `cam2v` app if the input/output is similar to lingbot?
Update: I see this app takes mouse input. We should probably wait and add that support to `cam2v` first. This can be done in the future

integrations_v2/waypoint/waypoint_v2/session.py
```diff
@@ -0,0 +1,264 @@
...
+        """Generate the seed result or one controlled autoregressive action."""
+        state = self.state
+        if step_index == 0:
+            if state.seed_emitted or state.controls_generated:
+                raise RuntimeError("Waypoint seed step is out of sequence")
+            state.seed_emitted = True
+            return [
+                StepResult(
+                    step_index=0,
+                    output=state.seed_frames,
```
gtong-nv: `output=state.seed_frames.detach()` 
make the `StepResult` outputs detached before they leave `WaypointModelLoop`.
That gives Waypoint the same ownership pattern as LingBot/Cam2V: : model internals can keep their caches/RNG/module state, while the runtime receives plain presentation tensors.

integrations_v2/waypoint/waypoint_v2/session.py
```diff
@@ -0,0 +1,264 @@
...
+            f"{tuple(output.shape)}"
+        )
+    target_size = (session_desc.video_height, session_desc.video_width)
+    if tuple(output.shape[-2:]) != target_size:
+        raise ValueError(
+            "Waypoint pipeline output spatial size must match the session: expected "
+            f"{target_size[1]}x{target_size[0]}, got "
+            f"{output.shape[-1]}x{output.shape[-2]}"
+        )
+    return output.contiguous()
```
gtong-nv: `return output.detach().contiguous()`
same here.

gtong-nv (approved):
Tested locally with webRTC mode, works as expected.
A few minor comments. Can be addressed in followups.

integrations_v2/waypoint/waypoint_v2/app.py
jmccaffrey-nv: (GPT Sol) Agreed. I am keeping WaypointApplication dedicated in this PR because the current shared Cam2V app converts keyboard input into camera poses, while Waypoint consumes raw keyboard/mouse buttons, mouse deltas, and wheel input and has a different seed/cache lifecycle. Converging on Cam2V makes sense once that shared app exposes the needed mouse/control contract; I have left this as the future follow-up you suggested.

integrations_v2/waypoint/waypoint_v2/session.py
jmccaffrey-nv: (GPT Sol) Addressed in 33f38780: _presentation_frames now returns output.detach().contiguous(). The same regression test forces the generated pipeline tensor to require gradients and verifies the published generated result is detached.

integrations_v2/waypoint/waypoint_v2/session.py
jmccaffrey-nv: (GPT Sol) Addressed in 33f38780: the seed StepResult now uses state.seed_frames.detach(). I also added a regression test that deliberately makes the model-owned seed tensor require gradients and verifies the published result has neither requires_grad nor a grad_fn.

PR #521, opened 2026-08-26
Require SPDX copyright tags in source headers

jmccaffrey-nv:
## Summary

- require non-empty SPDX copyright and license tags in first-party source headers
- allow contributor-specific and multiple copyright notices
- standardize contributor and agent guidance on the concise two-line SPDX form

## Validation

- executed the extracted CI header-check script across tracked source files
- verified NVIDIA, individual contributor, and contributor-group notices pass
- verified empty and whitespace-only copyright values fail
- parsed the workflow YAML and ran git diff --check


jmccaffrey-nv:
/ok to test https://github.com/NVIDIA/flashdreams/commit/18f27e2b29941b080af2a98409018e14a79d319f

PR #524, opened 2026-08-26
Add typed tensor artifact outputs to V2

jmccaffrey-nv:
## Summary

- add model-neutral typed tensor artifact schemas and outputs to the V2 session/result contracts
- add a generation-aware `ModelOutputSink` path that receives complete model-step channel batches independently of UI presentation and the legacy metrics sink
- persist requested artifacts as transactional NumPy files with a machine-readable commit manifest

## Contract details

- artifact names are unique across the complete result-channel batch for a model step
- reset generations use latest-generation-wins semantics; newer results clear buffered older output and late stale results are ignored
- schemas, ranks, NumPy-compatible dtypes, and concatenated/non-concatenated shapes are validated before a batch is accepted
- `--tensor-artifact-dir` fails before application/window startup when an application declares no artifact schemas
- the sink creates and validates its destination during `open`, before generation starts
- `tensor_artifacts.json` records completeness plus each declared artifact’s emitted state, dimension names, concatenation axis, path, dtype, and shape
- only successful model execution commits buffered arrays and a `complete: true` manifest; failed execution discards buffered arrays and leaves the manifest incomplete
- declared artifacts remain optional; a successful run removes stale files for declared artifacts it did not emit
- publication stages backups for every declared artifact and restores the prior artifact set if array replacement, stale-file removal, or final manifest promotion fails

## Validation

- `python -m pytest flashdreams/test_v2 -m ci_cpu -q` — 276 passed
- `pre-commit run --files <changed files>` — passed, including Ruff and `ty`

## Stack

This is the model-neutral prerequisite for #520. It contains no LingBot-specific code, tests, documentation, or assets. PR #520 will be rebased onto this branch and reduced to the LingBot model plus its V2 adapter.

### DIFF
--- a/flashdreams/flashdreams/runtime_v2/step_result.py
+++ b/flashdreams/flashdreams/runtime_v2/step_result.py
@@ -7,8 +7,8 @@ from dataclasses import dataclass, field
 
 from torch import Tensor
 
-from flashdreams.runtime_v2.video_tensor import VideoTensorLayout
 from flashdreams.runtime_v2.tensor_artifact import TensorArtifactOutput
+from flashdreams.runtime_v2.video_tensor import VideoTensorLayout
 
 
 @dataclass(frozen=True, slots=True)
--- a/flashdreams/flashdreams/runtime_v2/step_result.py
+++ b/flashdreams/flashdreams/runtime_v2/step_result.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass, field
 from torch import Tensor
 
 from flashdreams.runtime_v2.video_tensor import VideoTensorLayout
+from flashdreams.runtime_v2.tensor_artifact import TensorArtifactOutput
 
 
 @dataclass(frozen=True, slots=True)
@@ -34,3 +35,5 @@ class StepResult:
     metrics: dict[str, float | int] = field(default_factory=dict)
     """Measurements for this step, such as timings, keyed by name. Recorded only
     when a run asked for a metrics sink, and only from a model loop."""
+    tensor_artifacts: tuple[TensorArtifactOutput, ...] = ()
+    """Named tensor artifacts emitted alongside the video result."""
--- a/flashdreams/flashdreams/runtime_v2/tensor_artifact.py
+++ b/flashdreams/flashdreams/runtime_v2/tensor_artifact.py
@@ -1,0 +1,69 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Typed tensor artifacts emitted alongside generated video."""
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+
+from torch import Tensor
+
+_ARTIFACT_NAME = re.compile(r"^[A-Za-z0-9][A-Za-z0-9_.-]*$")
+"""Portable artifact name accepted as a filename stem."""
+
+
+@dataclass(frozen=True, slots=True)
+class TensorArtifactSchema:
+    """Describe one named tensor artifact produced by a session."""
+
+    name: str
+    """Stable name used to route and persist the artifact."""
+
+    dimension_names: tuple[str, ...]
+    """Semantic names for the tensor dimensions, in storage order."""
+
+    concatenate_axis: int | None = 0
+    """Axis joining outputs from multiple steps; ``None`` permits one output."""
+
+    def __post_init__(self) -> None:
+        if not _ARTIFACT_NAME.fullmatch(self.name):
+            raise ValueError(
+                "Tensor artifact names must start with an alphanumeric character "
+                "and contain only alphanumerics, dots, underscores, or hyphens."
+            )
+        if any(not name.strip() for name in self.dimension_names):
+            raise ValueError("Tensor artifact dimension names must be non-empty.")
+        if len(set(self.dimension_names)) != len(self.dimension_names):
+            raise ValueError("Tensor artifact dimension names must be unique.")
+        if self.concatenate_axis is not None and not (
+            -len(self.dimension_names)
+            <= self.concatenate_axis
+            < len(self.dimension_names)
+        ):
+            raise ValueError(
+                "Tensor artifact concatenate_axis must identify a declared dimension."
+            )
+
+
+@dataclass(frozen=True, slots=True)
+class TensorArtifactOutput:
+    """Carry one tensor matching a declared artifact schema."""
+
+    schema: TensorArtifactSchema
+    """Schema identifying and describing the tensor."""
+
+    tensor: Tensor
+    """Artifact tensor. Sinks detach and transfer it before persistence."""
+
+    def __post_init__(self) -> None:
+        if self.tensor.ndim != len(self.schema.dimension_names):
+            raise ValueError(
+                f"Tensor artifact {self.schema.name!r} declares "
+                f"{len(self.schema.dimension_names)} dimensions but received "
+                f"shape {tuple(self.tensor.shape)}."
+            )
+
+
+__all__ = ["TensorArtifactOutput", "TensorArtifactSchema"]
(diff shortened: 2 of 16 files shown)


flashdreams/flashdreams/runtime_v2/tensor_artifact_output_sink.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+            if dimension != axis and expected != received
+        ]
+        if mismatches:
+            raise ValueError(
+                f"Tensor artifact {schema.name!r} changed non-concatenated "
+                f"dimensions from {tuple(first.shape)} to {tuple(tensor.shape)}."
+            )
+
+    def close(self) -> None:
```
jarcherNV: Is it intentional that a failed run writes its artifacts? run_session's finally (session_runner.py:224) closes every sink regardless of how the run ended, and close() here persists whatever chunks it managed to buffer before re-raising the failure.

For a streaming model that emits one chunk per step, a run that dies at step 6 of 10 leaves an actions.npy with the right name, the right dtype, and a shape that looks perfectly reasonable, nothing in the file says how many chunks were supposed to be in it. The metrics sink and the MP4 writer don't really have this problem, since a short JSON or a short video is obviously short, but a tensor isn't self-describing.

Would it make sense to either skip the write when the run failed, or record somewhere the consumer can see what the file actually contains?

flashdreams/flashdreams/runtime_v2/tensor_artifact_output_sink.py
```diff
@@ -0,0 +1,212 @@
...
+            schema = schemas[name]
+            outputs[name] = (
+                artifact_chunks[0]
+                if schema.concatenate_axis is None
+                else torch.cat(artifact_chunks, dim=schema.concatenate_axis)
+            )
+        if not outputs:
+            return
+
+        self._output_dir.mkdir(parents=True, exist_ok=True)
```
jarcherNV: The mkdir happens in close(), so if --tensor-artifact-dir points at something that's already a file, or a directory the process can't write to, the run fails after the whole rollout has finished. open() on line 37 already receives the session and could mkdir(parents=True, exist_ok=True) there instead.

Asking because cli.py:80 already fails fast on the other half of this flag, it rejects the directory before window startup when the application declares no artifacts, so the intent seems to be to catch bad input early.

flashdreams/flashdreams/runtime_v2/tensor_artifact.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@dataclass(frozen=True, slots=True)
+class TensorArtifactSchema:
+    """Describe one named tensor artifact produced by a session."""
+
+    name: str
+    """Stable name used to route and persist the artifact."""
+
+    dimension_names: tuple[str, ...]
```
jarcherNV, line 24: dimension_names is doing one functional job here, giving the expected rank so TensorArtifactOutput can check it. The names themselves never reach the output: np.save with allow_pickle=False stores dtype and shape only, so someone loading actions.npy sees (64, 16) and has to read the integration source to find out which axis is which.

Since the API goes to the trouble of declaring them, would it be worth writing them out too? Like a sidecar JSON, or .npz with the names attached? Otherwise they're documentation that only exists on the producing side.

jarcherNV (approved):
Overall LGTM. I left a few comments but I don't think any are blocking. Up to you if you think they are worth addressing or not.

flashdreams/flashdreams/runtime_v2/tensor_artifact_output_sink.py
```diff
@@ -0,0 +1,212 @@
...
+            return
+        chunks = self._chunks
+        self._schemas = None
+        self._generation = None
+        self._chunks = {}
+
+        outputs: dict[str, Tensor] = {}
+        for name, artifact_chunks in chunks.items():
+            if not artifact_chunks:
+                continue
```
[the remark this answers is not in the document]
jmccaffrey-nv: Agreed and fixed in current head 0f0eb2a0. A successful close now removes the existing .npy file for every declared artifact omitted by the current latest generation, then installs the completed manifest last. The regression test reuses a directory containing a stale optional artifact and verifies that the stale file is removed. Publication rollback also restores that prior file if a later commit step fails.

Evaluated and implemented by GPT-5.6 Sol via Codex.

flashdreams/flashdreams/runtime_v2/tensor_artifact_output_sink.py
jmccaffrey-nv: This was not intentional, and the lifecycle now distinguishes commit from cleanup. ModelOutputSink.close(commit=False) is used when model execution fails, so the tensor sink discards buffered chunks. It writes an incomplete manifest during open() and changes it to complete: true only after successful array publication. An end-to-end regression emits one chunk, fails on the next step, and verifies that no partial .npy file is committed.

Evaluated and implemented by GPT-5.6 Sol via Codex.

flashdreams/flashdreams/runtime_v2/tensor_artifact_output_sink.py
jmccaffrey-nv: Agreed and moved this validation to open(). The sink now creates the directory and atomically writes its initial incomplete manifest before generation, so a path that is already a file or cannot be written fails at startup. The invalid-file destination is covered by a focused regression.

Evaluated and implemented by GPT-5.6 Sol via Codex.

flashdreams/flashdreams/runtime_v2/tensor_artifact.py
jmccaffrey-nv: Agreed. I kept the .npy format for compatibility and added a tensor_artifacts.json sidecar manifest. It preserves each declared artifact dimension name and concatenation axis, and records emitted state, path, dtype, and shape. The manifest also acts as the completeness marker. CLI, sink, and documentation tests verify the consumer-visible metadata.

Evaluated and implemented by GPT-5.6 Sol via Codex.

jmccaffrey-nv:
Review follow-up: current head 0f0eb2a0 addresses both Greptile findings and all three Jesse Archer suggestions. Successful runs remove omitted declared artifact files and publish a complete schema manifest; failed model execution discards buffered arrays and leaves an incomplete marker; destination validation happens before generation; and dimension metadata is consumer-visible. Publication now stages backups and restores overwritten or removed prior artifacts if any array or final-manifest publication step fails. After rebasing on current main, the complete V2 CPU suite passes with 276 tests, and pre-commit passes including Ruff and ty.

Evaluated and implemented by GPT-5.6 Sol via Codex.

flashdreams/flashdreams/runtime_v2/tensor_artifact_output_sink.py
```diff
@@ -0,0 +1,319 @@
...
+                    schemas, outputs, generation=generation, complete=True
+                ),
+            )
+            for temporary_path, destination in staged:
+                os.replace(temporary_path, destination)
+            for name in schemas.keys() - outputs.keys():
+                (self._output_dir / f"{name}.npy").unlink(missing_ok=True)
+            os.replace(
+                manifest_temporary_path,
+                self._output_dir / _MANIFEST_FILENAME,
```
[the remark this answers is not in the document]
jmccaffrey-nv: Agreed. Current head 0f0eb2a0 now stages backups for every declared artifact before mutating the destination set. If array replacement, stale-file removal, or final manifest promotion fails, it restores both overwritten and omitted prior arrays before re-raising. A regression forces the final manifest replacement to fail after both mutation kinds and verifies the prior files are byte-for-byte intact and the manifest remains incomplete.

Evaluated and implemented by GPT-5.6 Sol via Codex.

flashdreams/flashdreams/runtime_v2/step_result.py
```diff
@@ -34,3 +35,5 @@ class StepResult:
     metrics: dict[str, float | int] = field(default_factory=dict)
     """Measurements for this step, such as timings, keyed by name. Recorded only
     when a run asked for a metrics sink, and only from a model loop."""
+    tensor_artifacts: tuple[TensorArtifactOutput, ...] = ()
```
jmccaffrey-nv: Find >1 model that needs such non-video output before adding it. 
Should this be list of tensor ?

PR #527, opened 2026-08-26
Fix v2 WebRTC latest-frame delivery

gtong-nv:
# Fix v2 WebRTC Latest-Frame Delivery

## Summary

Fix v2 WebRTC browser latency caused by stale frames accumulating in the server-side video track. The WebRTC sender now keeps a single latest-frame slot, so interactive cam2v input is reflected in the newest available frame instead of waiting behind an unbounded FIFO backlog.

## Root Cause

Cam2V defaults to `PresentationMode.ONLY_PRESENT_NEW`, which previously made the v2 WebRTC server create `_VideoTrack` with an unbounded `asyncio.Queue`. When generated/UI frames arrived faster than the browser sender consumed them, old frames could pile up and delay the frame that reflected the latest key input.

## Changes

- Replaced `_VideoTrack`'s queue with a single latest-frame slot guarded by an `asyncio.Condition`.
- Made newly enqueued frames replace stale unsent frames unconditionally for the WebRTC transport.
- Kept the video track live by allowing `recv()` to repeat the latest frame at the paced cadence when no newer frame has arrived.
- Preserved pending CUDA transfer lifetime handling when overwritten frames are retired or the track closes.
- Updated v2 WebRTC tests to cover latest-frame delivery and repeated-frame behavior without buffering.

## Notes

- This changes WebRTC transport presentation semantics only; model generation and cam2v session defaults are unchanged.
- GPU/browser manual latency validation was not run locally.

## Tests

Local verification:

```bash
uv run pytest flashdreams/test_v2/test_webrtc_client_window.py apps/cam2v/tests/test_application.py
```

Result:

```text
19 passed
```

### DIFF
--- a/flashdreams/flashdreams/runtime_v2/serving/webrtc_server.py
+++ b/flashdreams/flashdreams/runtime_v2/serving/webrtc_server.py
@@ -25,7 +25,7 @@ from aiortc.mediastreams import MediaStreamError
 from av import VideoFrame
 from loguru import logger
 
-from flashdreams.runtime_v2.session_desc import PresentationMode, SessionDesc
+from flashdreams.runtime_v2.session_desc import SessionDesc
 from flashdreams.runtime_v2.step_result import StepResult
 from flashdreams.runtime_v2.user_input_event import (
     CloseUserInputEventData,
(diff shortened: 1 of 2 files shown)


gtong-nv:
/ok to test [2de62bd](https://github.com/NVIDIA/flashdreams/pull/527/commits/2de62bd94dea4078dbc44d1c8e79abe85a909068)

ArielG-NV (approved):
LGTM

PR #528, opened 2026-08-27
Fix WebRTC UI latency issue

gtong-nv:
## Summary

This PR makes v2 interactive presentation write-driven: the UI thread owns
frame selection, composition, and output cadence, while WebRTC delivers the
frames admitted by `window.write()` without a second pacer or repeated-latest
behavior.

It also orders model output, UI presentation, and WebRTC transfers across
dedicated CUDA streams, bounds sender congestion to two unsent frames, and
keeps Cam2V input and retained SlangPy status updates responsive independently
of model-generation cadence.

## Presentation behavior

- Rename the presentation modes to describe their behavior:
  - `PresentationMode.ON_DEMAND` runs the UI only for a newly selected model
    frame.
  - `PresentationMode.CONTINUOUS` runs the UI every UI tick and may reuse the
    current model frame.
- Run Cam2V in continuous mode so held controls and time-driven status can
  update while model generation is busy.
- Advance at most one model frame per UI tick. Presentation follows recent
  model-frame throughput and reanchors missed deadlines instead of emitting
  burst catch-up writes.
- Keep floating-point model-frame dtypes through composition by centralizing
  overlay conversion in `PresentationManager.composite()`.

## CUDA synchronization

- Automatically record a private readiness event when a CUDA `StepResult` is
  constructed on its producer stream.
- Let consumers call `StepResult.wait_for_output()` to enqueue an event
  dependency and retain the tensor allocation on their own stream.
- Give `PresentationManager` ownership of a high-priority CUDA presentation
  stream. The complete UI step, including SlangPy rendering and RGB/RGBA
  composition, runs in that context.
- Use a separate high-priority WebRTC transfer stream per CUDA device. A write
  waits for the UI result, converts it to RGB8, copies it through reusable
  pinned host storage, and finishes materialization before returning.
- Synchronize presentation and transfer streams during shutdown before
  releasing their resources. Failed CUDA transfers quarantine pinned storage
  rather than reusing it unsafely.
- Make MP4 and shared video conversion wait for CUDA result readiness before
  reading output.

## WebRTC delivery

Each WebRTC write must contain exactly one UI-composited frame. Once the video
track exists, `window.write()` synchronously materializes an independently
owned `VideoFrame` and admits it during connection negotiation as well as after
media connection.

The sender:

- Keeps at most two queued, unsent frames in FIFO order.
- Evicts only the oldest queued frame on overflow; a frame already handed to
  aiortc or the encoder remains committed.
- Returns a queued frame immediately when aiortc requests one, without its own
  sleep or pacing policy.
- Waits for a new frame instead of repeating the latest frame.
- Assigns monotonic RTP timestamps from frame-admission timing.
- Performs a bounded shutdown drain without synthesizing a duplicate frame.

A transient WebRTC `disconnected` state no longer terminates the session or
shows a terminal browser error. Terminal `failed` and `closed` states still
produce a close event.

## Interactive input and Cam2V status

- Split fixed-rate input-window advancement from keyboard held-state
  projection so the timeline can be reused by other input modalities.
- Preserve timestamp ordering, quick taps, key aliases, focus-loss release,
  reset behavior, and simultaneous physical/arrow-key sources.
- Keep reliable keyboard controls separate from pointer traffic. Coalesce
  supersedable pointer movement and wheel updates and avoid sending them when
  the pointer channel is congested.
- Retain the native SlangPy Cam2V overlay with active-key display and
  wall-time-weighted model throughput from AR steps completed in the trailing
  two seconds.
- Log one synchronized AR-step wall time and chunk FPS line for Lingbot while
  disabling device-wide pipeline profiling barriers in its interactive path.

## Scope

This keeps the standard SlangPy UI. It does not include the reverted custom
raster HUD, browser latency panel or input tracing, A/B benchmark harness,
presentation metrics sink, or Pillow dependency.

WebRTC remains an interactive low-latency sink and may evict an old unsent
frame during congestion. Use MP4 when every frame must be retained exactly.
CUDA stream priority can overtake queued lower-priority work, but cannot
preempt a model kernel that is already executing.

## Compatibility notes

- `PresentationMode.ONLY_PRESENT_NEW` and `ONLY_PRESENT_NEWEST` are replaced by
  `ON_DEMAND` and `CONTINUOUS` without compatibility aliases.
- Direct WebRTC callers must write one UI-composited frame at a time;
  multi-frame model chunks are rejected.
- A CUDA `StepResult` must be constructed while the stream that produced its
  output is current.
- Cam2V removes `--log-every-blocks`; the Lingbot interactive defaults enable
  one concise timing record for every AR step.


apps/cam2v/session.py
```diff
@@ -27,7 +31,73 @@
...
+from .ui import (
+    RECENT_MODEL_FPS_WINDOW_SECONDS,
+    Cam2VHUDLoop,
+    Cam2VModelStepTiming,
+    Cam2VUIState,
+    Cam2VUIStatus,
+)
+
+
+class _RecentModelFrameRate:
```
gtong-nv: This should be hoist out to be used by all apps.

apps/cam2v/session.py
```diff
@@ -27,7 +31,73 @@
...
+        """Discard observations from the previous rollout."""
+        self._steps.clear()
+
+    def snapshot(self) -> tuple[Cam2VModelStepTiming, ...]:
+        """Return immutable recent model timings for the UI thread."""
+        return tuple(self._steps)
+
+
+@dataclass(frozen=True, slots=True)
+class _PendingInputEventTrace:
```
gtong-nv: This belongs to a util class as well

apps/cam2v/session.py
```diff
@@ -335,6 +442,52 @@ def _buffer_keyboard_events(
...
+        return
+    pending.append(
+        _PendingInputEventTrace(
+            event_timestamp_s=event_timestamp_s,
+            event_id=event_id,
+        )
+    )
+
+
+def _resolve_input_event_traces(
```
gtong-nv: hoist this out to the input event class

apps/cam2v/defaults.py
```diff
@@ -88,18 +88,18 @@ class Cam2VApplicationDefaults:
     backpressure_mode: BackpressureMode = BackpressureMode.BLOCK
     """Preserve every generated model frame in presentation order."""
 
-    presentation_mode: PresentationMode = PresentationMode.ONLY_PRESENT_NEW
-    """Render and transmit the overlay once for each selected model frame."""
+    presentation_mode: PresentationMode = PresentationMode.ON_DEMAND
+    """Render selected model frames and UI-relevant browser input changes."""
```
ArielG-NV: What does this comment mean...?

flashdreams/flashdreams/runtime_v2/blit_model_output_to_screen_loop.py
```diff
@@ -37,11 +38,16 @@ def step(self, step_index: int, events: UserInputEvents) -> StepResult | None:
...
+        output_ready_event = None
+        if output.is_cuda:
+            output_ready_event = torch.cuda.Event()
+            output_ready_event.record(torch.cuda.current_stream(output.device))
         return StepResult(
             step_index=step_index,
             output=_frame_to_layout(output, self.output_layout),
             frame_count=1,
             output_layout=self.output_layout,
+            output_ready_event=output_ready_event,
```
ArielG-NV: We could simplify the code here

Why don't we just make `StepResult::output_ready_event` into `StepResult::_output_ready_event`. Then we still have a `StepResult::__init__` param named `output_ready_event` but it is optional. Default behavior when value is not provided and device is `is_cuda` assigns `output_ready_event` to  
```
            output_ready_event = torch.cuda.Event()
            output_ready_event.record(torch.cuda.current_stream(output.device))
```

When provided (`None`/custom `output_ready_event`) we do not synth our own event.

Goal: make it more ergonomic. With this change we would not need to change these IUILoop files and it will all be as fast as possible

-----------------------

Maybe we make this a Future?

flashdreams/flashdreams/runtime_v2/slangpy_ui_loop.py
```diff
@@ -85,13 +108,41 @@ def draw(ui: Any, index: int, current_events: UserInputEvents) -> None:
...
+                        "a CUDA device."
+                    )
+                output_ready_event = torch.cuda.Event()
+                output_ready_event.record(presentation_stream)
         return StepResult(
             step_index=step_index,
             output=frame.unsqueeze(0),
             frame_count=1,
             output_layout=self.output_layout,
+            output_ready_event=output_ready_event,
```
ArielG-NV: https://github.com/NVIDIA/flashdreams/pull/528/changes#r3876372584

Code looks "functionally" identical.

flashdreams/flashdreams/runtime_v2/step_result.py
```diff
@@ -33,4 +35,31 @@ class StepResult:
 
     metrics: dict[str, float | int] = field(default_factory=dict)
     """Measurements for this step, such as timings, keyed by name. Recorded only
-    when a run asked for a metrics sink, and only from a model loop."""
+    when a run asked for a metrics sink, and only from a model loop. Names that
+    start with ``runtime_`` are reserved for measurements added by the runtime."""
+
+    output_ready_event: torch.cuda.Event | None = field(
```
ArielG-NV: https://github.com/NVIDIA/flashdreams/pull/528/changes#r3876372584

apps/cam2v/defaults.py
gtong-nv: With the `should_redraw_for_input()` removed from this PR, this should be commented as
`Render and present only when there is new model frame`

flashdreams/flashdreams/runtime_v2/step_result.py
```diff
@@ -31,6 +36,116 @@ class StepResult:
...
+            consumer = torch.cuda.current_stream(self.output.device)
+        if resolve_cuda_device(consumer.device) != resolve_cuda_device(
+            self.output.device
+        ):
+            raise ValueError("StepResult output and consumer stream must match.")
+        if self._output_ready_event is not None:
+            consumer.wait_event(self._output_ready_event)
+        self.output.record_stream(consumer)
+
+    def replace(self, **changes: Any) -> "StepResult":
```
ArielG-NV: this is wrong for a data class

If a user wants a new `StepResult` they should construct a new `StepResult`, not hack in the class changes, we have a frozen object for good-reason

flashdreams/flashdreams/runtime_v2/webrtc_client_window.py
```diff
@@ -51,7 +53,17 @@ def __init__(
...
-            self._input_events.put(event)
+            with self._input_lock:
+                if (
+                    isinstance(event, MouseUserInputEvent)
+                    and event.action == "move"
+                    and self._input_events
+                    and isinstance(self._input_events[-1], MouseUserInputEvent)
+                    and self._input_events[-1].action == "move"
+                ):
+                    self._input_events[-1] = event
```
ArielG-NV: why do we need a special case like this for mouse movement?
This looks like a hack..?

flashdreams/flashdreams/runtime_v2/step_result.py
```diff
@@ -31,6 +36,116 @@ class StepResult:
...
+        consumer = stream
+        if consumer is None:
+            consumer = torch.cuda.current_stream(self.output.device)
+        if resolve_cuda_device(consumer.device) != resolve_cuda_device(
+            self.output.device
+        ):
+            raise ValueError("StepResult output and consumer stream must match.")
+        if self._output_ready_event is not None:
+            consumer.wait_event(self._output_ready_event)
+        self.output.record_stream(consumer)
```
ArielG-NV: This design makes it very easy to do UB, what we could do is:
- rename member `output` to `_output`
- make an accessor for output ==> `read_output(sync_with_event=True)`/`read_output(sync_with_event=False)`

flashdreams/flashdreams/runtime_v2/slangpy_ui_loop.py
```diff
@@ -85,22 +108,65 @@ def draw(ui: Any, index: int, current_events: UserInputEvents) -> None:
...
+        def render_frame() -> Tensor:
+            overlay = self.renderer.render(step_index, events, draw)
+            if (
+                back_buffer is not None
+                and back_buffer.is_floating_point()
+                and overlay.is_floating_point()
+                and overlay.dtype != back_buffer.dtype
+            ):
+                overlay = overlay.to(dtype=back_buffer.dtype)
+            return self._presentation_manager.composite(back_buffer, overlay)
```
ArielG-NV: conversions should be handled by  ` self._presentation_manager.composite`

flashdreams/flashdreams/runtime_v2/slangpy_ui_loop.py
```diff
@@ -85,22 +108,65 @@ def draw(ui: Any, index: int, current_events: UserInputEvents) -> None:
...
+            )
+
+        device = resolve_cuda_device(presentation_stream.device)
+        with torch.cuda.device(device), torch.cuda.stream(presentation_stream):
+            frame = render_frame()
+            if not frame.is_cuda or resolve_cuda_device(frame.device) != device:
+                raise ValueError(
+                    "The SlangPy UI output and presentation stream must share "
+                    "a CUDA device."
+                )
```
ArielG-NV: should occur in `StepResult`

gtong-nv:
/ok to test [27f2abe](https://github.com/NVIDIA/flashdreams/pull/528/commits/27f2abe6bedab58f0cb69eb3ac15373ede1c42b2)

flashdreams/flashdreams/api_v2/README.md
```diff
@@ -165,24 +171,23 @@ This loop runs forever. Override `is_finished` to make it stop.
...
-  io-thread's pace.
+  every generated frame and can slow the model thread to the UI thread's pace.
 - `BackpressureMode.DROP_OLDEST` discards old buffered work so the UI can catch
   up to newer output. This favors low latency over preserving every frame.
 
 `SessionDesc.presentation_mode` handles the UI loop ticking faster than the
 model-generation-loop produces frames:
 
-- `PresentationMode.ONLY_PRESENT_NEWEST` runs the UI every tick and may reuse
+- `PresentationMode.CONTINUOUS` runs the UI every tick and may reuse
```
ArielG-NV: Reuse still exists?

flashdreams/flashdreams/runtime_v2/README.md
```diff
@@ -147,16 +162,16 @@ full:
 `SessionDesc.presentation_mode` decides what the UI does when no new frame is
 ready:
 
-- `ONLY_PRESENT_NEWEST` runs the UI loop every tick, re-presenting the newest
+- `CONTINUOUS` runs the UI loop every tick, re-presenting the newest
   model frame.
-- `ONLY_PRESENT_NEW` runs the UI loop only on a tick where `advance` actually
-  moved to a new frame.
+- `ON_DEMAND` runs the UI loop only when `advance` moves to a new frame.
```
ArielG-NV: we should be very clear in defining `model frame` since we use this term a lot now

model frame: "A model frame is the multi-channel-frame produced by our presentation manager from a `StepResult` (chunk) returned by the model-loop/model-thread"

We should likely point agents.txt to this file for terminology definition?

ArielG-NV (commented):
PR is looking very good now. 
3 minor suggestions/questions.

flashdreams/flashdreams/runtime_v2/step_result.py
ArielG-NV: We would need to remove that `output_sink.py` comment change if we make this change.

flashdreams/flashdreams/runtime_v2/README.md
ArielG-NV: if the term is already defined, my apologies... its just a mouthful of a term that is kind-of hard to trace without being already familiar with flashdreams

flashdreams/flashdreams/runtime_v2/webrtc_client_window.py
gtong-nv: hmm, this is another event coalesce that AI thinks will help performance, which is true as there could be easily hundred of cursor changes for a few seconds, but we just care about the last one seen by this tick.
I will remove this from this PR for now.

flashdreams/flashdreams/runtime_v2/README.md
gtong-nv: right, 
latent frame/chunk/AR: usually 4 or 8 frames produced by one auto regression step.
model frames: RGBA frames produced by model.
frame: model frames composited with UI elements.

I will add something like that.

flashdreams/flashdreams/runtime_v2/webrtc_client_window.py
ArielG-NV: I see...
maybe we just search the user event queue backwrads

ArielG-NV: although a user may want to know that a mouse was moved a lot within a frame....

ArielG-NV: I suppose this is why many programs have callbacks and polling normally...

flashdreams/flashdreams/runtime_v2/step_result.py
gtong-nv: Removed `wait_for_output` and use the accessor to do sync

ArielG-NV (approved):
LGTM

gtong-nv:
/ok to test [cc2e651](https://github.com/NVIDIA/flashdreams/pull/528/commits/cc2e651fc8308430b16d0bd2bee1004401756e2a)

gtong-nv:
/ok to test [dfc66fe](https://github.com/NVIDIA/flashdreams/pull/528/commits/dfc66fe94f5707617a5dce9e3edab35dfc0b93e1)

gtong-nv:
/ok to test [df34c8e](https://github.com/NVIDIA/flashdreams/pull/528/commits/df34c8e9a550105d9090eb8b17dd34290b492339)

ArielG-NV:
breaks demo: uv run --no-sync flashdreams-run-v2 slangpy-ui-text-input --mode webrtc

PR #544, opened 2026-08-29
Presentation Clock Ownership And Backlog Pacing

### LINKED ISSUE #542: [API Cleanup] presentation_clock and presentation_manager refactoring
While investigating the latency issiue in crazy-robotaxi game, we notice high `finalize_to_present_ms`
```
[crazy-robotaxi-chunk-trace] phase=app_frame_presented time_ns=3043545065272674 generation=0 step=76 epoch=1 ar=76 frame=7 simulation_us=1700000020399796 step_lead=2 ar_lead=2 simulation_lead_ms=533.328 finalize_to_present_ms=1061.12858
```

`finalize_to_present_ms`, means the chunks are ready, but there is 1s delay to present that. Change the max_pending chunks in PresentationManager from 2 to 1, the finalize_to_present_ms was cut to 500ms

The UI loop ticks at 60, but it only consumes a new model frame when `presentation_clock.is_due(...)` says yes
This was added so that UI does not present frames all at once and stutters ( in the case of lingbot, model can only produce 13 frames a second, we don't want to finish presenting the frames in 0.3s and waiting for new ones)
but if for omnidreams, assume the model can produce 30~40 FPS, and if the UI pacing slightly slower, the backlog will build up. But now the UI ticks at 60Hz, technically the backlog won't build up. Maybe there are some hiccups in the UI thread? Anyway, to resolve the issue

- PresentationManager._buffer should be configured to 1 chunk for least latency
- presentation_clock should watch for backlog as well, not just the past 2 seconds average FPS

```
model thread:
  generated chunk -> PresentationManager._buffer # if this buffer is large, and UI thread does not drain it fast enough, there will be latency 

UI thread:
  PresentationManager chooses one frame from chunk
  app builds current UI/backbuffer
  ImGui overlay renders
  composite happens
  window.write() presents immediately(edited)
```

gtong-nv:
## Summary

Move model-frame presentation pacing into `PresentationManager`, so callers only
publish chunks and call `advance()`. The manager now owns both timing and backlog
decisions, which keeps queue pressure local to the component that consumes model
output.

## Changes

- Moved `_PresentationClock` from `session_runner` to `PresentationManager`.
- `PresentationManager.publish()` observes model-output timing, and
  `PresentationManager.advance()` automatically decides whether a model frame is
  due.
- Changed the pending presentation buffer to `_bufferedChunks`, a single-slot
  chunk queue.
- `PresentationManager` now holds at most one backlog chunk, plus one additional
  active chunk being presented and advanced frame by frame.
- A full pending chunk queue is treated as backlog, so `advance()` is allowed to
  proceed immediately and drain instead of waiting on the normal cadence.
- When not backlogged, `_PresentationClock` paces `advance()` from the recently
  observed model FPS with a drain margin, letting the UI thread recover slightly
  faster after hiccups and reducing model-frame latency.
- Removed the user-facing pending-buffer size knob and updated runtime docs,
  trace fields, and focused tests for chunk-based backlog semantics.

## Notes

- Drop and reset-discard counters are chunk-based. One chunk containing multiple
  model frames counts once.

Closes #542

### DIFF
--- a/apps/cam2v/tests/test_application.py
+++ b/apps/cam2v/tests/test_application.py
@@ -458,6 +458,11 @@ def test_slangpy_overlay_tracks_controls_and_model_status() -> None:
     """Keep immediate input display in UI-loop-owned state."""
     state = Cam2VUIState(total_blocks=4, target_fps=16, warmup_blocks=1)
     presentation_manager = PresentationManager()
+    presentation_manager.configure(
+        backpressure_mode=BackpressureMode.BLOCK,
+        stop=threading.Event(),
+        put_timeout=0.01,
+    )
     presentation_manager.publish(
         0,
         [
@@ -469,7 +474,7 @@ def test_slangpy_overlay_tracks_controls_and_model_status() -> None:
             )
         ],
     )
-    assert presentation_manager.advance(0)[0]
+    assert presentation_manager.advance(0, now=1.0)[0]
     ui = SimpleNamespace(
         screen=object(),
         Window=Mock(return_value=object()),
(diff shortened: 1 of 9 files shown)


gtong-nv:
/ok to test [20fd998](https://github.com/NVIDIA/flashdreams/pull/544/commits/20fd9980bf726a6b599aa3dddd8090c446dd926c)

ArielG-NV (approved):
Tested changes locally + Code looks good

PR #545, opened 2026-08-29
UI

3a1b2c3:


flashdreams/flashdreams/serving/webrtc/manager.py
```diff
@@ -1096,13 +1096,15 @@ async def _handle_event_message(
             # the mapping turns it into a session-global conditioning update
             # applied by the next step, so there is no separate runtime call.
             clears = state in clear_states
+            raw_prompt = payload.get("prompt")
```
ArielG-NV, line 1099: We will be no longer using this code, the new code we will be using moving forward is in `runtime_v2`

integrations/lingbot/lingbot/demo/providers.py
```diff
@@ -49,6 +49,11 @@
 _CLEAR_STATES = frozenset({"clear", "release", "off", "none"})
 _TRIGGER_STATES = frozenset({"trigger", "hold", "on"})
 
+_USER_PROMPT_EVENT_ID = "user_prompt"
+"""Reserved event id for a free-form custom prompt, not a catalog entry --
+must match the literal in ``lingbot/webrtc/session.py``,
+``lingbot/input_mapping.py``, and ``lingbot/webrtc/web/adapter.js``."""
+
```
ArielG-NV, line 56: We will be no longer using the `integrations` folder in the future.

The plan is to move all `model` implementations to `integrations_v2`, with all applications made "generic" and put in `apps`.

lingbot uses `cam2v`.

ArielG-NV (commented):
Notes regarding updates to Flashdreams

PR #548, opened 2026-08-31
Add multi-session lifecycle support to the v2 runtime

### LINKED ISSUE #502: [Session Runner] Make IO-Thread/UI-Thread reusable and able to restart (end_old+start_new) a `ISession`
We have a goal where io-thread may receive from client-window a `resetSession` request. If this occurs, we should end our ISession and re-create it (and everything should 'just-work'), and keep a persistent client-window alive to manage all of this.

gtong-nv:
## Summary

- Add multi-session lifecycle support while keeping the application, model
  weights, and client window alive between sessions.
- Add an ImGui prompt UI for T2V applications that can request new sessions
  before or after inference.
- Keep UI processing active after model completion and across WebRTC
  reconnects.
- Make `IModelLoop` the single source of inference lifecycle state.
- Suppress the known non-fatal `aioice` STUN retry race without hiding other
  asyncio errors.
  
  fixes https://github.com/NVIDIA/flashdreams/issues/502

### DIFF
--- a/flashdreams/flashdreams/api_v2/client_window.py
+++ b/flashdreams/flashdreams/api_v2/client_window.py
@@ -14,9 +14,8 @@ class IClientWindow(InputSource, OutputSink, ABC):
 
     The runtime opens the window with a session's description, then reads input
     and writes results until the run ends. A window stays open across a session
-    reset. A reusable window may also stay open when the client asks to replace
-    the session; the runtime opens it again for the replacement and closes it
-    when the application run ends.
+    reset/replacement. A window may be opened/closed by the runtime during a session
+    reset/replacement.
 
     A window does not describe the output shape. The session does, and the window
     is given that description in :meth:`OutputSink.open`.
--- a/flashdreams/flashdreams/api_v2/client_window.py
+++ b/flashdreams/flashdreams/api_v2/client_window.py
@@ -12,9 +12,11 @@ from .output_sink import OutputSink
 class IClientWindow(InputSource, OutputSink, ABC):
     """Handle application input and output for one client window.
 
-    The runtime opens the window with the session's description, then reads input
-    and writes results until the run ends, and closes it then. A window stays open
-    across a session reset.
+    The runtime opens the window with a session's description, then reads input
+    and writes results until the run ends. A window stays open across a session
+    reset. A reusable window may also stay open when the client asks to replace
+    the session; the runtime opens it again for the replacement and closes it
+    when the application run ends.
 
     A window does not describe the output shape. The session does, and the window
     is given that description in :meth:`OutputSink.open`.
--- a/flashdreams/flashdreams/api_v2/loop.py
+++ b/flashdreams/flashdreams/api_v2/loop.py
@@ -302,7 +302,12 @@ class IModelLoop(ILoop[StateT], ABC):
                 run = self._begin_run(events, generation)
                 if run.step_index is None:
                     break
-                last_run_started = self._pace(last_run_started)
+                if self.frequency != 0 and last_run_started is not None:
+                    earliest_start = last_run_started + 1.0 / self.frequency
+                    self._shutdown_event.wait(
+                        max(0.0, earliest_start - time.monotonic())
+                    )
+                last_run_started = time.monotonic()
                 if self._shutdown_event.is_set():
                     break
                 step_started_at = time.monotonic()
--- a/flashdreams/flashdreams/api_v2/loop.py
+++ b/flashdreams/flashdreams/api_v2/loop.py
@@ -113,6 +113,11 @@ class ILoop(ABC, Generic[StateT]):
         self._shutdown_event = shutdown_event
         self._failure_queue = failure_queue
         self._new_session_request: dict[str, Any] | None = None
+        self._initialize_loop_state()
+
+    def _initialize_loop_state(self) -> None:
+        """Initialize state specific to this kind of loop."""
+        return
 
     @abstractmethod
     def step(
@@ -257,6 +262,24 @@ class IModelLoop(ILoop[StateT], ABC):
     """
 
     @final
+    def _initialize_loop_state(self) -> None:
+        """Initialize model-inference state."""
+        self._inference_state = ModelInferenceState.NOT_STARTED
+        self._inference_state_lock = threading.Lock()
+
+    @property
+    @final
+    def inference_state(self) -> ModelInferenceState:
+        """Return whether this model loop has started, is running, or finished."""
+        with self._inference_state_lock:
+            return self._inference_state
+
+    def _set_inference_state(self, state: ModelInferenceState) -> None:
+        """Set this model loop's inference state."""
+        with self._inference_state_lock:
+            self._inference_state = state
+
+    @final
     def _run_model_loop(
         self,
         *,
@@ -276,6 +299,7 @@ class IModelLoop(ILoop[StateT], ABC):
         """
         steps_run = 0
         last_run_started: float | None = None
+        self._set_inference_state(ModelInferenceState.RUNNING)
         try:
             while not self._shutdown_event.is_set() and (
                 max_steps is None or steps_run < max_steps
@@ -297,6 +321,7 @@ class IModelLoop(ILoop[StateT], ABC):
         except BaseException as error:
             self._failure_queue.put(error)
         finally:
+            self._set_inference_state(ModelInferenceState.FINISHED)
             try:
                 self._shutdown()
             except BaseException as error:
@@ -326,21 +351,17 @@ class IUILoop(ILoop[StateT], ABC):
         """
         self.output_layout = output_layout
         self._presentation_manager = presentation_manager
-        self._model_inference_state = ModelInferenceState.NOT_STARTED
-        self._model_inference_state_lock = threading.Lock()
+
+    @final
+    def _set_model_loop(self, model_loop: IModelLoop[Any]) -> None:
+        """Store the model loop whose inference lifecycle this UI observes."""
+        self._model_loop = model_loop
 
     @property
     @final
     def model_inference_state(self) -> ModelInferenceState:
         """Return whether model inference has started, is running, or finished."""
-        with self._model_inference_state_lock:
-            return self._model_inference_state
-
-    @final
-    def _set_model_inference_state(self, state: ModelInferenceState) -> None:
-        """Set the model-inference state visible to this UI loop."""
-        with self._model_inference_state_lock:
-            self._model_inference_state = state
+        return self._model_loop.inference_state
 
     @final
     def request_new_session(self, metadata: Mapping[str, Any]) -> None:
--- a/flashdreams/flashdreams/api_v2/loop.py
+++ b/flashdreams/flashdreams/api_v2/loop.py
@@ -9,15 +9,20 @@ import queue
 import threading
 import time
 from abc import ABC, abstractmethod
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Generic, TypeVar, final
+from enum import Enum
+from typing import TYPE_CHECKING, Any, Generic, TypeVar, final
 
 from torch import Tensor
 
 from flashdreams.runtime_v2.event_buffer import EventBuffer
 from flashdreams.runtime_v2.step_result import StepResult
-from flashdreams.runtime_v2.user_input_event import CloseUserInputEvent
+from flashdreams.runtime_v2.user_input_event import (
+    CloseUserInputEvent,
+    NewSessionUserInputEvent,
+    UserInputEvent,
+)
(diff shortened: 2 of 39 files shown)


flashdreams/flashdreams/api_v2/loop.py
```diff
@@ -27,6 +32,33 @@
 StateT = TypeVar("StateT")
 
 
+class ModelInferenceState(Enum):
```
gtong-nv, line 34: We need this because the UI has to remain active while the inferencing is not running, for the `ON_DEMAND` presentation mode.
Another option is to remove `ON_DEMAND` entirely and make UI always tick. But it's still good for the app to know the status of the model inferencing in my opinion.

Should we keep these in loop.py, or move it to somewhere else like the PresentationManager?

gtong-nv:
/ok to test [cc238c2](https://github.com/NVIDIA/flashdreams/pull/548/commits/cc238c2e3dcf861f894fae5167f04415aca1a65b)

flashdreams/flashdreams/api_v2/client_window.py
```diff
@@ -12,9 +12,11 @@
...
     """Handle application input and output for one client window.
 
-    The runtime opens the window with the session's description, then reads input
-    and writes results until the run ends, and closes it then. A window stays open
-    across a session reset.
+    The runtime opens the window with a session's description, then reads input
+    and writes results until the run ends. A window stays open across a session
+    reset. A reusable window may also stay open when the client asks to replace
+    the session; the runtime opens it again for the replacement and closes it
+    when the application run ends.
```
ArielG-NV:
```suggestion
    The runtime opens the window with a session's description, then reads input
    and writes results until the run ends. A window stays open across a session
    reset/replacement. A window may be opened/closed by the runtime during reset/replacement of a session.
```

simplify comment

flashdreams/flashdreams/api_v2/loop.py
```diff
@@ -27,6 +32,33 @@
...
+    """Describe whether a loop should step or yield lifecycle control."""
+
+    step_index: int | None = None
+    """Index to step, or ``None`` when no step should run."""
+
+    stop_requested: bool = False
+    """Whether the session should stop without a replacement."""
+
+    new_session_request: dict[str, Any] | None = None
+    """Metadata for a replacement session, or ``None`` when none was requested."""
```
ArielG-NV: This maps to `session_desc.metadata` for our next-session, we can make this clearer with: `next_session_metadata`/`new_session_metadata`

flashdreams/flashdreams/api_v2/loop.py
```diff
@@ -189,6 +237,7 @@ def _run_message_batch(self) -> None:
             if result is not None:
                 raise TypeError("Message operations must return None.")
 
+    # TODO: We should do pace here. Remove this func
```
ArielG-NV: I don't fully understand the comment here, maybe link to issue with longer explanation of task.

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -83,82 +82,106 @@ def run_session(
...
+                    session_desc,
+                    metadata={
+                        **session_desc.metadata,
+                        **loopResult.new_session_request,
+                    },
+                )
+                stop.set()
+                return
+            if loopResult.step_index is None or not step_requested:
+                return
```
ArielG-NV: We can begin a run, and then never finish the run or `stop` the `ui_loop`? We should be very careful about introducing cases where we can `begin` something but never `end` them.

We should potentially think about even making `stop` flag setting a type of `_finish_run` for clarity reasons. A run should likely be forced to have a begin/end for debugging/code-readability/code-extensibility reasons.

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -204,21 +225,26 @@ def tick_ui() -> None:
...
-            # or until the model finishes and no generated frames remain.
-            # A finished UI loop produces no further window output.
             while not stop.is_set():
                 if (
                     not model_thread_handle.is_alive()
                     and not presentation_manager.has_pending_frames()
+                    and (
+                        not session._failure_queue.empty()
+                        or steps is not None
+                        or ui_loop.is_finished()
```
ArielG-NV: `            loopResult = ui_loop._begin_run(events, generation)`

Our ui_loop can finish, we can have errors, and we can have `None` steps, but then run `_begin_run`?
This may make sense in terms of "we may finish but then get a request to restart right at the end", but otherwise this seems like a sign that we need to redesign the structure of our UI loop a bit to not be such a complex branch of cases.

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -204,21 +225,26 @@ def tick_ui() -> None:
...
+                    )
                 ):
+                    # Input may have arrived with the final presented frame.
+                    # Establish the terminal boundary before deciding that a
+                    # naturally completed session has no replacement.
+                    collect_input()
+                    run_ui_once(step_requested=False)
```
ArielG-NV: I assume this is to handle stop/reset requests during some sort of situation where model thread is not alive yet and we are not presenting- although I don't entirely understand the point of the code here.

flashdreams/flashdreams/t2v_v2/README.md
```diff
@@ -44,7 +57,7 @@ After the `--`, the same for every model, and listed by
 
 | Argument | |
 | --- | --- |
-| `--prompt` | Text to generate from. Required. |
+| `--prompt` | Optional initial text to generate from. The UI supplies prompts for later sessions. |
```
ArielG-NV: should only be optional if ui is enabled

flashdreams/flashdreams/api_v2/loop.py
ArielG-NV: I am not too worried about where we store the enum.

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -83,82 +82,106 @@ def run_session(
...
+            if loopResult.new_session_request is not None:
+                next_session_desc = replace(
+                    session_desc,
+                    metadata={
+                        **session_desc.metadata,
+                        **loopResult.new_session_request,
+                    },
+                )
+                stop.set()
+                return
```
ArielG-NV: This means that `new_session_request` is a union of old metadata (`**session_desc.metadata,`) with new metadata (with priority for new metadata). We should be very clear about behavior here- `new_session_request` only overwrites metadata provided, it does not remove.

This introduces an issue though... how do I "remove" a key-value pair entirely from the new session? Is this disallowed?

ArielG-NV (commented):
Comments on design.

flashdreams/flashdreams/api_v2/client_window.py
gtong-nv: updated

flashdreams/flashdreams/api_v2/loop.py
gtong-nv: typo: We should NOT do pace here.
The pace already handled by `PresentationManager` which knows both ModelLoop and UILoop cadence. I'm going to delete this function in the new commit.

flashdreams/flashdreams/runtime_v2/session_runner.py
gtong-nv: We are handling new session request and other system events in `_begin_run`.
This branch is indeed too complex and hard to understand.
Let me see if I can simplify it a bit

gtong-nv: I moved it to a helper function with better naming and docs. No semantic changes yet.
We can revisit this if needed.
```
            def should_end_ui() -> bool:
                if model_thread_handle.is_alive():
                    return False
                if presentation_manager.has_pending_frames():
                    return False

                if (
                    session._failure_queue.empty()
                    and steps is None
                    and not ui_loop.is_finished()
                ):
                    # If there are no failures, no steps limit, and the UI is not finished,
                    # the run should continue.
                    return False
                collect_input()
                run_ui_once(step_requested=False)
                return True
```

flashdreams/flashdreams/runtime_v2/session_runner.py
gtong-nv: hmm, I guess this is indeed an issue. We should be able to create a completely new session desc. Updating

flashdreams/flashdreams/runtime_v2/session_runner.py
gtong-nv: > We can begin a run, and then never finish the run or stop the ui_loop

For serving, this is actually a valid use case. If we want to stop, we just kill the server/app.
For debugging purpose, yea, it's better to have a flag to stop the UI run even for the unbounded interactive case.
But how should we design the flag? something like `--timeout 60` for 1-min run and stop the app entirely?

flashdreams/flashdreams/t2v_v2/README.md
gtong-nv: I had this originally, but this requires an additional field in session_desc such as `accepts_user_input`.
The reason is that "--prompt" is parsed by this t2v app, and it has no knowledge if the output mode is `webrtc` or `mp4` (or if ui is enabled).

Even with that, a missing MP4 prompt is rejected in `create_session()`, after `application.init()` has already loaded the model.

I would not add this back until we have a better way to check whether UI/input is available from app's perspective.

flashdreams/flashdreams/runtime_v2/session_runner.py
ArielG-NV: `_finish_run` should really just be in a `finally:` block I suppose... and if this is not possible then something is wrong (begin at entry, finish at end)

flashdreams/flashdreams/t2v_v2/README.md
ArielG-NV: 👍

flashdreams/flashdreams/runtime_v2/session_runner.py
gtong-nv: oh I thought you were referring "launching the app, and let the app run forever".
yea, if what we want is matching `_begin_run` and `_finish_run`, this makes sense. I will update that

gtong-nv:
/ok to test [5a0933b](https://github.com/NVIDIA/flashdreams/pull/548/commits/5a0933bc5430bdf6852dd34dd190dfb637987844)

flashdreams/flashdreams/runtime_v2/session_runner.py
gtong-nv: Moved `_finish_run` to `finally:` block

flashdreams/flashdreams/api_v2/loop.py
gtong-nv: Updated. now `new_session_request` is just `session_desc`

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -56,109 +49,148 @@ def run_session(
...
     Args:
         session: Session to run.
         window: Source of input and destination for UI output.
         metrics_output_sink: Sink for model measurements, if requested. Receives
             the model loop's results rather than the UI loop's.
-        steps: Maximum model steps; ``None`` runs until stopped.
+        steps: Maximum model steps before ending the session; ``None`` leaves
+            session completion to the UI or client window.
+        timeout_seconds: Maximum session runtime; ``None`` waits for another
+            lifecycle exit. Expiration signals both registered loops to stop.
```
ArielG-NV:
```suggestion
        timeout_seconds: Maximum session runtime; ``None`` means session does not have a time-limit. Timeout expiry signals both registered loops to stop.
```

flashdreams/flashdreams/api_v2/loop.py
```diff
@@ -79,6 +112,13 @@ def register_session_loop_objects(
         self._lifecycle_lock = threading.Lock()
         self._shutdown_event = shutdown_event
         self._failure_queue = failure_queue
+        self._new_session_request: SessionDesc | None = None
+        self._new_session_event_desc: SessionDesc | None = None
```
ArielG-NV: we don't need both

ArielG-NV (commented):
Minor suggestions; PR looks very good now.

flashdreams/flashdreams/api_v2/loop.py
gtong-nv: addressed

gtong-nv:
/ok to test [7750f1c](https://github.com/NVIDIA/flashdreams/pull/548/commits/7750f1c2b47fc2cc52c51029dda3f446a28a5973)

ArielG-NV (approved):
PR looks good now, just need to fix failing tests.

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/548/commits/5d48dcceea6200779b97a18c5753ba6129668a5f

PR #560, opened 2026-09-01
Reduce Robotaxi input sampling latency

gtong-nv:


flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -106,6 +106,20 @@ def run_session(
 
     def collect_input() -> None:
         events = window.get_user_input_events()
+        if trace_chunk_lifecycle:
```
gtong-nv, line 109: this is for debug only. Removing

Issue #479, opened 2026-08-18
Porting Demos & Underlying Integrations To V2 API & File-Structure


jarcherNV:
# Port remaining demos to the FlashDreams application/session API

## Summary

The newer public demo API is the `flashdreams.demo` application/session contract (`IFlashDreamsApplication`, `IFlashDreamsApplicationSession`, `run_application`) with host-owned MP4, null, local-window, and WebRTC I/O. The host invokes the unified runtime internally; application/session API coverage is the only migration status tracked by this issue.

Port the two remaining plain T2V integration families and add application adapters for the already-runtime-migrated OmniDreams and Lingbot demos. Do **not** include OmniDreams Interactive Drive in this issue.

## Current status

| Area | Application/session API | Evidence |
| --- | --- | --- |
| Cosmos Predict2 T2V | Complete | `t2v-cosmos-predict2` application entry point |
| Causal Forcing T2V | Complete | `t2v-causal-forcing` application entry point |
| Self Forcing T2V | Complete | `t2v-self-forcing` application entry point |
| Wan 2.1 T2V | Missing | `wan21-t2v-1.3b-480p` has no application entry point |
| FastVideo CausalWan 2.2 T2V | Missing | `fastvideo-causal-wan2.2-t2v-14b` has no application entry point |
| OmniDreams demo (excluding Interactive Drive) | Missing | No `flashdreams.applications` entry point |
| Lingbot demo | Missing | No `flashdreams.applications` entry point |

The three existing T2V application packages are the intended reference: their integration owns a small `t2v/app.py` factory, depends on `flashdreams-t2v`, and registers a `flashdreams.applications` entry point.

### 1. Add the missing plain T2V applications

- Add `t2v` application factories and `flashdreams.applications` registrations for Wan 2.1 T2V and FastVideo CausalWan 2.2 T2V.
- Reuse `apps/t2v`'s `T2VApplication` / `T2VApplicationSession` where the model's pipeline contracts permit it; keep integration-specific defaults in the integration package.
- Seed defaults from the existing runner configs: `RUNNER_WAN21_T2V_1PT3B_480P` and `RUNNER_WAN22_T2V_14B`.
- Preserve the current prompt, resolution, FPS, block-count, decoder layout, compile override, and checkpoint/pipeline behavior. The new applications must stream every autoregressive block through the shared output path rather than call the legacy runner's `run()` method.
[...]

PR #561, opened 2026-09-01
[Demos] Refactor+Move flashvsr into v2

ArielG-NV:
Related to larger porting effort: #479
- flashvsr + new upscaling demo based on big-buck-bunny

Refactor+Move flashvsr into v2. Changes should be 'minimal' to accomplish this task. Core API changes were made to `IModelLoop::step` to be explicit on the return type to remove the nonsensical return type of `StepResult` vs. `None` vs `List[StepResult]` (Full work is for another issue/task, current work is to unblock regarding type-checking error: #562).


ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/561/commits/ac589702ac1d25807d65843af5e88eaeb8f5c32a

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/561/commits/5b7b570780f0119f13d92be50264410667ab578d

integrations/flashvsr/flashvsr/grpc/uplift_server.py
```diff
@@ -1,1430 +0,0 @@
...
-``upsampler.initialize_cache()``.
-
-``upscale_video`` pipelines receive/decode → buffer/coalesce → GPU → send/viewer:
-a reader thread decodes and enqueues incoming chunks (FIFO), a worker optionally frame-coalesces
-8-frame ingest into FlashVSR's 13-frame cold start plus steady 16-frame calls,
-and bounded queues provide backpressure plus paced browser playback.
-
-Usage (from repo root):
-
-    uv run --no-sync python -m flashvsr.grpc.uplift_server --viewer_port 8080
```
gtong-nv, line 31: This by itself is a standalone service that provides the uplifting capability. In my opinion we should not remove this as the new runtime API is not likely to provide the same service.
We can probably find another place to host it, but until then, better to keep the `uplift_server` for demo purpose

gtong-nv:
The new `apps/upsample_video` does not take a custom input video now and just use a predefined URL to upsample? Feels like not very useful as an app.
I'm thinking we can have a video to video app, that uses `flashVSR` as one of the integrations

integrations/flashvsr/flashvsr/grpc/uplift_server.py
ArielG-NV: I will keep it in `impl` then, a catch all location I suppose for stuff we need to organize isolated to a particular model implementation

ArielG-NV:
Note: Decided to separate ti2v from t2v for wan22

gtong-nv (approved):
LGTM

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/561/commits/5d8f98e8ef4e3192be02a067704856895d31f436

PR #566, opened 2026-09-02
[crazy-robotaxi] Add responsive OmniDreams history presets

aidanfnv:
# Responsive OmniDreams history for Crazy Robotaxi

## Summary

Crazy Robotaxi can update the simulated vehicle trajectory without the generated view following the corresponding turn. The failure is intermittent, including across runs with similar speed and steering input, and occurs with both the regular and native-DiT-accelerated OmniDreams configurations.

This PR adds an opt-in mixed-history mode to the repo-local OmniDreams integration. The first nine transformer blocks retain one autoregressive chunk of visual history, while the remaining nineteen blocks retain a four-latent-step rolling window. The goal is to make early blocks respond more strongly to current camera motion without discarding the longer appearance history used by later blocks.

Five Crazy Robotaxi runner presets enable mixed history:

- `crazy-robotaxi-omnidreams-responsive`
- `crazy-robotaxi-omnidreams-perf-responsive`
- `crazy-robotaxi-omnidreams-fast-perf-responsive`
- `crazy-robotaxi-omnidreams-optimized-gb300-responsive`
- `crazy-robotaxi-omnidreams-optimized-rtx-pro-6000-responsive`

The existing presets and the default transformer behavior are unchanged.

## Why mixed history

The configuration was selected through repeated interactive driving on the Grand Prix course. These observations are qualitative because model output is stochastic.

| Configuration tested | Turn response | Visual behavior |
| --- | --- | --- |
| Native DiT enabled or disabled | Turning failures occurred in both paths | Native acceleration was not isolated as the cause |
| Uniform four-step history | Turning failures remained | Most coherent and stable uniform window tested |
| Uniform three-step history | Fewer observed failures than four steps | Intermediate stability with continual scene changes |
| Uniform two-step history | No observed failures in successful test runs | Highly unstable and often visually incoherent |
| Turn-triggered cache reset | Turn response improved after reset | Produced a blue flash and compounded existing degradation |
| Camera-trajectory scaling | Did not produce a reliable fix | Rejected after interactive testing |
| Four-step history with the first nine blocks limited to one chunk | No observed failures in the selected test runs | Best visual result tested without observed turning failures; nearby houses, cars, and trees remained acceptably coherent |

The rejected alternatives in this table are investigation results, not implementations included in this PR. The selected configuration gave the best observed balance between motion response and appearance stability.

## Implementation

`CosmosTransformerConfig` gains `early_short_history_block_count`. When set, cache construction gives the first N transformer blocks one chunk of rolling capacity and gives every later block the configured full window. The responsive presets use a chunk length of two latent steps, a full window of four latent steps, and nine short-history blocks out of twenty-eight.

The responsive presets pair mixed history with cache-relative rotary position embeddings. In this mode, self-attention stores unrotated keys and applies RoPE after reading the currently visible cache. Mixed history can also be used with the existing rotate-before-cache path, which remains the default.

The implementation validates the combinations it supports:

- The early-block count must be between one and the model's block count.
- Native DiT assumes uniform per-block cache geometry and does not support cache-relative RoPE, so these responsive presets disable it.
- The full window must be at least one autoregressive chunk.

Each responsive pipeline config derives from an existing standard, perf, fast-perf, optimized GB300, or optimized RTX PRO 6000 config. Derivation preserves the base scheduler, cache-finalization policy, seed behavior, VAE configuration, and attention policy. Consequently, fast-perf-responsive retains the native FP8 VAE, and the hardware-specific responsive presets retain their optimized attention backends and quantization settings.

The fix is confined to the repo-local OmniDreams attention and cache path plus preset registration. Crazy Robotaxi vehicle physics and camera trajectory generation are unchanged.

## Validation

CPU tests cover:

- Per-block cache allocation and rolling contents.
- Construction of cache-relative RoPE.
- Storage of unrotated keys and application of RoPE after cache reads.
- Selection of after-cache RoPE by the optimized attention backend.
- Rejection of cache-relative RoPE with native DiT.
- Inheritance and registration of all five responsive presets.
- Crazy Robotaxi adapter bindings.

The focused application and integration suite completed with 72 passing tests; one GPU-dependent test remained marked manual. `flashdreams-run-v2 --help` listed all five responsive runner names.

Manual evaluation of the selected fast-perf-responsive configuration used:

```bash
uv run --package flashdreams-omnidreams flashdreams-run-v2 \
  crazy-robotaxi-omnidreams-fast-perf-responsive --mode native-window -- \
  --map apps/crazy_robotaxi/crazy_robotaxi/maps/flashdreams_raceway.robotaxi.yaml \
  --game-mode race \
  --race-course grand-prix \
  --prompt "A forward-facing view from a taxi moving through a quiet suburban district in daylight, with low commercial buildings, houses, landscaping, and parked cars."
```

## Limitations

- The nine-block/four-step split was selected for the current 28-block, chunk-2, single-view OmniDreams checkpoint and should not be assumed optimal for another model.
- Visual quality remains prompt-sensitive and was evaluated interactively rather than with a quantitative quality metric.
- The optimized GB300 and RTX PRO 6000 responsive presets have config and CPU-test coverage but have not received separate end-to-end generation runs on their target GPUs.
- Responsive presets cannot use native DiT acceleration until that implementation supports cache-relative RoPE and heterogeneous per-block history lengths.
- The absence of observed turn failures does not prove that stochastic failures are impossible.


integrations_v2/omnidreams/impl/transformer/network.py
```diff
@@ -429,16 +433,38 @@ def initialize_cache(
...
         if self.config.use_crossattn_projection:
             context = self.crossattn_proj(context)
 
         block_caches: list[BlockCache] = []
-        for block in self.blocks:
+        short_history_count = early_short_history_block_count or 0
+        for block_index, block in enumerate(self.blocks):
             assert isinstance(block, Block)
+            use_short_history = block_index < short_history_count
+            block_window_size = chunk_size if use_short_history else window_size
```
fangjunzhou-nv, line 460: If I understand correctly, the idea here is for the first `short_history_count` DiT block we only use one chunk window? Should we make this configurable so we have `window_size` and `short_window_size`. And for the first `early_short_history_block_count` block we use `short_window_size` instead of always make `short_window_size = chunk_size`

fangjunzhou-nv (approved):
Most changes looks good. This comment is just a suggestion for future improvement. It's not an issue to be fixed rn.

fangjunzhou-nv:
@aidanfnv make sure the optimized runner config runs on both platform before merge. Ping me if any of them doesn't work.

integrations_v2/omnidreams/impl/transformer/network.py
aidanfnv: Yes, I think it should be made more configurable. At the last minute I collapsed down some settings that I didn't need to tweak for my demo (and would cause some conflicts with it) into the new presets. Not ideal, but like you said it can be a future improvement.

PR #570, opened 2026-09-02
Persist torch.compile cache across GPU CI jobs

### LINKED ISSUE #500: Compile time cache for `torch.compile` in flashdreams
`torch.compile` is taking a very long time when flashdreams boots a large model. This impact benchmarks and CI significantly and it's wasting our CI capacity.

We need to investigate this and improve the model compile time.

I believe the Triton autotune is cached already but `torch.compile` is not.

Possible solution: https://docs.pytorch.org/tutorials/recipes/torch_compile_caching_tutorial.html

spaingnv:
## Summary
- GPU CI now restores and saves `$FLASHDREAMS_CACHE_DIR/torchinductor`, keyed by GPU arch and `uv.lock`.
- `compile_module` already pinned Inductor under that path locally; this is what was missing for CI.
- Add a `ci_gpu` two-process check that a warm compile is at least twice as fast as a cold one.

Fixes #500.

## Test plan
- [x] `pytest flashdreams/tests/test_compile.py -m ci_cpu`
- [x] `pytest flashdreams/tests/test_compile_cache_gpu.py -m ci_gpu` on RTX PRO 6000 (cold ~6s, warm ~1s)
- [ ] First GPU CI job on this PR (cold compile + cache upload)
- [ ] A later GPU job on the same arch/`uv.lock` shows an inductor cache restore

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,8 @@ jobs:
       OMNIDREAMS_QUALITY_REFERENCE_PATH: ${{ vars.OMNIDREAMS_QUALITY_REFERENCE_PATH || 'data/quality_references/single_view/omnidreams-sv-2steps-chunk2-loc6-lightvae-lighttae/239560dc-33d1-11ef-9720-00044bcbccac/reference_compare_region.mp4' }}
       OMNIDREAMS_QUALITY_REFERENCE_CLIP: /tmp/omnidreams_quality_reference/reference_compare_region.mp4
       OMNIDREAMS_QUALITY_ARTIFACT_DIR: /tmp/omnidreams_quality_artifacts
+      # Persistent Inductor/FX-graph cache for torch.compile across GPU CI jobs.
+      FLASHDREAMS_CACHE_DIR: /tmp/flashdreams-cache
     steps:
       - name: Get PR info
         if: startsWith(github.ref_name, 'pull-request/')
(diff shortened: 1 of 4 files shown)


fangjunzhou-nv (changes requested):
#500 is not a CI specific issue. This issue happens to torch.compile even outside the CI. Specifically, running `compile_module` from `flashdreams/flashdreams/infra/compile.py` usually takes a long time.

This indicates we're not using torch.compile cache properly and we need to investigate this before we propose a fix.

spaingnv:
Agreed — the CI cache was the wrong problem. `compile.py` already pins `TORCHINDUCTOR_CACHE_DIR`; a tiny MLP hitting that cache does not explain a slow real `compile_module` / first forward.

Converting to draft. Next I will measure a real OmniDreams compile (wrap vs first forward vs CUDA-graph warmup, cold vs second process) and report cache hit/miss before proposing a fix.

jmccaffrey-nv:
The change reuses the existing cache-root abstraction, stays narrowly scoped, and the added cache_dir() assertion is useful.

[P1 correctness/reliability] The GPU test can pass without a cache hit.  It writes to a fresh tmp_path, while CI persists /tmp/flashdreams-cache/torchinductor; moreover, the assertions only prove files remain, not that the warm worker reused them ([lines 61–103 flashdreams/tests/test_compile_cache_gpu.py#L61-L103))   . Can the test assert PyTorch’s fxgraph_cache_hit counter?  Test the desired outcome directly. .

Comment P1 security: GitHub also recommends treating restored cache contents as untrusted and keeping secrets out of cached paths ([GitHub cache security guidance](https://docs.github.com/en/actions/concepts/workflows-and-actions/dependency-caching)).

PR #574, opened 2026-09-03
Add correlated V2 runtime latency profiles

### LINKED ISSUE #538: [API] User Input Latency Instrumentation
[API] User Input Latency Instrumentation Instrumentation

ZenAlexa:
## Summary

I added `flashdreams-run-v2 --profile-path` for host-side input-latency JSONL profiles. `TimestampedInputSource` supplies the session clock origin; each session segment includes its runtime settings and measurement endpoints.

Closes #538.

## Measurements

- `input_to_ui_step_s`: input timestamp to the IUILoop claiming the event.
- `input_to_window_write_s`: input timestamp to the first following `IClientWindow.write` return. Native ends after the presenter call. WebRTC ends at the host write boundary, including queue admission when a video track is available.

Summaries retain exact counts and maxima, with a bounded 1,024-sample reservoir for median and p90. Replacement sessions receive independent profile segments.

## Validation

The V2 CPU suite passed: 218 tests, 3 deselected. Focused runtime coverage includes timestamp correlation, bounded summaries, session replacement, and profile-path conflicts.


flashdreams/flashdreams/runtime_v2/serving/webrtc_server.py
```diff
@@ -412,9 +420,9 @@ def write(self, result: StepResult) -> None:
             )
         track = self._video_track
         if track is None:
-            return
+            return False
         queued_frame = self._materialize_video_frame(result, frames[0])
-        track.enqueue(queued_frame)
+        return track.enqueue(queued_frame)
```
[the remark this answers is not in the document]
ZenAlexa: Fixed in `1b6cc195`. `_VideoTrack.set_sender_available()` now shares `_state_lock` with `enqueue()`, so an explicit `disconnected` transition rejects subsequent admissions and preserves profiler correlations. Reaching `connected` reopens admission, and negotiation-time queuing remains intact; the deterministic disconnect/reconnect regression and the full CPU suite pass (190 passed, 3 deselected).

flashdreams/flashdreams/api_v2/client_window.py
```diff
@@ -24,3 +24,13 @@ class IClientWindow(InputSource, OutputSink, ABC):
...
+
+    @property
+    def profile_endpoint(self) -> str | None:
+        """Name the observable boundary reached when ``write`` returns."""
+        return "window_write_return"
+
+    @property
+    def input_timestamp_origin_ns(self) -> int | None:
+        """Return the host monotonic origin for input-event timestamps."""
+        return None
```
ArielG-NV: I do not believe this is something that should be a part of the `IClientWindow` since this IClientWindow is supposed to be 3 parts:
- user pulls input from `InputSource` ==> likely where timestamps/telemetry will be pulled from (as a UserInputEvent in a special buffer in the same way `event_buffer` works?)
- user sends data to `OutputSource` for presentation
- user can request Client Window options via `IClientWindow` (this is not confirmed as a use-case, only a potential future for how we will modify the interface via addition of features)

ArielG-NV (commented):
Comment

flashdreams/flashdreams/api_v2/client_window.py
ZenAlexa: Yep, I see the ownership boundary here now 👍

I moved the clock bridge onto a dedicated `TimestampedInputSource` extension and removed both profiling properties from `IClientWindow`. I kept the source clock origin explicit because `UserInputEvent.timestamp` is session-relative; that origin makes the runtime delta valid. The profiler now covers the two perceived-latency boundaries from #538, and the transport-specific phase machinery is gone.

ZenAlexa:
Good catch to check this race. `input_to_window_write_s` deliberately ends at the existing `IClientWindow.write` return; for WebRTC that is the host's single-slot sender-mailbox write. Active-peer delivery, RTP transit, browser decode, composition, and display live beyond this host-side sample and need client telemetry. `_sender_available` changes transport queuing behavior; client telemetry supplies the presentation timestamp, so that state stays outside this profiling patch.

ZenAlexa:
Pulled #548's multi-session lifecycle into `5a90e4e9` and followed its new session ownership through the profiler path.

Each replacement now gets a fresh clock binding and an independent JSONL segment, with WebRTC's rebased timestamps mapped back to the correct monotonic session origin. The complete V2 CPU suite passes: 211 passed, 3 deselected. The two metrics remain anchored at IUILoop claim and the first following window write (ง •̀_•́)ง

ZenAlexa:
I rechecked this against current HEAD and the PR diff. `input_to_window_write_s` intentionally ends when the existing `IClientWindow.write` call returns; for WebRTC that records host-side materialization and mailbox admission, including the existing disconnected-state behavior. Active-peer delivery, RTP transit, decode, composition, and display require client timestamps, so they remain a separate telemetry continuation.

This review keeps the transport contract unchanged and keeps #574 scoped to the two host-side perceived-latency checkpoints (•̀ᴗ•́)و

ZenAlexa:
Synced current `main` through #584 in `0e007995` and resolved the new `UILoopRequests` flow at the profiling boundary. The V2 CPU suite completes with 215 passed / 3 deselected; Ruff, focused `ty`, compileall, and diff checks pass. The WebRTC endpoint description now follows #579's bounded two-frame sender queue.

flashdreams/flashdreams/runtime_v2/runtime_profiler.py
```diff
@@ -0,0 +1,262 @@
...
+        self._output = self._path.open(mode, encoding="utf-8")
+        self._opened_once = True
+        self._input_timestamp_origin_ns = input_timestamp_origin_ns
+        self._input_generation = None
+        self._pending_inputs.clear()
+        self._samples = {
+            "input_to_ui_step_s": [],
+            "input_to_window_write_s": [],
+        }
+        self._write(
```
jmccaffrey-nv, line 133: Could each session segment include enough metadata to identify its measurement boundary and runtime settings? Native ends at presenter return while WebRTC ends at host queue admission, but the records currently omit the client-window kind, FPS, resolution, presentation mode, and backpressure. Once a JSONL file is moved or aggregated, unlike measurements can be compared silently. Passing this context from `run_session` would preserve the current API ownership boundary.

-- reviewed using GPT-5.6 Sol

flashdreams/flashdreams/runtime_v2/runtime_profiler.py
```diff
@@ -0,0 +1,262 @@
...
+                    received_at_ns=received_at_ns,
+                    claimed_at_ns=started_at_ns,
+                    claim_duration_s=claim_duration_s,
+                    timestamp_us=int(event.get_timestamp()),
+                    event_type=type(event).__name__,
+                    generation=generation,
+                    ui_step=step,
+                )
+            )
+            self._samples["input_to_ui_step_s"].append(claim_duration_s)
```
jmccaffrey-nv: Could the summary use a bounded or streaming accumulator? `_samples` retains two Python floats per input until session close, and `_summarize` then sorts both full histories. A long-lived WebRTC session, or a high-rate client while profiling, can therefore grow memory without bound and make shutdown O(n log n), in addition to the expected JSONL disk growth.

-- reviewed using GPT-5.6 Sol

flashdreams/flashdreams/runtime_v2/runtime_profiler.py
```diff
@@ -0,0 +1,262 @@
...
+        for event in events.get_events():
+            received_at_ns = origin_ns + int(event.get_timestamp()) * 1_000
+            claim_duration_s = _elapsed_s(received_at_ns, started_at_ns)
+            self._pending_inputs.append(
+                _ClaimedInput(
+                    received_at_ns=received_at_ns,
+                    claimed_at_ns=started_at_ns,
+                    claim_duration_s=claim_duration_s,
+                    timestamp_us=int(event.get_timestamp()),
+                    event_type=type(event).__name__,
```
jmccaffrey-nv: Could this serialize `event.get_type_name()` instead? That API is the validated canonical event identifier; the Python class name is an implementation detail, so a rename or custom implementation can change schema-v1 values without changing the event type.

-- reviewed using GPT-5.6 Sol

jmccaffrey-nv (commented):
Reviewed the implementation against #538, including native/WebRTC clock bridging, replacement-session ownership, correlation, cleanup, path collisions, security, and profiler overhead.

The PR satisfies the narrowed host-side checkpoints discussed on the issue: event receipt to IUILoop claim, and event receipt to the next `window.write` return. The latter is not causal or end-to-end perceived response latency: the UI can re-render a held pre-input frame, and WebRTC stops at host queue admission before transport, decode, composition, and scanout. The documentation states these limits; please keep that distinction explicit when closing #538.

Local validation at `0e007995`: 215 V2 CPU tests passed; focused Ruff formatting/import checks, `ty`, compileall, and `git diff --check` passed. The documentation build reached the new section without a new warning; its warning-as-error run still reports 10 unrelated baseline warnings. GitHub currently shows only the successful Greptile check while NVIDIA runner validation awaits vetting.

I left three inline comments on self-describing profile metadata, unbounded in-memory summary retention, and stable event-type serialization. I found no new code-execution, deserialization, dependency, or credential-handling exposure.

-- reviewed using GPT-5.6 Sol

flashdreams/flashdreams/runtime_v2/runtime_profiler.py
ZenAlexa: Added in `020fddb4`. Each `session_started` record now carries an artifact type, both endpoint identifiers, the concrete client-window type, output layout, UI/model FPS, resolution, presentation mode, and backpressure mode. `run_session` supplies the resolved `SessionDesc`, so replacement sessions retain their own settings.

flashdreams/flashdreams/runtime_v2/runtime_profiler.py
ZenAlexa: Updated in `020fddb4`. The summary now keeps exact count/max and a 1,024-sample deterministic uniform reservoir for median/p90. Shutdown sorts at most 1,024 samples per metric, and each summary reports `quantile_sample_count` plus `quantiles_approximate`. Coverage exercises 1,100 paired samples.

flashdreams/flashdreams/runtime_v2/runtime_profiler.py
ZenAlexa: Updated in `020fddb4`. Profile records now serialize `event.get_type_name()`, and the regression asserts the canonical `keyboard` identifier.

ZenAlexa:
Addressed the three review threads in `020fddb4`: session segments now include artifact and runtime context; summary quantiles use a bounded 1,024-sample reservoir with explicit approximation metadata; input types use `UserInputEvent.get_type_name()`. Endpoint docs now cover WebRTC's active-track and no-track write behavior.

Validation: 185 V2 CPU tests passed, 5 skipped in the local optional-dependency environment; 88 focused tests passed, 1 skipped; Ruff, focused `ty`, compileall, and diff checks passed. Sphinx built all 38 sources with 10 pre-existing warnings; current `main` reproduces all 10. A 20,000-input JSONL probe completed at a median 6.14 µs per profiled input across three runs, with 1,024 quantile samples retained per metric.

ZenAlexa:
I've synced the profiling change with the new model-metrics sink on `main`.

Both outputs keep their session lifecycle, and replacement sessions retain independent profile segments. I've updated the validation section for the current branch.

PR #579, opened 2026-09-03
[Demos] Refactor+Move waypoint into v2

ArielG-NV:
Related to larger porting effort: https://github.com/NVIDIA/flashdreams/issues/479

Changes:
- Adds a **reusable `action2v` demo application** for interactive action-conditioned video models. It owns the shared session loop, input mapping, and (mouse) cursor setup so integrations only need to provide their model-specific adapter. This is just initial support, extensive controller support (etc...) can be managed later.

- Ports **Waypoint to the v2 API** using the new Action2V application. The Waypoint implementation, configuration, tests, and example controls are moved into the `integrations_v2/waypoint` dir.

- Extends the **`IClientWindow` and `IUILoop` protocols** with support for `hide_cursor` and `lock_cursor_to_window`. UILoop requests are now bundled all in one object called `UILoopRequest`.

- Adds runtime cursor control for native and WebRTC windows. Action2V requests a hidden cursor locked to the window through `invoke_async`. Default behavior is a visible not-locked-to-window cursor. WebRTC sends the initial cursor state when a client connects and only sends further updates when the effective options change.


ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/579/commits/20ba4b1e24e4cc136f1d8f91d2bc36bda81e2a99

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/579/commits/f1eef2a23b3308310dee2bda3f640533a87b3d80

apps/action2v/action2v/application.py
```diff
@@ -0,0 +1,267 @@
...
+        if self.pixel_width <= 0 or self.pixel_height <= 0:
+            raise ValueError("Action2VApplicationDefaults dimensions must be > 0.")
+        if self.fps <= 0 or (self.ui_fps is not None and self.ui_fps <= 0):
+            raise ValueError("Action2VApplicationDefaults frame rates must be > 0.")
+        object.__setattr__(self, "metadata", MappingProxyType(dict(self.metadata)))
+
+
+@dataclass(frozen=True, slots=True)
+class _ApplicationConfig:
+    first_frame: Path | None
```
gtong-nv: `first_frame` owned by application? I think it `Session` should own it so that we can create sessions with different first_frame

apps/action2v/action2v/application.py
```diff
@@ -0,0 +1,267 @@
...
+            type=int,
+            help="Non-negative model RNG seed; generated randomly when omitted.",
+        )
+        parser.add_argument(
+            "--device",
+            default=self.defaults.device,
+            help="Model device. Default: %(default)s.",
+        )
+        parser.add_argument(
+            "--profile",
```
gtong-nv: `--profile` should really be a cli entry arg, so that all apps can share it. 
Unless this app has its specific profiling logic

apps/action2v/action2v/application.py
```diff
@@ -0,0 +1,267 @@
...
+        self._configure_argument_parser(parser)
+        args = parser.parse_args(list(commandline_args))
+
+        if args.seed is not None and args.seed < 0:
+            raise ValueError(f"--seed must be non-negative, got {args.seed}")
+        if not math.isfinite(args.mouse_sensitivity) or args.mouse_sensitivity < 0:
+            raise ValueError("--mouse-sensitivity must be finite and non-negative")
+        self._validate_arguments(args)
+
+        seed = secrets.randbits(63) if args.seed is None else args.seed
```
gtong-nv: if we use random seed by default, make sure the benchmark/CI is aware of that and provides the seed arg

apps/action2v/action2v/application.py
```diff
@@ -0,0 +1,267 @@
...
+        """Return the model's native output shape and interactive rates."""
+        return SessionDesc(
+            output_layout=self.defaults.output_layout,
+            backpressure_mode=self.defaults.backpressure_mode,
+            presentation_mode=self.defaults.presentation_mode,
+            frames_per_second_for_ui=self.defaults.ui_fps or self.defaults.fps,
+            frames_per_second_for_step=self.defaults.fps,
+            video_width=self.defaults.pixel_width,
+            video_height=self.defaults.pixel_height,
+            metadata=dict(self.defaults.metadata),
```
gtong-nv, line 219: We should standardize the way the apps specify defaults.
Now some apps use dedicated file like `apps/cam2v/cam2v/defaults.py`, some just add a class in this `application.py` file, and it becomes hard to locate and modify the default parameters.
I'm thinking similar to `config.py` in each model integration, each app should have its `app_config.py`  to specify these options? 
Doesn't have to be done in this PR though

integrations_v2/waypoint/impl/model_session.py
```diff
@@ -0,0 +1,111 @@
...
+import torch
+from action2v import Action2VStep
+from torch import Tensor
+
+from flashdreams.runtime_v2.session_desc import SessionDesc
+from waypoint import WAYPOINT_1_5, WaypointControl
+from waypoint.impl.pipeline import WaypointInferencePipeline
+
+
+class WaypointModelSession:
```
gtong-nv: I think this file describes the app logic, not the model/integration impl. 

Does it belong to `waypoint/impl` ?

flashdreams/flashdreams/runtime_v2/session_runner.py
```diff
@@ -132,10 +132,19 @@ def run_ui_once(*, step_requested: bool = True) -> None:
                 if loop_result.stop_requested:
                     stop.set()
                     return
-                if loop_result.new_session_request is not None:
-                    next_session_desc = loop_result.new_session_request
-                    stop.set()
-                    return
+
+                request = ui_loop.flush_ui_loop_requests()
```
gtong-nv, line 136: Can a single `UILoopRequests` contain multiple requests? I feel this is a rare case and will be tricky to handle. Like what does it mean for one `UILoopRequests` contain both `new_session` and `hide_cursor`? Should it create a new_session with cursor hidden?

If we want to avoid this kind of complexity, maybe `UILoopRequest` should be a union with type and the payload, that just represents one request?

Also, probably should move the requests handling to a helper function

ArielG-NV: > Should it create a new_session with cursor hidden?

1. should request hidden cursor since session asked
2. recreate session ==> init of session can handle setting cursor-mode again if some default is desired.

apps/action2v/action2v/application.py
ArielG-NV: `--profile` was a mistake to include, this should be removed (along with associated junk) - this was likely legacy from old demo(?)

apps/action2v/action2v/application.py
ArielG-NV: I will ensure use matches cam2v.

apps/action2v/action2v/application.py
ArielG-NV: Agreed. 
I think this is the "next step" for demos - more standardization

remove hacks and other wacky things we've added to get demos to function

apps/action2v/action2v/application.py
ArielG-NV: I will change code to match how cam2v does things.

integrations_v2/waypoint/impl/model_session.py
ArielG-NV: Agreed, this is just the incorrect way to integrate a demo.

Only `generate` from the model config should be called in action2v, minimal integration code needed primarily for control->action mapping

apps/action2v/action2v/session.py
```diff
@@ -0,0 +1,234 @@
...
+                output=frames.detach(),
+                frame_count=frames.shape[0],
+                output_layout=state.session_desc.output_layout,
+                metrics=metrics,
+            )
+        ]
+
+    def is_finished(self) -> bool:
+        """Keep the live model loop running until the runtime stops it."""
+        return False
```
jmccaffrey-nv: [P1] Public MP4/null launches have no finite completion or deterministic controls

`Action2VModelLoop.is_finished()` always returns `False`, while the public Action2V app accepts neither a control file nor an action limit. The CLI only exposes a wall-clock timeout, not the internal `run_session(..., steps=...)` bound used by tests. As a result, normal MP4/null execution cannot finish deterministically or replay a specified control trace, and the packaged `example_controls.json` is unreachable from the production path. This misses the bounded finite-replay requirement in #479. Please add a finite control source/action limit, finish when it is exhausted, and cover the public ApplicationRunner/CLI path.

ArielG-NV: 👍

jmccaffrey-nv: Note: comment was from Sol, I reviewed it, I need to modify skills/settings so my agents always attribute .

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/579/commits/75907e9262ed297b6685ed18eacfb06cca6fa3f9

gtong-nv (approved):
LGTM

PR #583, opened 2026-09-03
[omnidreams] Preserve native cache reset interface

aidanfnv:
Fixes a regression caused by #566 that causes the existing omnidreams preset to crash.


ArielG-NV (approved):
👍

PR #586, opened 2026-09-04
Add postprocessing to cam2v lingbot

gtong-nv:
## Summary

- Add persistent, prewarmed postprocessing for the Lingbot Cam2V rollout.
- Let any configured postprocessor declare its output size; the app presents that size without mutating the model session descriptor.
- Add `--postprocess-comparison-ui` to show the raw/upscaled and postprocessed streams side by side.
- Keep postprocessing running when its UI toggle changes, and select only which output frames are presented.
- Preserve presentation backlog draining and scope the change away from Interactive Drive.

### DIFF
--- a/apps/cam2v/cam2v/application.py
+++ b/apps/cam2v/cam2v/application.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import argparse
 from collections.abc import Sequence
+from dataclasses import replace
 from pathlib import Path
 from typing import Any
 
@@ -15,6 +16,12 @@ import torch
 from flashdreams.api_v2.application import IApplication
 from flashdreams.api_v2.session import ISession
 from flashdreams.infra.config import derive_config
+from flashdreams.infra.postprocess import (
+    VideoPostprocessChainConfig,
+    VideoPostprocessStream,
+    VideoSpec,
+)
+from flashdreams.plugins.registry import discover_postprocess_presets
 from flashdreams.runtime_v2.session_desc import SessionDesc
 from flashdreams.runtime_v2.video_tensor import VideoTensorLayout
 
@@ -27,7 +34,7 @@ class Cam2VApplication(IApplication):
 
     The shared class owns command-line parsing, pipeline lifetime, session
     validation, and model-generation-loop construction. A concrete model
-    integration contributes a pipeline config and an input resolver through
+    integration contributes a runner config and an input resolver through
     :class:`Cam2VApplicationDefaults`.
     """
 
@@ -40,6 +47,9 @@ class Cam2VApplication(IApplication):
         self._use_ui = True
         self._input_values: dict[str, Any] | None = None
         self._pipeline: Any | None = None
+        self._postprocess = VideoPostprocessChainConfig()
+        self._postprocess_stream: VideoPostprocessStream | None = None
+        self._postprocess_profile = False
 
     @property
     def pipeline_config(self) -> Any:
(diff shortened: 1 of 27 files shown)


apps/cam2v/cam2v/application.py
```diff
@@ -181,10 +235,20 @@ def create_session(self, session_desc: SessionDesc) -> ISession:
...
+        model_width = int(
+            session_desc.metadata.get(
+                "cam2v_model_video_width", session_desc.video_width
+            )
+        )
+        model_height = int(
+            session_desc.metadata.get(
+                "cam2v_model_video_height", session_desc.video_height
+            )
+        )
```
gtong-nv: why is that width and height defined both in session_desc.metadata and and session_desc?

apps/cam2v/cam2v/session.py
```diff
@@ -222,17 +274,78 @@ def step(self, step_index: int, events: UserInputEvents) -> list[StepResult]:
...
+                ) = compose_postprocess_comparison(
+                    pending_generated_frames=state.comparison_pending_generated_frames,
+                    generated_frames=generated_frames,
+                    postprocessed_frames=postprocess_output_frames,
+                    final_chunk=final_chunk,
+                    output_layout=state.session_desc.output_layout,
+                )
+                _synchronize_output(output_frames)
+            elif state.postprocess_enabled:
+                output_frames = postprocess_output_frames
```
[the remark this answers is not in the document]
gtong-nv: new commit keeps SessionDesc fixed at the postprocessor’s final output size and normalize only the raw presentation branch to that size.

gtong-nv:
/ok to test [876eef1](https://github.com/NVIDIA/flashdreams/pull/586/commits/876eef1b8318c42a1335b4c0dc66fcaf404ddd4d)

gtong-nv:
/ok to test [c0e878b](https://github.com/NVIDIA/flashdreams/pull/586/commits/c0e878b280ce5c439bbdf9e06eaa823ad10d2416)

flashdreams/flashdreams/infra/postprocess/base.py
```diff
@@ -111,6 +111,17 @@ def prepare(self) -> None:
         stream when this method returns.
         """
 
+    def reset(self) -> None:
```
jmccaffrey-nv, line 114: [P1] Keep every advertised Cam2V preset reset-safe

Cam2V offers every discovered postprocess preset, and each rollout reset calls `VideoPostprocessStream.reset()`. The built-in RTX VSR session does not override this default, so selecting any `rtx-*` preset reaches a `NotImplementedError` on the first reconnect/reset (this reproduces by starting `POSTPROCESS_PRESET_RTX_SUPER_RESOLUTION` and calling `reset()`). Issue #479 explicitly requires Lingbot reconnect/reset semantics. Please either implement a resource-preserving reset for RTX sessions or restrict Cam2V choices to processors that support this lifecycle, with coverage for every advertised preset.

flashdreams/flashdreams/runtime_v2/presentation_manager.py
```diff
@@ -104,27 +109,33 @@ def observe_model_output(
...
+            self._has_observed_cadence = True
 
     def is_due(self, now: float, generation: int, *, backlog: bool = False) -> bool:
         """Return whether the next model frame may be selected."""
         with self._lock:
             if generation != self._generation:
                 self._reset_generation(generation)
-            return backlog or self._next_frame_at is None or now >= self._next_frame_at
+            return (
+                (backlog and not self._has_observed_cadence)
```
jmccaffrey-nv: [P1] Preserve backlog-driven catch-up in the shared runtime

After cadence has been observed, this condition ignores `backlog`, and `mark_advanced()` below also uses the learned cadence. A UI hiccup or producer burst can therefore leave queued chunks presenting near model rate instead of catching up at the UI cap, so added latency persists. That reverses the backlog behavior requested in #542 and changes the shared `PresentationManager` used by Interactive Drive, while #479 and this PR’s summary say that behavior is out of scope. Please retain backlog as an override here, or make the smoother postprocessed pacing a Cam2V-specific policy.

jmccaffrey-nv (changes requested):
Thanks for the thoughtful iteration here—the persistent stream, fixed presentation sizing, and comparison UI are a solid direction.

I’m requesting changes for two P1s:
- The new reset contract is not satisfied by every preset Cam2V advertises, so an RTX preset fails the reconnect/reset lifecycle required by [#479](https://github.com/NVIDIA/flashdreams/issues/479).
- Once cadence is learned, the shared presentation clock ignores backlog, undoing the latency safeguard from [#542](https://github.com/NVIDIA/flashdreams/issues/542) and changing Interactive Drive behavior despite that being out of scope.

The latest commit fixed the earlier pacing-test mismatch; the focused CPU suite now passes: **137 passed**.

Two P2 follow-ups:
- For a stream beginning with exactly 8/16 frames, pass the real first chunk as AR0 (the encoder accepts it) instead of priming temporal/KV state with a discarded repeated-frame chunk; please add a non-constant-input test.
- `--postprocess-compile` defaults to false even though the preset default and Lingbot docs imply the compiled/prewarmed path is the default. Please preserve the preset unless the flag is explicitly supplied (for example, `default=None`), or align the docs and help text.

I did not duplicate the existing inline comment about raw-frame presentation sizing.

— assisted by GPT 5.6 Sol

flashdreams/flashdreams/infra/postprocess/base.py
gtong-nv: Implemented `reset()` in `_RTXVideoSuperResolutionPostProcessorSession`

flashdreams/flashdreams/runtime_v2/presentation_manager.py
gtong-nv: This is a good catch! `_has_observed_cadence` is redundant and has been removed in the new commit

gtong-nv:
/ok to test [430caef](https://github.com/NVIDIA/flashdreams/pull/586/commits/430caefd65bbf8d7155b060b3b4be024a44fd583)

gtong-nv:
It measures the model.step(), but this is still not the end to end FPS that the user will perceive.
For example, for CUDA tensors in the stepResult, it generally does not measure actual GPU completion or D2H copy:

Maybe we name it `model_step_FPS`

jmccaffrey-nv (approved):

Issue #564, opened 2026-09-02
Cam2V reports step timing under non-canonical metric keys, so LingBot scenarios get no `total_s` or `generated_fps`


jarcherNV:
# Cam2V reports step timing under non-canonical metric keys, so LingBot scenarios get no `total_s` or `generated_fps`

## Summary

The Cam2V application reports its per-step performance as `model_step_wall_s`
and `chunk_fps`. The benchmark tooling's canonical names for those two
quantities are `total_s` and `generated_fps`. Because Cam2V reports neither
canonical key, the LingBot Cam2V scenarios drop out of every part of the
benchmark pipeline that is keyed on them: no `generated_fps` is derived, no
run highlights are recorded, and the generated HTML report omits its two
headline performance cards.

The measurement itself is correct — this is a naming and derivation problem,
not a real performance regression. But it makes LingBot's numbers invisible to
the standard tooling, and anything that gates on benchmark performance has to
carry a second set of metric names to cope with it.

## Details

`apps/cam2v/session.py` adds only these two keys on top of whatever the
pipeline's `finalize()` returns:

```python
metrics.update(
    {
        "model_step_wall_s": model_step_wall_s,
        "chunk_fps": frame_count / model_step_wall_s,
    }
)
```

`integrations_v2/cam2v_lingbot/README.md` documents `model_step_wall_s` as
covering input preparation, generation, finalization, and decode — i.e. exactly
the quantity `total_s` names elsewhere.

Downstream of that, in `flashdreams/tools/benchmarks/`:

- **`metrics.py`** — `_generated_fps_for_step()` derives `generated_fps` from
  `total_s`, falling back to `model_step_s`. Cam2V reports neither, so
  `generated_fps` is never derived for LingBot steps, and
  `_generated_fps_summary()` likewise produces nothing.
- **`harness.py`** — run highlights take `startup_step_total_s` from the first
  step's `total_s`, and medians for `total_s` and `generated_fps`. All three
  are absent for LingBot runs.
- **`report.py`** — the per-scenario headline cards are emitted only when
  `total_s_median` / `generated_fps_median` are present, so the LingBot
  scenarios render with no step time and no throughput up top. Neither
  `model_step_wall_s` nor `chunk_fps` has an entry in `_LABEL_OVERRIDES` or
  `_CHART_ORDER`, so they also appear with raw key names rather than proper
  labels and chart placement.

The affected scenarios in `configs/v2_model_benchmarks.json` are
[...]

PR #589, opened 2026-09-04
Backfill total_s from Cam2V's model_step_wall_s for LingBot benchmark scenarios

AmirF194:
Refs #564

Cam2V's session.py reports its per-step duration and rate as `model_step_wall_s`/`chunk_fps`, and `_generated_fps_for_step`/`_generated_fps_summary` only ever look at `total_s` (falling back to `model_step_s`), so LingBot's Cam2V scenarios never get a derived `generated_fps`, and `harness.py`'s `total_s` highlights come up empty too.

I went with the "report `total_s` alongside the existing key" option rather than renaming: the runtime-metric-samples ingest path had no key-normalization step at all (unlike the stats-rows path, which already has `_KEY_OVERRIDES` for this), so I added a small alias step there that backfills `total_s` from `model_step_wall_s` when a step doesn't already report `total_s`. `model_step_wall_s` stays in the output. Once `total_s` is present, the existing fallback in `_generated_fps_for_step`/`_generated_fps_summary` derives `generated_fps` on its own, and `harness.py`'s `total_s_median`/`startup_step_total_s` highlights and `report.py`'s headline cards pick it up with no changes needed there.

This covers the LingBot scenarios dropping `generated_fps` and the headline cards, not the regression-gate part of the issue (the metric name lists that skip instead of failing). I couldn't find that gate's config in this repo, so if it lives in an internal tool this half may need a separate change there.

Added `test_runtime_benchmark_stats_backfills_total_s_from_model_step_wall_s`: feeds a `model_step_wall_s`/`chunk_fps` sample through `records_from_stats_file` and checks both `total_s` and `generated_fps` come out. The new test fails without the change. The existing runtime-metric-samples tests (group-by-step, warmup exclusion, the v2 sink one) still pass unchanged.

Couldn't run the full suite locally, this workspace needs CUDA/transformer-engine to build. Checked the change with ruff (check and format, both clean) and by loading `tools/benchmarks/metrics.py` directly with no other project dependencies, since the module itself only imports the standard library.


ZenAlexa (changes requested):
The alias is applied inside `records_from_stats_file()`, after the runtime stats artifact has already been written. Cam2V's `StepResult.metrics` therefore continues to contain `model_step_wall_s`/`chunk_fps` and omits `total_s`, so the raw artifact and every direct consumer keep the non-canonical contract described in #564; the first acceptance criterion remains open. I added `assert result.metrics["total_s"] == result.metrics["model_step_wall_s"]` to `test_model_loop_maps_wasd_to_shared_camera_input_and_metrics`, and the PR head fails with `KeyError: 'total_s'`. Assigning `total_s` beside `model_step_wall_s` in `Cam2VModelLoop.step()` gives every consumer the canonical key. This parser alias can also remain for historical artifacts.

AmirF194:
You're right, the alias only patched the ingest path, so StepResult.metrics itself never got total_s. Pushed 7f1270a8: Cam2VModelLoop.step() now sets total_s alongside model_step_wall_s at the source, and the parser alias stays for historical artifacts.

Added your assertion to the existing WASD test: fails with KeyError: 'total_s' on the prior commit, passes now. Ran it CPU-only (torch cpu wheel, no GPU here) along with the rest of test_application.py and test_benchmark_harness.py, 59 passed, 1 skipped, ruff clean on both files.

ZenAlexa (approved):
Cam2V now emits `total_s` in the producer, and the parser preserves an explicit canonical value. The runtime-metrics and runtime-records tests pass at `7f1270a8` (59 passed, 1 skipped); restoring the previous producer reproduces `KeyError: 'total_s'` in the regression. The raw artifact, throughput summary, and report paths now share the metric.

AmirF194:
Rebased onto main. NVIDIA/flashdreams#607 (merged after this PR was opened) rewrote `Cam2VModelLoop.step()`'s metrics flow: `StepResult` now returns the raw `state.pipeline.finalize()` dict directly instead of the locally enriched `metrics` dict, so `model_step_wall_s`, `chunk_fps` and every other value this step used to add are no longer present in `StepResult.metrics` at all. The `total_s` line I pushed on top of the old `metrics.update()` call no longer reaches anything, so I dropped it and kept this PR scoped to the `metrics.py` key-normalization fix, which is independent of that flow and still verified working against current main.

That leaves the deeper gap open: Cam2V's runtime metric samples carry none of the per-step timing fields post-#607, not just `total_s`. Whether `StepResult` should go back to returning the enriched dict, or something else given #607's own note that it is temporary pending #603, is a call for someone closer to that migration. Happy to open a separate issue for it if that's useful.

AmirF194:
This rebased cleanly onto the new metrics flow five days ago and hasn't had a maintainer look yet, flagging it in case it slipped by.

jmccaffrey-nv:
Thanks @AmirF194 , we'll check the PR .

jmccaffrey-nv:
Thanks for rebasing this onto the new metrics flow. Could we add an end-to-end regression test that follows the actual post-#607 path? StepResult.metrics → MetricsOutputSink → records_from_stats_file(), and verifies that total_s and generated_fps appear correct.

The current test injects model_step_wall_s directly, but Cam2V no longer writes that metric to StepResult. Testing the real producer path will confirm whether this change is still needed for current LingBot scenarios and guard the behavior we ship.

AmirF194:
Added it: pushed 80730eca with a test that writes a StepResult carrying model_step_wall_s and chunk_fps through MetricsOutputSink, closes it, then reads it back via records_from_stats_file and checks total_s and generated_fps land. Disabling the alias backfill makes it fail with KeyError on total_s, so it does discriminate.

The existing test that uses model_step_wall_s builds the stats JSON by hand and skips the sink; the one that goes through the sink uses total_ms, which is already canonical. Neither covered both at once.

PR #604, opened 2026-09-09
[API] Report FPS in a standardized way

### LINKED ISSUE #603: [API] Standardize reporting of metrics 'better'
FPS has many different meanings for `flashdreams`:
- end-to-end fps - Users care about this value since this is the "fluidity of an experience". This includes ui_step, model_step, and presentation overhead
- model_step fps/throughput - Developers care about this value since this is the "upper-bound" in perf. If your model_step is too slow, your ui_step will be bottlenecked by it. This technically is also useful for model developers since a multi-step system can then easily be benchmarked on only the cost of the "model".
- ui_step fps/throughput - Developers care about this value since if you are not hitting ui_step FPS consistently, something may be going very wrong. This number is mostly useless though.
- how many pipeline.generate() we can run in a second - Model developers care about this value for performance/benchmarking reasons

Notes: we need to handle multiple `StepResult` as output as well

ArielG-NV:
resolves: #603

Extend `MetricsOutputSink` to report model step 'FPS' in a standardized way when trying to emit stats via `--stats-path`.

We will generally try to present 'FPS of something' via model_step FPS since ui_step FPS is quite easy to raise (cheap to run ui_step, expensive to run model_step).


apps/t2v/t2v/testing.py
```diff
@@ -337,7 +338,7 @@ def __init__(self, spatial_compression_ratio: int) -> None:
         self.spatial_compression_ratio = spatial_compression_ratio
 
 
-class _FrameInspector(OutputSink):
+class _FrameInspector(MetricsOutputSink):
```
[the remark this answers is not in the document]
ArielG-NV: This debug object can implement its own hacks, at the end of the day, it still conforms to our interface.

flashdreams/flashdreams/runtime_v2/metrics_output_sink.py
```diff
@@ -38,15 +40,30 @@ def __init__(self, path: str | Path) -> None:
...
+        self, step_elapsed_s: float, step_results: list[StepResult]
+    ) -> None:
+        """Record wall time and the largest channel frame count for one step.
+
+        Visual update cadence follows the largest channel in the step, so FPS
+        is that frame count divided by elapsed seconds.
+        """
+        if not step_results:
+            return
+        largest_chunk_size = max(result.frame_count for result in step_results)
```
gtong-nv: do we have a case where the `frame_count` in the `step_results` are different?
Trying to understand why is this needed

ArielG-NV: It is not needed, I realized we have this:
```
        if frame_count <= 0 or any(item.frame_count != frame_count for item in chunk):
            raise ValueError("Every channel in a chunk must have the same frame_count.")
            ```

File: flashdreams/flashdreams/runtime_v2/session_runner.py [hunk 3]
```diff
@@ -168,6 +169,7 @@ def publish_model_results(
                 step_elapsed_s=step_elapsed_s,
             )
             if metrics_output_sink is not None:
+                metrics_output_sink.new_model_step(step_elapsed_s, results)
```
gtong-nv: the `new_model_step` function is essentially just recording the results with `step_elapsed_s`, right? The name seems confusing. Maybe name it `record_model_step_elapsed_s()` ?

ArielG-NV:
Current design does not account for `StepResult::_output_ready_event`, I need to amend for this

flashdreams/flashdreams/runtime_v2/session_runner.py
ArielG-NV: renamed 👍

ArielG-NV:
added 
```
        for result in step_results:
            result.wait_until_ready()
            ```
            
         This is to ensure we measure model_step fully+correctly.

**Comment by ArielG-NV** [comment]
/ok to test https://github.com/NVIDIA/flashdreams/pull/604/commits/7eac48888bf35cfebc5f7bb8b334215b4d5e4897

File: flashdreams/flashdreams/runtime_v2/metrics_output_sink.py [hunk 4]
```diff
@@ -38,15 +40,34 @@ def __init__(self, path: str | Path) -> None:
...
+        Visual update cadence follows the largest channel in the step, so FPS
+        is that frame count divided by elapsed seconds.
+        """
+        assert step_results and len(step_results) > 0
+
+        # Ensure all results are ready before we measure the chunk size.
+        # Since we are measuring "elapsed ms average", the cost of presenting
+        # all 8 at once is being measured via this wait.
+        for result in step_results:
+            result.wait_until_ready()
```
gtong-nv, line 68: I understand we need to measure the model step(), but we need to be very careful and avoid this sync point in production (when not measuring the model.step()). This might hurt end to end performance.

gtong-nv (approved):

flashdreams/flashdreams/runtime_v2/metrics_output_sink.py
ArielG-NV: I suppose a more elegant solution exists here, and I should try to implement it.

ArielG-NV:
we decided to take a different approach, and side-line this idea

ArielG-NV:
For a clearer justification:

FPS has many different meanings for `flashdreams`:
- end-to-end fps - Users care about this value since this is the "fluidity of an experience". This includes ui_step, model_step, and presentation overhead
- model_step fps/throughput - Developers care about this value since this is the "upper-bound" in perf. If your model_step is too slow, your ui_step will be bottlenecked by it. This technically is also useful for model developers since a multi-step system can then easily be benchmarked on only the cost of the "model".
- ui_step fps/throughput - Developers care about this value since if you are not hitting ui_step FPS consistently, something may be going very wrong. This number is mostly useless though.

For now we will just stick with `.generate` metrics since we can refine that system further and have not decided how to move forward with tracking the metrics described above.

PR #613, opened 2026-09-11
[Crazy RoboTaxi] Improve Crazy Robotaxi controller handling

aidanfnv:
Split off from #553, this change modifies the controller handling to make the game feel better on a controller, such as by restoring analog steering. Previously, to make steering more responsive, it was apparently changed to be digital.


spaingnv (approved):
LGTM. New tests cover command shape, dead zones, and keyboard vs pad ownership.

PR #617, opened 2026-09-11
Add multi-view inference support

aidanfnv:
## Summary

This PR adds primitives for multi-view autoregressive inference and extends the FlashDreams attention interfaces to support masked grouped-query attention.

- Add multi-view geometry, packing, prefill, visibility-mask, mRoPE, cache, and rollout APIs under `flashdreams.core.attention`.
- Support dense boolean and FlexAttention block masks across the functional, reference, PyTorch, and optimized attention paths.
- Expose one `FlexAttentionOptions` policy across functional and module APIs, including asymmetric Q/KV mask block sizes and kernel tuning options.
- Add native grouped-query attention handling and explicit errors for backend combinations that do not support masks or GQA.
- Add call-level understanding/generation routing through `MoTMultiHeadAttention`.
- Update OmniDreams cross-view attention to use the new pack/unpack API.

## Why

Multi-view inference needs explicit definitions for view-major token layout, cross-view visibility, temporal positions, bounded streaming memory, and deterministic chunk scheduling. The new core APIs define those operations independently of model execution.

The attention changes expose the matching lower-level contract: callers can provide either a dense mask or a block-sparse mask and can retain grouped K/V heads instead of expanding them before attention.

### Multi-view core

- `ClipGeometry`, `ChunkPlan`, `MemoryLayout`, and `ChunkMetadata` describe clip geometry and each autoregressive pass.
- `pack_cross_view_attention` and `unpack_cross_view_attention` define the cross-view layout.
- A single visibility specification builds equivalent dense and block-sparse masks.
- `PrefillPack` defines clean text/control/target prefill metadata, retained-memory rows, and fixed-slot control alignment.
- `FixedSlotKVCache` and `TokenWindow` provide bounded circular storage for K/V memory and generated tokens.
- `ChunkRollout` and `run_rollout` implement deterministic fixed-step generation through small model and state protocols.
- Position IDs are allocated on the requested model device, partial cache writes clear unused slot tails, and cache writes/resets validate all layers before mutation.
- Cache reset clears the complete non-prefix suffix, and mRoPE/rollout entry points reject invalid frame timing before model preparation.

### Attention

- `AttentionMask` accepts a two-dimensional boolean tensor or FlexAttention `BlockMask`.
- `masked_attention` validates shapes and selects dense SDPA or compiled FlexAttention from the mask type.
- `FlexAttentionOptions` controls mask block geometry, dynamic compilation, and forwarded FlexAttention kernel options for both `masked_attention` and `OptimizedMultiHeadAttention`.
- `TorchMultiHeadAttention` and `OptimizedMultiHeadAttention` accept masks and infer when native GQA is required.
- Unsupported FP8, FA2, mask, and head-layout combinations fail explicitly at the backend boundary.

Existing callers can continue omitting `attn_mask`; their unmasked behavior is unchanged.

## Validation

- The repository-wide pre-commit suite passed, including Ruff lint/format, lockfile and version checks, and `ty`.
- Complete multi-view, attention, and OmniDreams test surface: 203 passed; 427 unavailable GPU cases skipped.
- Coverage includes dense/Flex mask parity, GQA, invalid geometry, slot rotation, partial chunks, deterministic rollout, pack/unpack round trips, and OmniDreams cross-view output/cache geometry.


jmccaffrey-nv:
Hey Aidan, I reviewed #617 at `dd9e42b`. The earlier Greptile findings are fixed, and the overall design looks thoughtful.

One correctness issue :

- `FixedSlotKVCache.reset()` resets cursors and visible length but leaves suffix storage intact. `write(A) → write(B) → reset() → write(B)` exposes A’s stale K/V again. Please clear or otherwise invalidate the suffix during reset and add a regression test.

Two smaller hardening fixes:

- `write()` and `reset()` validate while mutating, so a malformed later layer leaves earlier layers modified. Please validate all layers before copying.
- `_mrope_ids()` should reject non-finite FPS values. Require `math.isfinite(fps) and fps > 0`, preferably before `prepare_rollout()` and at the helper boundary.

-- assisted by GPT 5.6 Sol

jarcherNV:
Thanks for doing this. Overall it LGTM, but I did notice one potential issue.

FlexAttention is implemented, but masked_attention() only exposes dynamic, does not forward kernel_options, and only supports square mask block sizes, and I think we will need that. Could you expose a shared FlexAttention configuration (including asymmetric (Q, KV) block sizes) across the functional and module APIs? Otherwise, performance-sensitive multi-view consumers will have to maintain their own attention wrapper instead of using the shared inference stack.

jmccaffrey-nv (approved):

PR #619, opened 2026-09-11
Add tensor and context parallel inference to the core runtime

jmccaffrey-nv:
## Summary

- Add reusable tensor- and context-parallel inference primitives, including autograd-aware sequence collect and parallel linear, attention, and normalization modules.
- Add device-mesh construction and runtime state with independently selectable backends for model-parallel axes.
- Coordinate multi-GPU API v2 sessions at model-step boundaries while keeping presentation and metrics ownership on rank zero.
- Document the runtime architecture and add CPU fault-injection coverage for distributed primitives, device ownership, and session lifecycle behavior.

## Review follow-up

- Clean up process groups created by `init_parallel()`: successful v2 CLI runs use coordinated shutdown, failures avoid peer waits, and the owned-world exit fallback skips potentially blocking destruction after an uncaught rank failure.
- Preserve the caller-selected CUDA device when `LOCAL_RANK` is unavailable.
- Keep replacement requests alive across session cleanup and application deadlines.
- Keep an unfinished UI open after inference while workers poll rank zero once per UI tick.
- Bypass session-result collectives on failures so the original exception reaches the process supervisor promptly.
- Make rank zero's terminal or replacement decision authoritative across all runtime ranks.

## Rebase update (2026-09-17)

- Rebased the full 13-commit series onto GitHub `main` at `2dbfdec0`.
- Preserved cumulative unpublished-step timing from current `main` while retaining empty-result worker semantics.
- Apply synchronized rank-zero inputs before reset-generation timing bookkeeping.
- Published rewritten head `cafd8d98` with an exact force-with-lease.

## Validation

- Post-rebase focused CPU suite: **168 passed** (`test_distributed_parallel`, `test_step_agreement`, `test_application_runner`, and `test_session_runner`).
- Post-rebase Ruff format and lint: **17 changed Python files passed**.
- Post-rebase `git diff --check`: passed.
- GitHub checks on `cafd8d98`: **CPU, GPU, MP4 demo, WorldLens canary, docs, REUSE, OSRB, and Greptile all passed**.

## Release review

Reviewed for public release. No secrets, internal infrastructure references, or non-public benchmark results were found in the final GitHub diff.


jarcherNV:
Thanks for doing this, the overall approach makes sense to me. Just a few comments (mostly from me asking codex about this PR):

P1 - Could you please make sure the process group created by init_parallel() is cleaned up? It looks like this path bypasses the existing atexit registration, and the v2 CLI never calls the distributed shutdown helper. Perhaps it could follow the shutdown pattern used by flashdreams-run.

P2 - It looks like reusing an existing process group without LOCAL_RANK resets every rank to cuda:0, even if the caller already selected a device. Do we need to preserve the current CUDA device or accept an explicit device?

P2 - It looks like request_new_session() is discarded during distributed execution, causing the application to exit instead of starting the replacement session. Is that intended behavior?

P2 - It looks like distributed sessions skip the normal unfinished-UI check once inference completes. Should rank zero retain the existing post-inference UI lifecycle while the worker ranks wait for a coordinated exit?

Up to you if you think the P2s are worth addressing.

flashdreams/flashdreams/api_v2/loop.py
```diff
@@ -257,6 +259,13 @@ def _run_message_batch(self) -> None:
             if result is not None:
                 raise TypeError("Message operations must return None.")
 
+    def _pace(self, last_run_started: float | None) -> float:
+        if self.frequency == 0 or last_run_started is None:
+            return time.monotonic()
+        earliest_start = last_run_started + 1.0 / self.frequency
+        self._shutdown_event.wait(max(0.0, earliest_start - time.monotonic()))
+        return time.monotonic()
+
```
ArielG-NV: can we name the method something more correct like `pace_ui_thread_to_target_frequency` or something like that

flashdreams/flashdreams/api_v2/loop.py
```diff
@@ -271,6 +280,12 @@ class IModelLoop(ILoop[StateT], ABC):
     :meth:`ILoop.step` must return ``list[StepResult]`` here, one entry per
     channel, with every channel reporting the same ``frame_count``. Returning a
     bare :class:`StepResult` or ``None`` raises :class:`TypeError`.
+
+    **An empty list is a step that presented nothing**, which is how a loop says
```
ArielG-NV: a model step may also emit nothing if it has... no model generating frames at a different cadence than UI

flashdreams/flashdreams/api_v2/loop.py
```diff
@@ -313,29 +330,51 @@ def _run_model_loop(
...
+                torch.cuda.set_device(device)
+            while True:
+                stopping = self._shutdown_event.is_set() or (
+                    max_steps is not None and steps_run >= max_steps
+                )
+                if agreement is not None:
+                    if not agreement.ready(stopping=stopping):
+                        break
+                elif stopping:
+                    break
```
ArielG-NV: two different ways to stop? should we potentially canonicalize our ways to stop~?

flashdreams/flashdreams/api_v2/loop.py
```diff
@@ -313,29 +330,51 @@ def _run_model_loop(
...
+                            except BaseException:
+                                # A departed peer must not replace the original failure.
+                                pass
+                        raise
+                    stopping = run.step_index is None or self._shutdown_event.is_set()
+                    if agreement is not None:
+                        if not agreement.ready(stopping=stopping):
+                            break
+                    elif stopping:
                         break
```
ArielG-NV: we have this exact same code earlier too

jmccaffrey-nv:
Thanks - I agreed these were worth addressing and pushed the fixes through `d4446e41`.

- Process-group cleanup: `init_parallel()` registers cleanup only when it creates the default group. The v2 CLI now follows the existing runner's bounded shutdown policy: synchronized cleanup after success and no peer wait/destruction after a rank-local failure. The exit fallback is also failure-aware for programmatic callers, so an uncaught exception cannot block the failed rank in `destroy_process_group()`.
- Existing CUDA worlds: when `LOCAL_RANK` is absent, `init_parallel()` preserves the caller's current CUDA device.
- Replacement sessions: rank zero's replacement request and deadline decision are coordinated after cleanup and returned on every rank.
- Post-inference UI: rank zero retains the normal unfinished-UI lifecycle; workers participate in a lightweight coordinated decision once per UI tick.

I also incorporated the follow-on distributed edge cases Greptile found (idle UI timeout, failure-path collectives, and deadline divergence). Final-head CPU, GPU, MP4, WorldLens, docs, REUSE, and OSRB checks all pass. Greptile is now 5/5 with no outstanding/actionable findings.

flashdreams/flashdreams/core/distributed/parallel.py
```diff
@@ -0,0 +1,208 @@
...
+        local_rank = torch.cuda.current_device()
+    device = torch.device(f"cuda:{local_rank}" if cuda_available else "cpu")
+    tp_size, cp_size = plan_mesh(world_size, tensor_parallel, head_groups=head_groups)
+    if device.type == "cuda":
+        torch.cuda.set_device(device)
+    if world_size == 1:
+        return ParallelContext.single(device)
+    if not dist.is_initialized():
+        dist.init_process_group(backend="nccl" if device.type == "cuda" else "gloo")
+        atexit.register(_safe_destroy_pg)
```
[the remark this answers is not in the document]
jmccaffrey-nv: Addressed in `d4446e41`. The shared `_safe_destroy_pg()` callback now detects the interpreter's uncaught-exception state (`sys.last_exc`, with the Python 3.10 `sys.last_value` fallback) and returns without calling `destroy_process_group()` on a failed exit. Clean exits still destroy an owned group. Focused tests cover both exception-state variants and the clean-exit path.

PR #621, opened 2026-09-11
[crazy-robotaxi] Finish UI loop on exit

aidanfnv:
Fixes a regression that occurred in the demo after I rebased it onto 41846b7b7a770ab833813cf1caa520c1d26aef32. With that change, the lifecycle change now deliberately keeps an unfinished UI alive after its model loop finishes, while Crazy Robotaxi’s Exit button only marks the model loop finished.

This changes the demo to properly finish the UI loop on exit instead of ending the model loop and assuming that will also close the UI and the application, now that the exit behavior has changed in the API.


ArielG-NV (approved):
LGTM

PR #623, opened 2026-09-11
Stop decoding video frames past the requested window

### LINKED ISSUE #622: [Perf] -- Video loading decodes the whole file even when only a window of frames is needed
## Problem

`load_video_tensor` and `read_first_frame_rgb` in `flashdreams/infra/runner_io.py`
decode every frame of a video, then slice out the frames the caller wants.
When only a short window of a long clip is needed (the first frame, or a fixed
span from some offset), this wastes time and host memory on frames that are
thrown away.

## Proposal

Read frames with `mediapy.VideoReader` and stop once the requested window is
collected:

- New helper `read_video_window_rgb(path, start_frame, frame_count)`.
- `read_first_frame_rgb` uses it with `frame_count=1`.
- `load_video_tensor` gets optional `start_frame` / `frame_count`. Defaults
  keep reading the whole file, so existing callers are unchanged.
- A file shorter than the requested window raises a clear `ValueError`.

Output is byte-identical to the current code. Only load time changes.

jmataiNV:
Closes #622 -- 

This PR does: 
`load_video_tensor` and `read_first_frame_rgb` decoded every frame of a video
with `mediapy.read_video` and then sliced out the frames the caller wanted.
This PR reads through `mediapy.VideoReader` and stops as soon as the requested
window is collected.

- New helper `read_video_window_rgb(path, start_frame, frame_count)`.
- `read_first_frame_rgb` uses it with `frame_count=1`.
- `load_video_tensor` gets optional `start_frame` / `frame_count`. Defaults
  still read the whole file, so existing callers are unchanged.
- A file shorter than the requested window raises a clear `ValueError`.

Output for the same window is byte-identical to the previous slicing. Only
load time changes; generation is untouched.

# How we verified?  
Verified locally: `pre-commit run -a` clean, `pytest -m ci_cpu` 1658 passed,
and a round-trip check of the window reader against the whole-file reader on
a real mp4.


gtong-nv (approved):
LGTM

PR #624, opened 2026-09-12
[Metrics] Emit `SwiftVR` metrics

ArielG-NV:
Emit `SwiftVR` metrics. Previously they were not emitted

### DIFF
--- a/.github/workflows/lingbot-demo-runtime.yml
+++ b/.github/workflows/lingbot-demo-runtime.yml
@@ -148,6 +148,7 @@ jobs:
             frun "${APP_ID}" \
               --mode mp4 \
               --output-path "${output_dir}/lingbot-demo-replay.mp4" \
+              --stats-path "${output_dir}/lingbot-demo-replay-stats.json" \
               --pixel-height "${HEIGHT}" \
               --pixel-width "${WIDTH}" \
               --fps "${FPS}" \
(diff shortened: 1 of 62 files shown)


ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/624/commits/85e7b373e866c13804760ac70e30175ef03d1def

gtong-nv:
sorry for missing this timing tracking. I looked at what's going on with the base pipeline which has the eventProfile built in to the `generate(), and realized -

There are already two protocols mixed into one class:

1. Runner lifecycle (what every inheritor actually shares): initialize_cache → generate(ar, cache, input) → finalize(ar, cache), plus optional flush.
2. Default AR-diffusion body (what the base implements): one encoder tensor → one DiffusionModel.generate (scheduler sample + FinalState) → one decode.

Wan, Cosmos, Lingbot, Omnidreams, and Waypoint stay in (2): they preprocess input, then call super().generate(). FlashVSR and SwiftVR only share (1). Their middle is not one public AR step = one scheduler denoise.

So `FlashVSR` and `SwiftVR` are unique such that they can't call `super.generate()` and have to do the timing tracking themselves.
I guess we have to do that for now until we have the proper `composable pipeline` where the `generate()` is not fixed 3 stages steps.

ArielG-NV:
> I guess we have to do that for now until we have the proper composable pipeline where the generate() is not fixed 3 stages steps.

Agreed. Added `format_result_as_ms` to make formatting result consistent (code-wise) until we manage to make this new system (rather than adhoc formatting of results).

ArielG-NV:
/ok to test https://github.com/NVIDIA/flashdreams/pull/624/commits/0e42f1b26252fe0c6cc70445b6df8477e0aa325a

gtong-nv (approved):
LGTM

Issue #470, opened 2026-08-17
[API] Add to Backend of API NVTX Annotations For Ease Of Profiling


ArielG-NV:
[API] Add to backend of API NVTX profiling annotations to enable tracking utilization across GPU<->CPU context and if things overlap correctly.

We require *CUDA* tracing on top of CPU+Vulkan, so we need Nsight integration and cannot use other tools like Tracy.

ArielG-NV:
Needs:
1. annotate functions accordingly.
2. add command to launch demo with nsys to collect metrics. Must function on linux, preferably functions on windows.
3. provide guide on how to use the profile command (arguments?) and how to run nsys with the generated results.
4. agent file should note that performance optimization attempts should refer to the profiling guide for collecting an nsys report.

Note: default output location of nsys reports should be somewhere under `<repo>/artifacts` folder.

psamin:
Hi! I’m Praneeth, a CS undergrad at Georgia Tech. I’ve built a prototype for this issue and wanted to check two things before opening a PR.

It includes:

- NVTX annotations covering the model loop, pipeline stages, denoising steps in all three schedulers, and UI activity. These also don’t add device synchronization.
- A `flashdreams-profile` command saving Nsight Systems reports to `<repo>/artifacts/profiles`, with `--duration` for interactive runs and `cuda,nvtx,osrt,vulkan` tracing enabled by default.
- A profiling guide linked from `AGENTS.md`.

I tested on Georgia Tech’s Linux cluster using an L40 and `self-forcing-wan2.1-t2v-1.3b`, with `torch.compile` disabled. The trace showed the expected 7 model steps and 28 denoising iterations. I’m in the process of getting access to a Hopper GPU to finish testing with `torch.compile` and CUDA graphs, and to measure annotation overhead. I also had some questions:

1. Is enabling NVTX through an environment variable ok? (`FLASHDREAMS_NVTX=1`)
2. I don’t have a Windows machine, so is Linux testing enough for the initial PR?

One more thing to add on: FlashVSR is covered at the model-loop level, but I can add annotations for its individual stages in a follow-up PR.

I'm also happy to split the annotations and profiling tools into smaller PRs if that’s easier to review.

ArielG-NV:
I think this task is closely linked to #603 design wise. Flashdreams is in need of a cleanup/modernization of its sampling/profiling logic as a whole.

Any effort here to improve the system is very much appreciated!

[...]

PR #634, opened 2026-09-18
[API] Add NVTX ranges and a flashdreams-profile command

psamin:
Adds NVTX ranges and a `flashdreams-profile` command to show model, UI, and GPU activity together
in Nsight Systems.

Contributes to #603, following the closure of #470. This covers NVTX range support; the
`IProfiler` interface remains a follow-up.

- Adds ranges around the model loop, pipeline stages, denoising iterations in all three schedulers,
  and the UI loop.
- Enables ranges with `FLASHDREAMS_NVTX=1`, with a CPU-safe no-op when disabled or CUDA is
  unavailable. Ranges add no CUDA synchronization.
- Adds `flashdreams-profile`, which enables the ranges, traces `cuda,nvtx,osrt,vulkan`, and saves
  reports under `artifacts/profiles/`, with a `--report-path` override.
- Adds a profiling guide linked from the developer docs and `AGENTS.md`.

Adds 9 CPU tests and verified on an L40 and an H200 with self_forcing (flow-match), wan21 (UniPC) and HY-WorldPlay (Euler).

Limitations in this first PR: FlashVSR and SwiftVR get model and UI ranges but no pipeline ranges; Waypoint has no denoising ranges. Windows is untested.

Two questions for review:

1. I kept `FLASHDREAMS_NVTX` separate from `FLASHDREAMS_SYNC_AND_PROFILE` because the latter builds
   an `EventProfiler` that synchronises twice per step, which changes the CPU/GPU overlap a trace is
   meant to show. Would you like it to enable ranges as well, so `--stats-path` runs carry labels?
2. Should these call sites be integrated into `IProfiler` for #603, or do you want the interface finished first?


jmccaffrey-nv:
Thanks for contribution @psamin !
Can you incorporate the two P2 bot comments please ?

ArielG-NV:
I wonder if we should just go directly to the `IProfiler` API since it may add design constraints that require some significant rework of how we end up measuring things here (additionally, `IProfiler` should remove all direct use of nvtx from the code since all nvtx would be issolated in `NVTXProfiler(IProfiler)`)

psamin:
Thanks @jmccaffrey-nv! Both issues got fixed in the latest commit.

@ArielG-NV that makes sense. I can switch to `IProfiler` in this PR or leave it for the next one, whichever you prefer. NVTX calls are already within `infra/nvtx.py`, so most of this should carry over. I can also work on FPS/metrics collection, dependencies, and Tracy support in follow-up PRs.

gtong-nv:
I would suggest adding the `IProfiler` interface to avoid duplicated `EventProfiler` and the new NVTX addition, then the runtime and pipeline would depend only on `IProfiler`, injected at the session/runner boundary:

```
with profiler.range("pipeline.diffuse"):
...
```

I don't think it's necessary to go deeper to the concrete schedulers to add the NVTX instrumentation for now. We should keep the tracing logic in shared path so new schedulers don't need to worry about the profiling logic.

psamin:
Thanks @gtong-nv, that makes sense, and I'll go ahead and get started. Just to make sure I'm understanding things right, I'll ensure the session uses one profiler, created at the session/runner boundary and passed to loops through register_session_loop_objects(), and then handed to its pipeline. I'll push it here after I test it on a GPU.

I won't touch FlashVSR/SwiftVR’s profiling for now, since they override generate() and don't use the shared path.

jmccaffrey-nv:
Add Ariel to shepherd in review, also aligning to an IProfiler direction (for NVTX with nsys , or alternative instrumented profilers)

ArielG-NV:
>  I'll ensure the session uses one profiler, created at the session/runner boundary and passed to loops through register_session_loop_objects(), and then handed to its pipeline. I'll push it here after I test it on a GPU.

Yes something like that would work:
IApplication -> ISession -> (`register_session_loop_objects`) -> ILoop.

We likely also need a ContextVar with a default IProfiler that governs our inferencing API logic, and once application_runner starts we override it via some sort of `with set_flashdreams_inference_profiler(profiler)`

PR #636, opened 2026-09-18
[Crazy RoboTaxi] Fit Crazy Robotaxi on a 32 GB GeForce RTX 5090

mccaffrey-jonathan:
Stacked on #616 (base: `dev/aidanf/game/crazy-robotaxi/debug-views`); the diff is the five commits after Aidan's stack.

## Summary

- **`--ui/--no-ui`** for Crazy Robotaxi: a HUD-free UI loop that presents the raw video channel and keeps the HUD state, so `--mode mp4` runs need no Vulkan device (e.g. WSL2). Requires `--game-mode`, `--map`, and `--total-blocks`. `frame_to_layout` becomes public.
- **`run_on_cpu` / `embedding_cache_size`** on `CosmosReason1TextEncoderConfig`. Game hosts keep the pipeline resident and cannot release the encoder; its ~15 GiB is most of the 19.4 GiB load footprint. Running it on the host (about 2 s per 512-token prompt) and caching prompt embeddings frees that VRAM for the rollout. Defaults unchanged; unit-tested with a stub model.
- **`prefer_sage3_fp8`** native attention selection: SageAttention-3 FP8 where the native build, device, and fp8 DiT backend allow it, cuDNN otherwise (Windows, stub builds), so one preset name is one recipe on every platform.
- **`crazy-robotaxi-omnidreams-rtx-5090` / `-rtx-5090-fast`** runners (`OMNIDREAMS_RTX_5090[_FAST]_PIPELINE_CONFIG`): perf / fast-perf base + CPU text encoder + `prefer_sage3_fp8` + `window_size_t=4`. The window matters: the native KV-cache roll buffer is a ~7.5 GiB allocation outside the torch allocator at window 6, which on a 32 GB card tips the process into unified-memory spill under WSL2 (2–6 fps, never recovers).
- `--stats-path` now records the full per-step metrics (engine, PhysX, realtime margin) that the model loop already computed; previously only the pipeline finalize metrics reached the `StepResult`.

## Results (GeForce RTX 5090 32 GB, `boulevard_district`, 8-frame chunks)

| Runner | Frame | Model step | End to end | Process peak |
|---|---|---|---|---|
| stock `perf` | 1168x640 | — | OOM at prewarm (with CPU encoder alone: 32 GB peak, spills to 2–6 fps) | 32 GB |
| `rtx-5090` (WSL2, headless) | 1168x640 | 258 ms | 21 fps | 27–31 GiB prewarm, 21.7 GiB steady |
| `rtx-5090-fast` (WSL2, headless) | 1024x560 | 156 ms | 30.5 fps | 25.3 GiB |
| `rtx-5090-fast` (Windows native, headless) | 1024x560 | 194 ms | 31.1 fps | 15.3 GiB torch reserved |
| `rtx-5090-fast` (Windows native, `--mode native-window` + HUD) | 1024x560 | 204 ms | 28.7 fps over 150 blocks | 15.2 GiB torch reserved |

Windows uses cuDNN FP8 attention (Sage3 is Linux-only; 157 vs 119 ms diffuse at 1024x560). End to end includes the game engine on the model thread (60–100 ms per chunk: PhysX traffic prep + conditioning render), which is the remaining gap at 1168x640; overlapping engine and model work is future work.

Harness, phase-tagged VRAM probe, and the write-up are on `dev/jmccaffrey/taxi-5090-bench` (not part of this PR).

## How to run

```bash
uv run --package flashdreams-omnidreams flashdreams-run-v2 \
  crazy-robotaxi-omnidreams-rtx-5090-fast --mode native-window

# headless / WSL2
uv run --package flashdreams-omnidreams flashdreams-run-v2 \
  crazy-robotaxi-omnidreams-rtx-5090-fast --mode mp4 --output-path out.mp4 -- \
  --no-ui --game-mode taxi --total-blocks 300 \
  --map apps/crazy_robotaxi/crazy_robotaxi/maps/boulevard_district.robotaxi.yaml
```

## Validation

- `apps/crazy_robotaxi/tests`, `integrations_v2/omnidreams/tests/test_recipe_configs.py`, `flashdreams/tests/test_cosmos_reason1_host.py`, `flashdreams/tests/test_encoder_lifecycle.py`: 280 passed. (`test_application_bridge.py::test_application_reset_waits_for_an_in_flight_step` fails identically on the base branch.)
- GPU runs above on WSL2 (CUDA 13, torch 2.12.1+cu130) and Windows native (MSVC 2022 builds of the native FP8 DiT, PhysX, and Ludus plugin).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


apps/crazy_robotaxi/crazy_robotaxi/headless_ui.py
```diff
@@ -0,0 +1,63 @@
...
+
+from flashdreams.api_v2.loop import IUILoop, ModelInferenceState
+from flashdreams.runtime_v2.blit_model_output_to_screen_loop import frame_to_layout
+from flashdreams.runtime_v2.step_result import StepResult
+from flashdreams.runtime_v2.user_input_events import UserInputEvents
+
+from crazy_robotaxi.ui import TaxiHudState
+
+
+class CrazyRobotaxiHeadlessUILoop(IUILoop[TaxiHudState]):
```
ArielG-NV, line 18: Looks like you can just do something like: `class CrazyRobotaxiHeadlessUILoop(BlitModelOutputToScreenLoop[TaxiHudState]):`

instead of reimplementing the `BlitModelOutputToScreenLoop` class fully, and just replace `reset` (call `super().reset()`, then reset `self.state`)

ArielG-NV (commented):
Edit: Reviewed PR, 1 Suggestion.

aidanfnv:
Is the model framerate supposed to be uncapped here? Or rather, is it uncapped?
I tested this on my RTX 6000 Pro out of curiosity, and it runs at 50+ fps, throwing everything off.
Not necessarily a problem to fix with this PR (I can create a follow-up) but we should probably cap that framerate to 30, instead of relying on target hardware not performing that well.

That being said, I am surprised how well it manages to hold up quality-wise.

aidanfnv:
Accidentally closed this. I merged #616, which merged the branch this was targeting, and that branch got automatically deleted, automatically closing this PR.

apps/crazy_robotaxi/crazy_robotaxi/session.py
```diff
@@ -564,7 +563,13 @@ def step(self, step_index: int, events: UserInputEvents) -> list[StepResult]:
                 output=video,
                 frame_count=count,
                 output_layout=VideoTensorLayout.tchw,
-                metrics=finalize_metrics,
+                metrics=metrics,
+            ),
+            StepResult(
```
aidanfnv, line 568: Looks like a duplicate `hdmap` `StepResult` got introduced by merging my PR stack and then merging `main` into this PR.

integrations_v2/omnidreams/config.py
```diff
@@ -211,6 +211,50 @@
...
+SageAttention-3 FP8 attention is used where the native build supports it.
+"""
+
+
+OMNIDREAMS_RTX_5090_FAST_PIPELINE_CONFIG = cast(
+    OmnidreamsPipelineConfig,
+    derive_config(
+        OMNIDREAMS_FAST_PERF_PIPELINE_CONFIG,
+        name="omnidreams-rtx-5090-fast",
+        text_encoder=dict(run_on_cpu=True, embedding_cache_size=8),
```
aidanfnv, line 242: I can confirm that `run_on_cpu=True` breaks the prompt-changing live-edit features in that the game effectively pauses on each prompt change.
Maybe that feature ("MAP CONTEXT::Enabled" on the settings menu) should be broken up to allow more configuration?
Currently turning on that feature results in prompts that consist of a `scenery description` + `taxi movement` (i.e. "taxi is stationary", "taxi is driving forward") + `road shape` (i.e. "The road curves left ahead") + `weather`. If those are made individual toggles that wouldn't make the pauses go away but should make them far less frequent, especially given that caches are involved.
Again, not necessarily something that ought to be fixed by this PR. I can make that a follow-up.

apps/crazy_robotaxi/crazy_robotaxi/headless_ui.py
mccaffrey-jonathan: Done in 15ca40e1 — `BlitModelOutputToScreenLoop` is now generic over the loop state and grew a `frames_to_blit()` hook, so this class is the two overrides that actually differ (`frames_to_blit` + `reset`); `step`, `is_finished`, and `_initialize_loop_state` are inherited, and `_frame_to_layout` goes back to private now that nothing outside the module needs it.

It can't inherit the blit `step` verbatim, though: since #616 the model loop publishes `[video, hdmap, bev]`, and compositing all of them would draw the HD map over the generated video in the mp4. The hook keeps that difference to one line.

apps/crazy_robotaxi/crazy_robotaxi/session.py
mccaffrey-jonathan: Confirmed — fixed in f1733933. The channels are back to `[video, hdmap, (bev)]`.

Worth noting it was also failing `test_leaderboard_does_not_finish_the_v2_model_loop`, which asserts the terminal-state step publishes two results; that test passes again.

integrations_v2/omnidreams/config.py
mccaffrey-jonathan: Agreed, and a follow-up is the right place — but I measured the parts on the 5090 so the follow-up can be scoped, and the numbers argue for exactly the split you propose.

| | |
|---|---|
| CPU encode (`run_on_cpu=True`) | **3.0–3.2 s per prompt** (the in-game log agrees: `host encode of 1 prompt(s) took 2.9s`), no batching win |
| GPU-resident encode | **51 ms** steady, 47 ms each in a batch of 4 |
| Encoder weights on GPU | 15.7 GiB |
| One `full_concat` embedding `[1, 512, 100352]` bf16 | **98.0 MiB** |
| Host→device copy of a cached embedding | 7.0 ms pageable, **4.0 ms pinned** |

So with live edit on (native DiT disabled, 306 ms chunks at 1024x560) every *new* prompt state costs ~10 chunks ≈ 2.4 s of video. Every repeat is free, because `StyleAbility._prompt_embeddings` already caches it.

The suffix space is fully enumerable from the resolved map (`road.prompt_context` × node phrase/context × curve{–,left,right} × motion{forward,stationary,reverse}):

| map | roads / nodes | distinct map-context prompts | embeddings |
|---|---|---|---|
| `flashdreams_raceway` | 13 / 13 | 126 | 12.1 GiB |
| `boulevard_district` | 81 / 75 | 630 | 60.3 GiB |

Two things fall out of that:

1. **Your toggle split is the big lever.** `boulevard_district`'s 81 roads share only **7** distinct `prompt_context` strings — the curve and motion clauses are the 9× multiplier, not the neighborhoods. A scenery-only mode is 7 prompts (0.7 GiB) on boulevard and 14 (1.3 GiB) on the raceway, which is small enough to pre-encode outright.
2. **Storage, not compute, is the constraint.** Encoding all 126 raceway prompts on the GPU is ~6 s; all 630 boulevard prompts ~30 s. But `_prompt_embeddings` holds *device* tensors, so today every new map state permanently costs 98 MiB of VRAM on the card this PR just barely fit — a lap can add GiBs. `replace_text_from_embeddings` already does `.to(self.device)`, so that cache can live in host memory for a 4 ms copy per swap.

I'm prototyping the follow-up as: enumerate the map's suffix space, encode off the model thread (the map suffix is advisory and already deferred to a chunk boundary, so landing it a chunk or two late is invisible), and keep the cache host-side with an LRU bound. That removes the pause without needing the toggle split — but the split still makes the whole thing 9× cheaper, so I'd take both.

For completeness, the two options I ruled out: keeping the encoder resident doesn't fit (15.7 GiB next to a 15.1 GiB rollout is past the spill cliff), and hot-swapping it per prompt is worse than the CPU encode (~1 s per direction over PCIe). Layer-streamed execution would be ~0.5 s but `accelerate.cpu_offload` fails on this encoder (`Cannot copy out of meta tensor`).

mccaffrey-jonathan:
Pushed three review fixes (`287abcfb..10b80b6f`):

- **f1733933** — drop the duplicate HD-map `StepResult` the `main` merge introduced, so the model loop publishes `[video, hdmap, (bev)]` again and the HUD stops compositing the HD map as BEV.
- **15ca40e1** — build the headless UI loop on `BlitModelOutputToScreenLoop` per @ArielG-NV's suggestion: the blit loop is now generic over its loop state with a `frames_to_blit()` hook, and `_frame_to_layout` goes back to private.
- **10b80b6f** — `--no-ui --game-mode race` with no `--race-course` used to hang forever (the HUD stops at the course menu, which a headless run has no way to answer, so it published menu frames instead of generating `--total-blocks`). It now fails at session init naming the flag to pass. Found while re-benching this branch.

Validation: 406 CPU tests pass across `apps/crazy_robotaxi`, `apps/t2v`, `apps/cam2v`, `integrations_v2/omnidreams/tests/test_recipe_configs.py`, `flashdreams/tests/test_cosmos_reason1_host.py`, `flashdreams/tests/test_encoder_lifecycle.py`, and `flashdreams/tests/test_benchmark_harness.py`; ruff and ty are clean on the changed files. On the 5090 (WSL2), `crazy-robotaxi-omnidreams-rtx-5090-fast` headless at 1024x560 still measures 157.8 ms model step / 121.2 ms diffuse / **30.5 fps end to end**, torch peak 14.5 GiB, no realtime misses — unchanged from the table above, and the mp4 carries the generated video channel rather than the HD map.

@aidanfnv on the uncapped framerate: it is uncapped, and the reason is that the pacer's `frequency` is `SessionDesc.frames_per_second_for_step` (30) while one model step is a whole 8-frame chunk — so the ceiling is 30 chunks/s = 240 video fps. `register_session_loop_objects` also requires an integer frequency, so the honest 30/8 = 3.75 steps/s cannot be expressed there; a cap would have to sleep the `realtime_margin_ms` the taxi model loop already computes, which seems like its own change. Happy to take that follow-up if you haven't filed it.

aidanfnv (approved):
LGTM. Re: the framerate cap, I have a potential fix at #642, but it contains this PR's commits as well. I will drop those once this is merged and I rebase that PR.

integrations_v2/omnidreams/config.py
aidanfnv: I have a follow-up for the split at #643

mccaffrey-jonathan:
Two more commits, both from a self-review pass rather than new review feedback:

- **d9709d9f** — `patch.multiple(create=True)` in `apps/t2v/t2v/testing.py` applied `create=True` to every attribute it patched, so adding `frames_to_blit` there also stopped `step`, `is_finished`, and `reset` from failing if the blit loop ever renames one. The new attribute is patched on its own now, matching what `tests/test_cli.py` already did with `raising=False`.

- **cadc2b90** — `CosmosReason1TextEncoder._embedding_cache` (added by this PR) is not safe for two callers, and a game host that encodes prompts off its model thread has exactly that. The read path mutated it: `get` followed by `move_to_end` raises `KeyError` if the other caller evicted that key in between, which at `embedding_cache_size=8` is a routine eviction rather than an exotic race, and `while len(...) > size: popitem()` is check-then-act, so two inserts can drop each other's fresh entries.

  The fix evicts by insertion instead of by use, which makes a hit a plain lookup needing no lock — it reads either the old mapping or the new one, both valid — and takes a lock for the insert, the eviction, and the invalidating `clear` in `_apply`. **This is a deliberate behavior change**: the cache holds a handful of entries, so use-ordering was not buying meaningful hit rate, and the alternative was a lock on every hit. `test_embedding_cache_hits_and_evicts_in_insertion_order` is updated accordingly and fails against the previous encoder.

407 CPU tests pass; ruff and ty are clean on the changed files. Re-measured on the 5090 at 1024x560 with live edit on, 300 blocks: 307 ms model step, 39.6 ms engine, 21.1 fps end to end — unchanged.

mccaffrey-jonathan:
Found and fixed the `cpu` job failure — **fa93e62b**. It was ours, and it has been failing since `0c04bf62`.

`from torch import Tensor`, added alongside the headless UI loop in 94323869, sorted above the first-party block, which `ruff check --select I` rejects. CI runs `pre-commit run -a` repo-wide, so this failed the **Run linter checks** step and every step behind it was skipped — including `Run CPU unit tests`. That means the unit tests have never actually executed in CI for this PR, on any commit.

Locally, with the same commands CI uses:

- `uv run --group lint pre-commit run -a` — all hooks pass, no files modified (previously: `ruff fix` failed, 1 error).
- `uv run --no-sync pytest -m ci_cpu` — **1982 passed, 2 skipped**.

@aidanfnv the `/ok to test d9709d9` run predates this fix, and your approval is on `d9709d9f` — `cadc2b90` (the embedding-cache change described above) and `fa93e62b` landed after it, so the current head is `fa93e62b` and needs a fresh vet before CI will run on it.

aidanfnv (approved):
Still LGTM

